PR #4, opened 2019-12-04
ZT-458: add ziti-enroller

rentallect:


rentallect:
> Is there another PR to remove it from ziti-edge?

Yeah, but I hadn't pushed it until just a moment ago.

PR #5, opened 2019-12-05
Update ziti-edge to latest

plorenz:


go.mod
```diff
@@ -2,6 +2,10 @@ module github.com/netfoundry/ziti-cmd
 
 go 1.13
```
plorenz: I added these b/c I thought it would be convenient to have them there, but commented out. If people disagree I can take them back out.

plorenz:
Superceded

PR #10, opened 2019-12-11
Update CLI for edge router policy changes. Pull in 0.7 edge changes. …

plorenz:
…Fixes GH-9.


plorenz:
> We still need to consider pushing the `ziti-edge`-related CLI bits back into the `ziti-edge` repository, and turn the part of the `ziti` command that's present in `ziti-cmd` into just a shell that pulls those components into the `cobra` infrastructure.
> 
> When that's done, we can jettison the `ziti-fabric` CLI, and merge that into `ziti fabric`.

I added GH-11 to track this

PR #14, opened 2019-12-18
Netfoundry/ziti edge#18

andrewpmartinez:


andrewpmartinez:
Need to update ziti-edge dep once it is built and merged.

PR #24, opened 2020-01-06
Adds .gitattribute file

andrewpmartinez:


.gitattributes
```diff
@@ -0,0 +1,42 @@
+*.txt eol=lf
```
scareything: I'm not familiar with managing line endings in git... Is this a high maintenance approach - e.g. specifying each file extension that we might care about

I'm curious why you didn't employ the settings mentioned in https://www.git-scm.com/docs/gitattributes#_end_of_line_conversion (`core.autocrlf=true`, `* text=auto`)?

andrewpmartinez: `* text=auto` will treat any file not specified in the `.gitattributes` file as text.

Seems fine unless someone adds a new file type that isn't text and git treats it as such and then looks for bytes that match line endings.

`autocrlf` tries to "do the right thing" with line endings. It seems great until you have a tool that only works w/ LF. In that case it fails.

The safest way in my opinion to deal with it is to be explicit and if not defined, take it as the committer writes it. As for upkeep, it is only an issue if you introduce a new file extension that is text. I doubt that will be often.

PR #25, opened 2020-01-08
Service policies

plorenz:


ziti/cmd/ziti/cmd/edge_controller/root.go
```diff
@@ -50,7 +51,9 @@ func newCmdEdgeController(f cmdutil.Factory, out io.Writer, errOut io.Writer) *c
 }
 
 func setJSONValue(container *gabs.Container, value interface{}, path ...string) {
-	if _, err := container.Set(value, path...); err != nil {
-		panic(err)
+	if value != nil {
+		if _, err := container.Set(value, path...); err != nil {
+			panic(err)
+		}
```
andrewpmartinez: You might want to set a value to nil for something like a PATCH.

plorenz: Fixed

PR #32, opened 2020-01-17
fixes for sidecar quickstart

qrkourier:


quickstart/kubernetes/sidecar-tunnel/kubernetes-sidecar-tunnel-quickstart.md
```diff
@@ -99,6 +99,15 @@ Save the following yaml to a file named tunnel-sidecar-demo.yaml
...
             - name: ziti-tunnel-persistent-storage
               mountPath: /netfoundry
+            securityContext:
+              capabilities:
+              add:
+                - NET_ADMIN
+          dnsPolicy: "None"
+          dnsConfig:
+            nameservers:
+              - 127.0.0.1
```
scareything: It doesn't matter for our demo/QS, but in general this technique of configuring DNS will prevent any other sidecars in the pod from referring to other Kubernetes services via DNS (e.g. other-svc.other-namespace.svc.cluster-domain.example).

The only alternative I can come up with is to update the ziti-tunnel image so that it stuffs 127.0.0.1 into /etc/resolv.conf when it detects that it is running in a pod. There may be other ways but that's the only option I know of right now.

If we don't have a technical solution, it may be worth adding a comment here to point that this won't work if your pod needs Kubernetes DNS.

ziti-tunnel/docker/BUILD.md
```diff
@@ -15,12 +15,11 @@ ziti binaries are in a different artifactory server and/or repository.
 You can also build for other architectures (e.g. linux/arm) with docker's `buildx`
 sub-command. See https://docs.docker.com/buildx/working-with-buildx/ for specifics.
 
-    $ ziti_version="0.5.7-2546"
+    $ ziti_version="0.5.8-2554"
     $ docker buildx create --use
     $ docker buildx build \
         --platform linux/amd64,linux/arm/v7 \
         --build-arg ZITI_VERSION="${ziti_version}" \
-        --build-arg ARTIFACTORY_USERNAME=shawn.carey \
```
scareything: I think we should keep this line, even if the specific username is ellipsed out. The ARTIFACTORY_USERNAME argument is required, and referring to it here makes this text more usable as a copy/paste source.

scareything: disregard :)

quickstart/kubernetes/sidecar-tunnel/kubernetes-sidecar-tunnel-quickstart.md
dovholuknf: sounds like a comment in the docker file would be prudent if it prevents other sidecars. might be the breadcrumb someone needs someday to figure out why their sidecar isn't working

scareything: My $0.02 - this NET_ADMIN info exists in both the Docker image README and in the ziti-tunnel README, along with a bunch of other info about running ziti-tunnel. Adding this one specific fact as a comment in the Dockerfile seems like low value.

PR #35, opened 2020-01-23
Allow linking configs to services

plorenz:


ziti/cmd/ziti/cmd/edge_controller/update_identity.go
```diff
@@ -0,0 +1,99 @@
+/*
+	Copyright 2020 Netfoundry, Inc.
```
dovholuknf: NetFoundry (capital f)

PR #42, opened 2020-02-10
Try travis build

plorenz:
Note: this is not publishing to github or artifactory yet, but I wanted to get the fixes in ziti-edge out so that Jeremy is unblocked on ZAC work


test/data-maker/data-maker.go
```diff
@@ -1,3 +1,5 @@
+// +build utils
+
 /*
 	Copyright 2019 NetFoundry, Inc.
```
dovholuknf: touched files should have the copyright updated to 2019-2020

common/version/info_generated.go
```diff
@@ -0,0 +1,11 @@
+// Code generated by ziti-ci. DO NOT EDIT.
+
+package version
+
+const (
+	Version   = "v0.9.0"
+	Revision  = "ea556fc18740"
+	Branch    = "travis-build"
+	BuildUser = "Paul Lorenz"
```
dovholuknf: will ci overwrite this?

test/data-maker/data-maker.go
plorenz: looking forward to some automation around this :)

common/version/info_generated.go
plorenz: yes. Right now it will only be reflected in the tag. Considering having master be updated as well, so even if you don't run it locally the version will be closer to accurate

test/data-maker/data-maker.go
plorenz:
```plorenz@carrot:~/work/nf/ziti-cmd$ cat ~/bin/update-copyright 
git diff --name-only origin/master | grep "\.go$" | xargs -n 1 -I {} sed -i 's/Copyright 2019 NetFoundry/Copyright 2020 NetFoundry/g' {}```

PR #54, opened 2020-02-28
Update 0.11

ekoby:


doc/release-notes-0.11.md
```diff
@@ -0,0 +1,12 @@
+This page discusses the changes that you need to be aware of when migrating your Ziti deployment from version 0.10.x to version 0.11.x
+
+# Theme
+Ziti 0.11 includes the following: 
+ 
+ * Ziti connections from Ziti SDK client to services hosted by SDK are encrypted end-to-end (no API changes)
```
smilindave26: Rather than "are encrypted" should this say "can be encrypted"?

go.mod
```diff
@@ -21,10 +21,10 @@ require (
 	github.com/kardianos/osext v0.0.0-20190222173326-2bc1f35cddc0
 	github.com/keybase/go-ps v0.0.0-20190827175125-91aafc93ba19
 	github.com/michaelquigley/pfxlog v0.0.0-20190813191113-2be43bd0dccc
-	github.com/netfoundry/ziti-edge v0.10.3
-	github.com/netfoundry/ziti-fabric v0.9.5
+	github.com/netfoundry/ziti-edge v0.11.0
+	github.com/netfoundry/ziti-fabric v0.10.0
```
smilindave26: Fabric at 10.0?

doc/release-notes-0.11.md
ekoby: "can be" sounds like it is an option left up to the user. with v0.11 SDK on both sides connection *will be* encrypted

go.mod
ekoby: fabric versions will be lagging behind because we don't push bumps upstream only down

doc/release-notes-0.11.md
smilindave26: It should be optional, no?  E.g., the app is already end-to-end encrypted and the customer doesn't want/need the extra overhead.

PR #55, opened 2020-02-28
Add version command to CLI and do some session management tidying

plorenz:


ziti/cmd/ziti/cmd/edge_controller/version.go
```diff
@@ -0,0 +1,75 @@
+/*
+	Copyright 2019 NetFoundry, Inc.
```
dovholuknf: 2020 :)

PR #62, opened 2020-03-22
Add support for role attribute querying to CLI

plorenz:


doc/release-notes-0.13.md
```diff
@@ -0,0 +1,55 @@
+# 0.13 is not yet released! These are pre-release notes for features coming in 0.13. Some of these features may be released in 0.12 point releases
```
andrewpmartinez: Before 0.13 goes out I want to add a few endpoints that are aliases for existing functionality but clean up swagger/auto-gen.

doc/release-notes-0.13.md
```diff
@@ -0,0 +1,55 @@
...
+    ziti edge controller list identity-role-attributes
+    ziti edge controller list service-role-attributes
+    
+Example output:
+
+    $ ec list service-role-attributes "true sort by id desc limit 5" -j
+    {
+        "meta": {
+            "filterableFields": [
+                "id"
```
andrewpmartinez: filterable field `id`

I assume that this means this returns a list of string ids.

andrewpmartinez:
Do we want to commit to master?
This will cause what is documented as 0.13 functionality will be in 0.12.x.

doc/release-notes-0.13.md
plorenz: It returns a list of role attributes. I was going to use roleAttribute instead of id, but there were a couple of technical reasons I went with ID. If you think roleAttribute is significantly better than id, I will address those technical issues and change it.

plorenz:
> Do we want to commit to master?
> This will cause what is documented as 0.13 functionality to be in 0.12.x.

I'll squash all the release notes into a single doc and give these a point release note. I can't guarantee their correctness unfortunately. Ordinarily I'll bump the minor, but that amounts to a major bump now. I can't guarantee that the autogenerated build number will line up exactly, if someone else merges around the same time as I do, or if a smoke build flakes. 

I may just bump the minor anyway. If all the release notes are together, it doesn't feel like as big a deal to bump the number.

plorenz:
Merged all release notes into single file.
Updated PR target to release-0.13 so we can stage more changes for the next release

PR #63, opened 2020-03-23
Remove separate release notes

plorenz:


andrewpmartinez:
Lets get this merged! I added the Auto CA identity role attributes capability and need to add it to this new file!

plorenz:
> Lets get this merged! I added the Auto CA identity role attributes capability and need to add it to this new file!

New file is already in release-0.13, so you should be good to go. This just removes all the other files.

PR #74, opened 2020-04-01
Phase 14: Chapter 1

michaelquigley:
Updated configuration files and package paths for the `ziti-fabric` phase 14, chapter 1 work.


etc/001.yml
```diff
@@ -1,4 +1,4 @@
-v: 2
```
andrewpmartinez, line 1: Does the `edge.router.yml` file also need to be updated?

michaelquigley: Update pushed.

PR #82, opened 2020-04-20
Add.channel2.connect.options

andrewpmartinez:


doc/release-notes.md
```diff
@@ -1,3 +1,56 @@
+# Release 0.14
```
andrewpmartinez, line 1: I wasn't sure if this was the right section or if it should be in the 13 section or a new 13.x section.

plorenz: I think it should probably go in a 0.13.3 section

andrewpmartinez: Done.

PR #96, opened 2020-05-19
update deps for netfoundry/ziti-edge#158

andrewpmartinez:
- fixes for identity update
- migration to fix orphaned enrollments/authenticators


andrewpmartinez:
See https://github.com/netfoundry/ziti-cmd/pull/97

PR #104, opened 2020-05-27
fixes issue 103 by trimming whitespace

dovholuknf:


dovholuknf:
fixed at https://github.com/openziti/sdk-golang/pull/53/files which is a better place for it

Issue #106, opened 2020-05-29
Add CLI support for updating terminators


plorenz:
Allow updating terminator attributes including

* router
* binding
* address
* precedence
* static cost

```
$ ziti edge controller update terminator --help
updates a service terminator

Usage:
  ziti edge controller update terminator <id> [flags]

Flags:
      --address string      Set the terminator address
      --binding string      Set the terminator binding
  -c, --cost int32          Set the terminator cost
  -h, --help                help for terminator
  -j, --output-json         Output the full JSON response from the Ziti Edge Controller
  -p, --precedence string   Set the terminator precedence ('default', 'required' or 'failed')
      --router string       Set the terminator router
```

Issue #105, opened 2020-05-29
Add CLI support for managing identity service config overrides


plorenz:
Enable commands to list, update and remove identity level service config overrides.

```
$ ziti edge controller list identity service-configs --help
lists service-configs related to a identities instanced managed by the Ziti Edge Controller

Usage:
  ziti edge controller list identity service-configs <id or name> [flags]

Flags:
  -h, --help          help for service-configs
  -j, --output-json   Output the full JSON response from the Ziti Edge Controller
```
```
$ ziti edge controller update identity-configs --help
for the specified identity, use the given config for the given service

Usage:
  ziti edge controller update identity-configs <identity id or name> <service id or name> <config id or name> [flags]

Flags:
  -h, --help          help for identity-configs
  -j, --output-json   Output the full JSON response from the Ziti Edge Controller
  -r, --remove        Remove the sevice config override
```

PR #107, opened 2020-05-29
Add identity service config CLI commands. Add update terminator CLI c…

plorenz:
…ommand. Fixes GH-105. Fixes GH-106

### DIFF
--- a/ziti/cmd/ziti/cmd/edge_controller/common.go
+++ b/ziti/cmd/ziti/cmd/edge_controller/common.go
@@ -49,7 +49,7 @@ func mapNameToID(entityType string, val string) (string, error) {
 
 	entity := list[0]
 	entityId, _ := entity.Path("id").Data().(string)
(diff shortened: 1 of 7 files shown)


ziti/cmd/ziti/cmd/edge_controller/common.go
```diff
@@ -48,7 +49,9 @@ func mapNameToID(entityType string, val string) (string, error) {
 
 	entity := list[0]
 	entityId, _ := entity.Path("id").Data().(string)
-	fmt.Printf("Found %v with id %v for name %v\n", entityType, entityId, val)
+	if val, found := os.LookupEnv("DEBUG"); found && strings.EqualFold("true", val) {
```
andrewpmartinez, line 52: Should we standardize these early w/ a prefix? `ZITI_DEBUG`

plorenz: I like it. Renamed to ZITI_CLI_DEBUG

PR #116, opened 2020-06-09
bascially copy the enroller into a ziti-tunnel subcmd

### LINKED ISSUE #115: ziti-tunnel should be overloaded to enroll
Right now the only tunneler that requires the enroller is the linux-based tunnel. Add enroll capability to the ziti-tunnel and possibly deprecate the enroller - maybe

dovholuknf:
resolve #115 by adding a subcommand named 'enroll' to ziti-tunnel

### DIFF
--- a/ziti-tunnel/cmd/ziti-tunnel/subcmd/enroll.go
+++ b/ziti-tunnel/cmd/ziti-tunnel/subcmd/enroll.go
@@ -1,0 +1,156 @@
+/*
+	Copyright NetFoundry, Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package subcmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/michaelquigley/pfxlog"
+	"github.com/openziti/foundation/identity/certtools"
+	"github.com/openziti/sdk-golang/ziti/enroll"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"io/ioutil"
+	"os"
+	"strings"
+)
+
+// global state used by all subcommands are located here for easy discovery
+var verbose bool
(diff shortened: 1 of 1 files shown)


dovholuknf:
ziti-tunnel now has enroll

```
build/ziti-tunnel
Ziti Tunnel

Usage:
  ziti-tunnel [command]

Available Commands:
  enroll      enroll an identity
  help        Help about any command

... 

Usage:
  ziti-tunnel enroll [flags]

Flags:
      --ca string       Additional trusted certificates
  -c, --cert string     The certificate to present when establishing a connection.
  -h, --help            help for enroll
  -n, --idname string   Names the identity. Ignored if not 3rd party auto enrollment
  -j, --jwt string      Enrollment token (JWT file). Required
  -k, --key string      The key to use with the certifcate. Optionally specify the engine to use. supported engines: [pkcs11]
  -o, --out string      Output configuration file.

Global Flags:
  -i, --identity string        Path to JSON file that contains an enrolled identity
      --log-formatter string   Specify log formatter [json|pfxlog|text]
  -r, --resolver string        Resolver configuration (default "udp://127.0.0.1:53")
      --svcPollRate uint       Set poll rate for service updates (seconds) (default 15)
  -v, --verbose                Enable verbose mode

```

example output:
image

ziti-tunnel/cmd/ziti-tunnel/subcmd/enroll.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	flags := enroll.EnrollmentFlags{
+		CertFile:      certPath,
+		KeyFile:       keyPath,
+		Token:         tkn,
+		IDName:        idname,
+		AdditionalCAs: caOverride,
+	}
+	conf, err := enroll.Enroll(flags)
```
smilindave26, line 122: Should we keep the stand-alone enroller too?  I've used it in the past, but have no problem using `ziti-tunnel` for an enrollment json I use somewhere else.  A little sad to have duplicate code, but updating `ziti-enroller` to call `ziti-tunnel` seems like a waste...

dovholuknf: yeah - i dunno. i added that same comment on the issue but not the pr:
> Add enroll capability to the ziti-tunnel and possibly deprecate the enroller - maybe

i would probably err on removing it with the 'we can put it back if we need it' idea

dovholuknf: `alias ziti-enroller=ziti-tunnel enroll $0`? :)

scareything: I think ziti-tunnel could look at argv to see how it was called and go into enroller mode if args[0] is 'ziti-enroller`?

PR #122, opened 2020-06-16
Release 0.15.0

andrewpmartinez:


doc/release-notes.md
```diff
@@ -1,6 +1,23 @@
 # Release 0.14.13
```
emoscardini, line 1: @andrewpmartinez Release 0.15.0?

dovholuknf: @andrewpmartinez i have a pr to this branch out there that updates the readme to 0.15.0 as edward points out. also shouldn't the databaser snapshot stuff get put into 0.15.0?

go.mod
```diff
@@ -2,13 +2,13 @@ module github.com/openziti/ziti
 
 go 1.14
 
-// replace github.com/openziti/foundation => ../foundation
+replace github.com/openziti/foundation => ../foundation
```
dovholuknf, line 5: re-comment?

ziti-fabric/subcmd/snapshot_db.go
```diff
@@ -0,0 +1,60 @@
+/*
```
dovholuknf, line 1: why is this in fabric ? i'd not have thought to look for it in ziti-fabric

ziti-fabric/subcmd/snapshot_db.go
```diff
@@ -0,0 +1,60 @@
...
+var snapshotDbClient *mgmtClient
+
+func snapshotDb(cmd *cobra.Command, args []string) {
+	ch, err := snapshotDbClient.Connect()
+	if err != nil {
+		panic(err)
+	}
+
+	requestMsg := channel2.NewMessage(int32(mgmt_pb.ContentType_SnapshotDbRequestType), nil)
+	responseMsg, err := ch.SendAndWaitWithTimeout(requestMsg, 5*time.Second)
```
dovholuknf, line 46: i mean- 5 seconds _should_ be enough but is there any harm in waiting longer? just in case? if the timeout hits can we cancel the request to backup somehow?

go.mod
plorenz: fixed

ziti-fabric/subcmd/snapshot_db.go
plorenz: It needs to be in both places. It's fabric functionality, so it needs to be in ziti-fabric, but it should be available via the REST API as well. I haven't added it to the edge yet, not sure if I need to or not. When the fabric management API gets consolidated into REST, it'll be available there. I'll have to ask if ziti-fabric is sufficient for now

ziti-fabric/subcmd/snapshot_db.go
plorenz: I wouldn't worry too much about it for now. ziti-fabric will go away in the short to medium term.

ziti-fabric/subcmd/snapshot_db.go
plorenz: Went ahead and added it to edge so it's available in both now

PR #154, opened 2020-07-27
Docker compose for ziti-tunnel

qrkourier:


ziti-tunnel/docker/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
+version: "3.3"
```
scareything, line 1: Should this file named in a way that suggests it is an example; e.g. docker-compose-example.yml? _docker-compose.yml_ feels a little misleading since anyone who uses it will need to modify it, and then be left with a modified file in their git workspace.

qrkourier: There's some incentive to use that filename in that it is the default filename for a Docker Compose file. An alternative name would just need to be given as a parameter when running `docker-compose`.

PR #159, opened 2020-07-31
E2e.terminators

andrewpmartinez:


andrewpmartinez:
> Should we add 'encryption required' to services list?

Yes we should. Adding.

PR #175, opened 2020-08-11
Update deps for xgress race condition fix

plorenz:


michaelquigley:
Tested in repro environment with `for i in {1..500}; do date | nc 127.0.0.1 4506; done`... Looks good!

PR #179, opened 2020-08-17
refine compose file and README example

qrkourier:


plorenz:
Hi @qrkourier can you please target release-next, instead of master?

qrkourier:
Please let me know if you'd like to see something different before this merges.

PR #184, opened 2020-08-24
Ziti Probe API Session Reuse

andrewpmartinez:
- no longer exits on connection issues (required sdk golang merge)
- no longer exists on missing service
- misc small changes

### DIFF
--- a/common/version/info_generated.go
+++ b/common/version/info_generated.go
@@ -3,9 +3,9 @@
 package version
 
 const (
-	Version   = "v0.9.0"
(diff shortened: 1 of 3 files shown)


ekoby:
should we some form of exponential backoff if we can't auth/connect for some reason

andrewpmartinez:
> should we implement some form of exponential backoff if we can't auth/connect for some reason

I am certainly okay with us doing that as an additional enhancement.

andrewpmartinez:
Added the `version` command as it was useful during debugging sessions.

PR #187, opened 2020-09-01
Update deps. Improve ziti-fabric cli for terminators and links

plorenz:


etc/edge.router.tmpl.yml
```diff
@@ -0,0 +1,90 @@
+v: 3
```
andrewpmartinez, line 1: Interesting. I have a number of scripts that make use of template files just like this - just outside of the repo. 

Would it make sense to:
- move this to `etc/tmpl` as to not clutter this directory
- parameterize most settings

?

plorenz: sure. The controller host/ip could be templated. Anything else you'd want to templatize?

michaelquigley: We really need some sort of namespaced+`FABLAB_TEMPLATE_PATH'd template file resolution. If we had a `FABLAB_TEMPLATE_PATH` variable with `:` separated entries, people could put that stuff wherever they want.

michaelquigley: Wrong project. Disregard. 💯

plorenz: Moved and templated controller host/port

ziti-fabric/subcmd/list_links.go
```diff
@@ -58,22 +58,24 @@ func listLinks(cmd *cobra.Command, args []string) {
...
+						status := "up"
 						if l.Down {
-							down = "->down<-"
+							status = "down"
 						}
 						// Convert nanoseconds to fractional seconds
-						srcLatency := float64(l.SrcLatency) / 1000000000.0
-						dstLatency := float64(l.DstLatency) / 1000000000.0
-						out += fmt.Sprintf("%-6s | %-24s -> %-24s | %-12s | %-4d | %-0.4f %-0.4f %s\n", l.Id, l.Src, l.Dst, l.State, l.Cost, srcLatency, dstLatency, down)
+						srcLatency := float64(l.SrcLatency) / 1_000_000_000.0
```
dovholuknf, line 74: i still like constants over magic numbers but maybe that's just me? :)

dovholuknf (approved):

ziti-fabric/subcmd/list_links.go
plorenz: I'm not sure `float64(l.SrcLatency) / OneBillion` would provide any additional clarity :)

`float64(l.SrcLatency) / NanosecondsToSecondsScaleFactor` might, but I think the comment directly above does the same

Issue #155, opened 2020-07-29
enroll subcommand w/out args should print help


qrkourier:
If only for consistency w/ `ziti-router` and `ziti` CLI,

```bash
❯ ./ziti-tunnel enroll
```

should be equivalent to

```bash
❯ ./ziti-tunnel enroll --help
```
optionally after complaining about the missing required flag `--jwt` on stderr

```bash
❯ ./ziti-tunnel version
v0.15.2
```

PR #196, opened 2020-09-15
Show help output on when enroll is missing required args. Fix configT…

plorenz:
…ypes=all processing on list services. Fixes GH-155

### DIFF
--- a/common/enrollment/enroll.go
+++ b/common/enrollment/enroll.go
@@ -32,7 +32,7 @@ import (
 
 // global state used by all subcommands are located here for easy discovery
 var verbose bool
-var jwtpath, outpath, keyPath, certPath, idname, caOverride string
+var outpath, keyPath, certPath, idname, caOverride string
 
 const verboseDesc = "Enable verbose logging."
 const outpathDesc = "Output configuration file."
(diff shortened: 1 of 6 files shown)


common/enrollment/enroll.go
```diff
@@ -45,7 +45,7 @@ const outFlag = "out"
 func NewEnrollCommand() *cobra.Command {
```
plorenz, line 45: Should we just make the JWT the first positional argument to be consistent with other CLIs?

andrewpmartinez: I'm leaning toward yes. If it is required for all uses of the command, it should be a required positional.

andrewpmartinez (approved):

common/enrollment/enroll.go
plorenz: Change made. Also marked ziti-enroller as deprecated, so we can remove it in a few releases.

emoscardini:
@plorenz @andrewpmartinez Making the JWT positional does make sense, but we've already created wrapper scripts to enroll the edge-routers.  This change will break existing scripts already released into the production environment.   Cloud we do both?  positional or have a switch(--jwt/-j)?

PR #207, opened 2020-10-09
fix for openziti/edge#336

andrewpmartinez:
- deal with init/run separation

Depends on: 
- https://github.com/openziti/edge/pull/337
- and a dep update

### DIFF
--- a/ziti-controller/subcmd/run.go
+++ b/ziti-controller/subcmd/run.go
@@ -18,10 +18,10 @@ package subcmd
 
 import (
 	"fmt"
-	"github.com/openziti/ziti/common/version"
-	"github.com/openziti/ziti/ziti/agent"
 	"github.com/openziti/edge/controller/server"
 	"github.com/openziti/fabric/controller"
+	"github.com/openziti/ziti/common/version"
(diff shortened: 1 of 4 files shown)


ziti-controller/subcmd/run.go
```diff
@@ -57,8 +57,9 @@ func run(cmd *cobra.Command, args []string) {
 		}
 
 		ec.SetHostController(c)
+		ec.Initialize()
 
-		go ec.InitAndRun()
+		go ec.RunAndWait()
 
 		if err = c.Run(); err != nil {
```
dovholuknf, line 64: RunAndWait calls Run then will 'wait for shutdown'...  Run is then called again here in this if.  Is that really 'correct'? seems strange to me

dovholuknf: oh, i see that one's calling run on "ec" and one's on "c" so i expect this is correct.

dovholuknf (approved):

ziti-controller/subcmd/run.go
andrewpmartinez: You are correct.

plorenz (approved):

PR #239, opened 2020-11-17
Add GH publishing

plorenz:


.github/workflows/main.yml
```diff
@@ -38,7 +38,7 @@ jobs:
           go get github.com/mitchellh/gox
           $(go env GOPATH)/bin/ziti-ci generate-build-info common/version/info_generated.go version
           CGO_ENABLED=true $(go env GOPATH)/bin/gox -os=darwin -arch=amd64 -output=$GOX_OUTPUT ./...
-          aws s3 sync release/ s3://ziti-cmd-build-tmp/${{ github.run_id }}
+          aws s3 sync --no-progress release/ s3://ziti-cmd-build-tmp/${{ github.run_id }}
```
ekoby, line 41: why not use github workflow artifacts for passing binaries downstream?

ekoby (approved):

.github/workflows/main.yml
plorenz: I didn't know it existed. I'll check it out

plorenz: Looks like I should be able to this capability. I'll add it in the next pass

PR #264, opened 2021-01-12
Add sequential block type to loop tool

plorenz:


michaelquigley (approved):
Still looking forward to the future `loopX` re-write and consolidation.

PR #345, opened 2021-03-23
watch for signal interrupts

andrewpmartinez:
This implementation tries to use the router.Shutdown() method.

Currently it does exit but only after this panic:

```
goroutine 56 [running]:
reflect.MakeSlice(0x1e38570, 0x1991e40, 0x0, 0xffffffffffffffff, 0x0, 0x0, 0x0)
        .../go/src/reflect/value.go:2297 +0x1ac
github.com/openziti/foundation/util/cowslice.Delete(0xc000005b18, 0x19ecd40, 0xc000686160)
        .../repos/openziti/foundation/util/cowslice/registry.go:41 +0x1c5
github.com/openziti/foundation/events.RemoveMetricsEventHandler(...)
        .../repos/openziti/foundation/events/registrations.go:13
github.com/openziti/fabric/router.(*Router).Shutdown(0xc00047bd40, 0x4, 0xc0004ede98)
        .../repos/openziti/fabric/router/router.go:151 +0x8c
github.com/openziti/ziti/ziti-router/subcmd.waitForShutdown(0xc00047bd40)
        .../repos/openziti/ziti/ziti-router/subcmd/run.go:124 +0x1ab
created by github.com/openziti/ziti/ziti-router/subcmd.run
        .../repos/openziti/ziti/ziti-router/subcmd/run.go:96 +0xd4e
```

I am still hunting that down. The other option is simply to call `os.Exit()` as code in `foundation` used to do for us.

Edit: above is fixed in a fabric PR.

### DIFF
--- a/ziti-router/subcmd/run.go
+++ b/ziti-router/subcmd/run.go
@@ -123,7 +123,5 @@ func waitForShutdown(r *router.Router) {
 
 	if err := r.Shutdown(); err != nil {
 		pfxlog.Logger().Info("error encountered during shutdown: %v", err)
-		os.Exit(1)
 	}
-	os.Exit(0)
 }
--- a/ziti-router/subcmd/run.go
+++ b/ziti-router/subcmd/run.go
@@ -31,6 +31,9 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"io"
+	"os"
+	"os/signal"
+	"syscall"
 )
 
 func init() {
@@ -45,6 +48,7 @@ var runCmd = &cobra.Command{
 }
 
 func run(cmd *cobra.Command, args []string) {
+
 	logrus.WithField("version", version.GetVersion()).
 		WithField("go-version", version.GetGoVersion()).
 		WithField("os", version.GetOS()).
@@ -89,6 +93,8 @@ func run(cmd *cobra.Command, args []string) {
 			logrus.Panicf("error registering edge tunnel in framework (%v)", err)
 		}
 
+		go waitForShutdown(r)
+
 		if err := r.Run(); err != nil {
 			logrus.WithError(err).Fatal("error starting")
 		}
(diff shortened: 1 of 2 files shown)


ziti-router/subcmd/run.go
```diff
@@ -105,3 +111,19 @@ func getFlags(cmd *cobra.Command) map[string]*pflag.Flag {
 	})
 	return ret
 }
+
+func waitForShutdown(r *router.Router) {
```
scareything, line 115: I'm pretty sure calling exit here will prevent the tunneler from cleaning up routes and iptables rules on shutdown. See https://github.com/openziti/edge/blob/main/tunnel/intercept/svcpoll.go#L58

andrewpmartinez: I feel like signal handlers to shutdown need to be handled by executing cmds and not in the lower deps. The lower deps should expose a `Shutdown()` function and code in the cmds should call those directly or indirectly.

michaelquigley: the framework should take a `func` as a parameter that the code in the `cmd`s can provide.

michaelquigley: personally, i think this is kind of a big deal... and definitely a reason to hold up 0.19.8.

scareything: I agree with setting up the handlers at a higher level closer to the command logic. I also think that calling exit directly in a signal handler function makes it impossible to have more than one signal handler in the process.

michaelquigley: agreed. and to that end, the framework should support the higher-level calls providing multiple `func`s, and they should all be called at the shutdown event.

andrewpmartinez: I don't see a clean way to coordinate multiple separate signal handlers. If there isn't a good way to do that, I don't think we should support it as I think it would be a good idea to try to stop all components, but if we have a misbehaving one, I don't think we should block on it forever. Instead, we should give a certain amount of time, then output a log message and exit. With separate notify handlers, I don't see a clear way to do that.

michaelquigley: there's one signal handler. that signal handler delegates to one or more functions that are provided to the framework to clean up some part of our runtime. if one of those functions blocks indefinitely, then we shouldn't do anything. in that case, a `kill` will be required to kill the process.

if we end up in cases where we're blocking regularly or often on one of those cleanup functions, then we need to fix that part of our code. but i don't think the framework should try to do anything to mitigate that.

andrewpmartinez: Seems reasonable.

scareything (approved):

PR #349, opened 2021-03-25
Update deps/changelog

plorenz:


CHANGELOG.md
```diff
@@ -2,69 +2,71 @@
...
+## Converged Tunneler/Router (Beta)
 
-## Transwarp beta_1
+ziti-router can now run with the tunneler embedded. It has the same capabilities as ziti-tunnel. As ziti-tunnel gains new features, the combined ziti-router/tunnel should maintain feature parity.
 
-See the Transwarp beta_1 guide [here](doc/transwarp_b1/transwarp_b1.md).
+This is a beta release. It should be relatively feature complete and bug-free for hosting services but intercept support is still nascent. Some features are likely to be changed based on user
+feedback.
+
+NOTE: The converged tunneler/router does not support running in tun mode. This mode may be deprecated for ziti-tunnel as well at some point in the future if no advantages over tproxy mode are evident.
```
dovholuknf, line 30: was any of the converged tunneler work tested in windows and do we need to say "linux only" (or rather "not windows")?

etc/tmpl/edge.router.yml
```diff
@@ -81,6 +81,8 @@ dialers:
 
 listeners:
   - binding: tunnel
+    options:
+      mode: host
```
dovholuknf, line 85: i did a quick scan and did not see 'mode host' referenced. probably nbd but figured i'd point that out

dovholuknf (approved):

etc/tmpl/edge.router.yml
plorenz: I added a note about it in os compatibility. It's also noted in the config section, which was in a previous PR

CHANGELOG.md
plorenz: I added some notes on OS compatibility

PR #361, opened 2021-04-14
changelog and deps

andrewpmartinez:


ziti-router/subcmd/run.go
```diff
@@ -36,7 +36,10 @@ import (
 	"syscall"
 )
```
plorenz, line 38: This doesn't look like it's used anywhere

andrewpmartinez: It is read from the flag meta data. See Line 107

andrewpmartinez: The flag data is passed through the config machinery in fabric/router/config.go to edge/router

plorenz (approved):

PR #370, opened 2021-04-27
add a token dir /enrollment-token for Helm chart

qrkourier:
[This Helm chart](https://github.com/netfoundry/charts)  dynamically creates a storage volume to persist the enrolled identity on /netfoundry, and a separate mount point is needed for the Kubernetes secret volume. This PR proposes /enrollment-token for the secret volume mount. 

The entrypoint script is also improved to handle the error that occurs when ${dir} is not a directory.

### DIFF
--- a/ziti-tunnel/docker/docker-entrypoint.sh
+++ b/ziti-tunnel/docker/docker-entrypoint.sh
@@ -32,20 +32,29 @@ fi
 
 json="${persisted_dir}/${NF_REG_NAME}.json"
 if [ ! -f "${json}" ]; then
-    echo "identity configuration ${json} does not exist"
-    for dir in "/var/run/secrets/netfoundry.io/enrollment-token" "${persisted_dir}"; do
-        _jwt="${dir}/${NF_REG_NAME}.jwt"
-        echo "looking for ${_jwt}"
-        if [ -f "${_jwt}" ]; then
-            jwt="${_jwt}"
-            break
-        fi
+    echo "WARN: identity configuration ${json} does not exist" >&2
+    for dir in  "/var/run/secrets/kubernetes.io/enrollment-token" \
+                "/enrollment-token" \
+                "${persisted_dir}";
+        do
+            [[ -d ${dir} ]] || {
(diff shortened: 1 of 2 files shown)


ziti-tunnel/docker/docker-entrypoint.sh
```diff
@@ -32,20 +32,29 @@ fi
...
-            jwt="${_jwt}"
-            break
-        fi
+    echo "WARN: identity configuration ${json} does not exist" >&2
+    for dir in  "/var/run/secrets/kubernetes.io/enrollment-token" \
+                "/enrollment-token" \
+                "${persisted_dir}";
+        do
+            [[ -d ${dir} ]] || {
+                echo "WARN: ${dir} is not a directory" >&2
```
scareything, line 41: Should this be a WARN? It feels more DEBUG/INFO to me, since the script is probing here. The only requirement is that at least one of the directories is found.

ziti-tunnel/docker/docker-entrypoint.sh
```diff
@@ -32,20 +32,29 @@ fi
...
+            if [ -f "${_jwt}" ]; then
+                jwt="${_jwt}"
+                break
+            else
+                echo "WARN: failed to find ${_jwt} in ${dir}" >&2
+            fi
     done
-    if [ -z "${jwt}" ]; then
-        echo "ERROR: ${NF_REG_NAME}.jwt was not found in the expected locations"
+    if [ -z "${jwt:-}" ]; then
```
scareything, line 53: Is this a default parameter expansion with an empty default? I'm curious how is this different from expanding the variable without using expansion?

qrkourier: Yes, with BASH option no unset in force it is an error to interpolate ${jwt} when unset, and so the script exits prematurely.

qrkourier: That is, when both nounset and errexit are set, which they are, intentionally.

scareything: I see. Good catch!

PR #379, opened 2021-05-07
use xweb, changelog, ziti cmd updates

andrewpmartinez:
- use xweb in fabric/edge updates some startup code for the cmds
- the ziti cmd now looks up the management API path for new and old
  versions


plorenz:
The changelog says the changelog was split but I don't see anything split

andrewpmartinez:
> The changelog says the changelog was split but I don't see anything split

I removed it from the branch but not the message. Will update the comment. The split will be in another pr.

plorenz (approved):

PR #403, opened 2021-05-26
Permit PIPE/FIFO based type as well as FILE type

NicFragale:
The addition of PIPE/FIFO allowance creates a situation where the launcher can use the same code to parse the pipe which does two things.
1.  The PIPE/FIFO can have multiple JWTs in it.
2.  The use of a pipe is not permanent like that of a file.  (no cleanup of the JWT)


dovholuknf:
FYI: updated to point to release-next

PR #417, opened 2021-06-10
Update deps and changelog. Add support for debug operation dumping ap…

plorenz:
…i-sessions


ziti/cmd/ziti/cmd/edge_controller/update_identity.go
```diff
@@ -100,12 +100,12 @@ func runUpdateIdentity(o *updateIdentityOptions) error {
...
 	}
 
-	if o.Cmd.Flags().Changed("defaultHostingCost") {
+	if o.Cmd.Flags().Changed("default-hosting-cost") {
 		setJSONValue(entityData, o.defaultHostingCost, "defaultHostingCost")
 		change = true
 	}
 
-	if o.Cmd.Flags().Changed("defaultHostingPrecedence") {
+	if o.Cmd.Flags().Changed("default-hosting-precedence") {
```
dovholuknf, line 108: is this a bug or a normalization? should it be called out in the release notes?

dovholuknf: lol - i searched for the 'new term' not the 'old term' i see it in there now . would be nice to mention the change though imo

dovholuknf (approved):

ziti/cmd/ziti/cmd/edge_controller/update_identity.go
plorenz: It's not changing the flag. I had the JSON field name there instead of the flag name. So the UI is unchanged, just unbroken now :)

PR #446, opened 2021-07-26
uses API client for CA list logic

andrewpmartinez:


ziti/cmd/ziti/cmd/edge_controller/list.go
```diff
@@ -813,15 +846,42 @@ func runListCAs(o *edgeOptions) error {
...
-		if _, err := fmt.Fprintf(o.Out, "id: %v    name: %v    cluster-id: %v\n", id, name, cluster); err != nil {
+	payload := result.GetPayload()
+
+	if payload == nil {
+		return errors.New("unexpected empty response payload")
+	}
+
+	for _, entity := range result.GetPayload().Data {
+		id := *entity.ID
+		name := *entity.Name
```
plorenz, line 857: I usually use nil for empty slices, i.e. `var identityRoles []string`

plorenz (approved):

ziti/cmd/ziti/cmd/edge_controller/list.go
andrewpmartinez: I took a look at how services and identities are outputting role attributes they both do the same thing which is to output an empty list as an empty object `{}` or if none empty as an array.

```
id: -FdjcL86G    name: proxy    type: Service    role attributes: ["dial"]
id: FQYTNKlth    name: Default Admin    type: User    role attributes: {}
```

Seems odd they are outputting `{}` but it is visual for humans so I'm not overly concerned with that at the moment.

plorenz: In the CLI output context it makes sense :+1:

PR #454, opened 2021-08-05
Improve login

plorenz:
See CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,35 @@
 ## Breaking Changes
 Fabric sessions renamed to circuits. See below for details.
 
+Commands under `ziti edge` now reserve the `-i` flag for specifying client identity. 
+Any command line argumet which previously had a `-i` short version now only has a long version.
+
+For consistency, policy roles parameters must all be specified in long form
+
+This includes the following flags:
+* ziti edge create edge-router-policy --identity-roles --edge-router-roles
+* ziti edge update edge-router-policy --identity-roles --edge-router-roles
+* ziti edge create service-policy --identity-roles --service-roles
+* ziti edge update service-policy --identity-roles --service-roles
+* ziti edge create service-edge-router-policy --service-roles --edge-router-roles
+* ziti edge update service-edge-router-policy --service-roles --edge-router-roles
+* ziti edge create posture-check mfa --ignore-legacy
+* ziti edge update posture-check mfa --ignore-legacy
+* ziti edge update authenticator updb --identity 
+* ziti egde update ca --identity-atributes (now -a)
+
+The `ziti edge` commands now store session credentials in a new location and new format. Existing sessions will be ignored.
+
+The `ziti edge controller` command was previously deprecated and has now been removed. All commands that were previously available 
+under `ziti edge controller` are available under `ziti edge`.
+
 # What's New
 
 * Refactor: Fabric Sessions renamed to Circuits
 * Feature: Links will now wait for a timeout for retrying 
 * Bug fix: Sessions created on the controller when circuit creation fails are now cleaned up
+* Feature: Enhanced `ziti` CLI login functionality
+* Feature: new `ziti edge list summary` command, which shows database entity counts
 
 ## Fabric Sessions renamed to Circuits
 Previously we had three separate entities named session: fabric sessions, edge sessions and edge API sessions. In order to reduce confusion, fabric sessions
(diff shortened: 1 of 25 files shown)


ziti/cmd/ziti/util/certs.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, "", errors.Wrapf(err, "error while statting cert file for %v", id)
+	}
+	result, err := ioutil.ReadFile(certFile)
+	if err != nil {
+		return nil, "", errors.Wrapf(err, "error while reading cert for %v", id)
+	}
+	return result, certFile, nil
+}
+
+func IsServerSelfSigned(host string) (bool, error) {
```
andrewpmartinez, line 57: This is more aptly named `IsServerTrusted`. A self-signed cert can be trusted it if is loaded into the trust store. IIRC golang will defer to the windows trust store.

ziti/cmd/ziti/cmd/edge_controller/login.go
```diff
@@ -80,70 +81,199 @@ func newLoginCmd(f cmdutil.Factory, out io.Writer, errOut io.Writer) *cobra.Comm
...
+	fmt.Printf("Saving identity '%v' to %v\n", id, configFile)
+	config.Identities[id] = loginIdentity
+
+	err = util.PersistRestClientConfig(config)
+
+	return err
+}
+
+func (o *loginOptions) ConfigureCerts(host string, ctrlUrl *url.URL) error {
+	selfSigned, err := util.IsServerSelfSigned(host)
```
andrewpmartinez, line 176: `selfSigned` is more aptly named `isNotTrusted` or `isTrusted`

ziti/cmd/ziti/cmd/edge_controller/login.go
```diff
@@ -80,70 +81,199 @@ func newLoginCmd(f cmdutil.Factory, out io.Writer, errOut io.Writer) *cobra.Comm
...
+					if err != nil {
+						return err
+					}
+				}
+			}
+		} else {
+			if len(certs) == 0 {
+				fmt.Print("WARNING: server supplied cert chain was empty, continuing but login will likely fail")
+			} else {
+				fmt.Printf("Self-signed server certificate chain retrieved from server\n")
```
andrewpmartinez, line 212: Most ziti instances aren't 'self-signed'. Self-signing is using a private key to sign its own certificate.

In ziti, we are usually dealing with an "untrusted CA". CAs are self-signed certs, but for clarity, they are called CAs because they have the ability to delegate trust via the CA flag in their certificate extensions.

I would change the wording to "Unknown certificate authority retrieved from sever".

plorenz: fixed

ziti/cmd/ziti/util/certs.go
```diff
@@ -0,0 +1,150 @@
...
+		if ue, ok := err.(*url.Error); ok && errors.As(ue.Err, &x509.UnknownAuthorityError{}) {
+			return true, nil
+		}
+		return false, err
+	}
+	_ = resp.Body.Close()
+	return false, nil
+}
+
+func GetServerCertChain(host string) ([]byte, []*x509.Certificate, error) {
```
andrewpmartinez: More aptly named "GetWellKnownCerts`. The certs contained in the well-known bundle aren't guaranteed to be any number of chains. It can be a set of individual roots,1 chain with intermediates, or nothing at all.

plorenz: fixed

ziti/cmd/ziti/cmd/edge_controller/login.go
```diff
@@ -80,70 +81,199 @@ func newLoginCmd(f cmdutil.Factory, out io.Writer, errOut io.Writer) *cobra.Comm
...
+				fmt.Printf("Issuer: %v\n", certs[0].Issuer.String())
```
andrewpmartinez: The certs coming back from the well-known bundle aren't ordered in the spec. Showing the first cert in there might not be useful.

The real value to show is to look through the certs in the bundle, order the chains by signer, and find the cert/certs that is validating the server. Then show the highest cert in that chain (root, intermediate, leaf) or show the entire chain from root->leaf.

plorenz: This might be a bit beyond the level of effort I want to put into this at the moment. I don't know how to order certs or tell which one is validating the server. In the short term should I just take out the `Issuer:` line, until either I have more time to dig into certs or someone who's already deeply steeped in cert lore can poke at it?

ziti/cmd/ziti/util/certs.go
```diff
@@ -0,0 +1,151 @@
...
+			Bytes: cert.Raw,
+		})
+		buf.Write(encoded)
+	}
+
+	return buf.Bytes(), certs.Certificates, nil
+}
+
+func AreCertsSame(p common.Printer, server, local []byte) bool {
+	serverCerts := decodeCerts(server)
```
andrewpmartinez, line 117: The order of the certificates in the well-known bundle are not in any order. It is possible to re-order them and have the exact same bundle with the same result.

I don't know if it is worth covering that case, but there can be false positives from the way this is handled now.

ziti/cmd/ziti/cmd/edge_controller/login.go
```diff
@@ -80,70 +81,199 @@ func newLoginCmd(f cmdutil.Factory, out io.Writer, errOut io.Writer) *cobra.Comm
...
+
+func (o *loginOptions) ConfigureCerts(host string, ctrlUrl *url.URL) error {
+	isServerTrusted, err := util.IsServerTrusted(host)
+	if err != nil {
+		return err
 	}
 
-	token, ok := jsonParsed.Path("data.token").Data().(string)
+	if isServerTrusted && o.Cert == "" {
+		wellKnownCerts, certs, err := util.GetWellKnownCerts(host)
```
andrewpmartinez, line 182: After we retrieve them, we should make sure the bundle actually validates the server's certificate.

ziti/cmd/ziti/util/certs.go
plorenz: Ah, ok. Maybe I can order them lexographically to give them a consistent, if arbitrary, order

ziti/cmd/ziti/cmd/edge_controller/login.go
andrewpmartinez: I think that is fine. I can take that on.

ziti/cmd/ziti/util/certs.go
andrewpmartinez: If you order them by their raw bytes, that would cover re-ordering issues for now.

Subject information is only as good as the CA that created it - so I generally avoid using it for anything.

ziti/cmd/ziti/cmd/edge_controller/login.go
plorenz: Think that should be straight forward to add

CHANGELOG.md
```diff
@@ -1,3 +1,131 @@
...
+The `ziti edge` commands now store session credentials in a new location and new format. Existing sessions will be ignored.
+
+## What's New
+
+* Feature: Enhanced `ziti` CLI login functionality
+* Feature: new `ziti edge list summary` command, which shows database entity counts
+
+## Enhanced CLI Login Functionality
+### Certificate Management
+#### Working with self signed certs
```
andrewpmartinez, line 30: Title chainge of some kind. Maybe `Working With CAs` or `Validating Server Identity` or `Server Trust`.

ziti/cmd/ziti/util/certs.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, "", errors.Wrapf(err, "error while statting cert file for %v", id)
+	}
+	result, err := ioutil.ReadFile(certFile)
+	if err != nil {
+		return nil, "", errors.Wrapf(err, "error while reading cert for %v", id)
+	}
+	return result, certFile, nil
+}
+
+func IsServerTrusted(host string) (bool, error) {
```
andrewpmartinez, line 58: Is the bool state on this reversed?

On error that is UnknownAuthority -> return true.

ziti/cmd/ziti/cmd/edge_controller/login.go
```diff
@@ -80,70 +81,199 @@ func newLoginCmd(f cmdutil.Factory, out io.Writer, errOut io.Writer) *cobra.Comm
...
+					o.Cert, err = util.WriteCert(o, ctrlUrl.Hostname(), wellKnownCerts)
+					if err != nil {
+						return err
+					}
+				} else {
+					o.Print("WARNING: no certificate authority provided for server, continuing but login will likely fail")
+				}
+			}
+		}
+	} else if !isServerTrusted && o.Cert != "" {
```
andrewpmartinez, line 228: Do we need options for automated clients that what to force or skip some of the prompts/checks in this code?

ziti/cmd/ziti/util/certs.go
plorenz: Ah, the name changed, but the bools didn't. Fixing

plorenz: fixed

CHANGELOG.md
plorenz: Fixed

ziti/cmd/ziti/cmd/edge_controller/login.go
plorenz: Not sure. Automated clients can always provide their own certs, which bypasses everything. I could also add a --yes flag.

plorenz: Added --yes flag

ziti/cmd/ziti/util/certs.go
plorenz: Fixed: Added ordering before compare

ziti/cmd/ziti/cmd/edge_controller/login.go
plorenz: Fixed: Added verify after retrieval

andrewpmartinez (approved):

CHANGELOG.md
```diff
@@ -32,6 +56,107 @@ Now we'll checking for links in Pending state, and if they haven't hit a configu
...
+### Identities
+The Ziti CLI now suports multiple identities. An identity can be specified using `--cli-identity` or `-i`. 
+
+Example commands:
+```
+$ ziti edge login -i dev localhost:1280
+Enter username: admin
+Enter password: 
+Token: 76ff81b4-b528-4e2c-ad73-dcb0a39b6489
+Saving identity 'dev' to ~/.config/ziti/ziti-cli.json
```
dovholuknf, line 80: should we decide on 'a standard' to follow? `~/.config/ziti` or `~/.ziti'`?

CHANGELOG.md
```diff
@@ -32,6 +56,107 @@ Now we'll checking for links in Pending state, and if they haven't hit a configu
...
+$ ziti edge -i cust1 logout  
+Removing identity 'cust1' from ~/.config/ziti/ziti-cli.json
+```
+
+#### Read-Only Mode
+When logging in one can mark the identity as read-only. This is a client side enforced flag which will attempt to make sure only
+read operations are performed by this session. 
+
+```
+$ ziti edge login --read-only localhost:1280
```
dovholuknf, line 132: should read-only be the default? force people to write mode?

CHANGELOG.md
```diff
@@ -32,6 +56,107 @@ Now we'll checking for links in Pending state, and if they haven't hit a configu
...
+$ ziti edge list services
+id: -JucPW0kGR    name: ssh    encryption required: true    terminator strategy: smartrouting    role attributes: ["ssh"]
+results: 1-1 of 1
+
+$ ziti edge create service test
+error: this login is marked read-only, only GET operations are allowed
+```
+
+NOTE: This is not guaranteed to prevent database changes. It is meant to help prevent accidental changes, if the wrong profile 
+is accidentally used. Caution should always be excercized when working with sensitive data!
```
dovholuknf, line 147: excercized  -> exercised ?

CHANGELOG.md
plorenz: It uses `.config/ziti` on linux, since that's linux standard, and `.ziti` elsewhere. It can be overridden with ZITI_HOME.

CHANGELOG.md
plorenz: I think most people probably want it in RW mode. Especially for new developers who are trying out the system I'd hate to put a roadblock in their way as they are experimenting. 

One thing that might make sense is it use read-only by default if you are using --token, since that implies a workflow where you've been granted temporary access to a system.

Thoughts?

CHANGELOG.md
plorenz: fixed

PR #459, opened 2021-08-06
ziti-fabric fix and changelog update

plorenz:


ziti-fabric/subcmd/root.go
```diff
@@ -80,6 +81,7 @@ var updateCmd = &cobra.Command{
 
 func Execute() {
 	if err := Root.Execute(); err != nil {
-		fmt.Printf("error: %s", err)
+		fmt.Printf("error: %s\n", err)
+		os.Exit(-1)
```
dovholuknf, line 85: shouldn't this be positive?

plorenz: yes, looks like. I got used to System.exit(-1) in Java (not sure it was correct their either, but seemed common (ah, here's some context: https://stackoverflow.com/questions/2434592/difference-in-system-exit0-system-exit-1-system-exit1-in-java))

plorenz: Fixed

dovholuknf (approved):

PR #467, opened 2021-08-18
do not truncate controller path from args

andrewpmartinez:
If a url + path was supplied to `ziti edge login`, the path would be truncated. The result would be a login attempt against the root path of the host.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,13 +3,14 @@
 ## What's New
 
 * Bug fix: Fabric v0.16.93 fixes `xgress.GetCircuit` to provide a `ctrl not ready` error response when requests arrive before the router is fully online.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,13 +3,14 @@
...
 
 * Bug fix: Fabric v0.16.93 fixes `xgress.GetCircuit` to provide a `ctrl not ready` error response when requests arrive before the router is fully online.
+* Bug fix: `ziti edge login` will no longer truncate paths on controller URLs
 
 # Release 0.22.0
 
 ## What's New
 
 * Refactor: Fabric Sessions renamed to Circuits (breaking change)
-* Feature: Links will now wait for a timeout for retrying
```
andrewpmartinez, line 12: My editor auto-removed dead whitespace at the end of lines.

dovholuknf (approved):

PR #477, opened 2021-08-31
fix float64 types in Westworld3 example profile

qrkourier:


doc/transwarp_b1/transwarp_b1.md
```diff
@@ -80,7 +80,7 @@ The `westworld3` protocol includes support for _profiles_, which allow the behav
 	tx_portal_min_sz                16384
 	tx_portal_max_sz                4194304
 	tx_portal_increase_thresh       224
-	tx_portal_increase_scale        1
+	tx_portal_increase_scale        0.05
```
michaelquigley, line 83: Shouldn't these still be 1... i.e... `1.0`?

qrkourier: @michaelquigley Probably. I blindly borrowed the 0.05 value from another occurrence of the same property in documentation. I'll change to 1.0.

michaelquigley (approved):
:+1:

PR #481, opened 2021-09-03
use Buster until we troubleshoot Bullseye

qrkourier:


plorenz (approved):

scareything:
Do problems crop up with bullseye, or is this just a cautious change?

PR #482, opened 2021-09-03
Update deps

plorenz:


ziti/cmd/ziti/cmd/edge/login.go
```diff
@@ -92,15 +92,21 @@ func (o *loginOptions) Run() error {
 
 	var host string
 	if len(o.Args) == 0 {
-		if defaultId := config.Identities[id]; defaultId != nil && !o.IgnoreConfig {
-			host = defaultId.Url
-			o.Printf("Using controller url: %v from identity '%v' in config file: %v\n", host, id, configFile)
-		} else {
-			if host, err = term.Prompt("Enter controller host[:port] (default localhost:1280): "); err != nil {
+		defaultHost := "localhost:1280"
```
andrewpmartinez, line 95: Why would we ever want this as a default? Seems odd. I know it matches our example configs, but this is wierd.

andrewpmartinez: The more I think about this the more I don't like it. It is nice for dev, but what if someone using ziti doesn't use 1280 as their default and they use something else. What if some other software uses 1280 and it causes odd errors.

plorenz: If we're not promoting it a standard port then I'll remove it

plorenz: fixed

ziti/cmd/ziti/cmd/edge/login.go
```diff
@@ -74,8 +74,8 @@ func newLoginCmd(out io.Writer, errOut io.Writer) *cobra.Command {
...
 	cmd.Flags().StringVarP(&options.Cert, "cert", "c", "", "additional root certificates used by the Ziti Edge Controller")
 	cmd.Flags().BoolVar(&options.ReadOnly, "read-only", false, "marks this login as read-only. Note: this is not a guarantee that nothing can be changed on the server. Care should still be taken!")
-	cmd.Flags().BoolVarP(&options.Yes, "yes", "y", false, "If set, responds to prompts with yes. This will result in untrusted certs being accepted or updated.")
-	cmd.Flags().BoolVar(&options.IgnoreConfig, "ignore-config", false, "If set, does not use value from the config file for hostname or username. Values must be entered or will be prompted for.")
+	cmd.Flags().BoolVarP(&options.Yes, "yes", "y", false, "if set, responds to prompts with yes. This will result in untrusted certs being accepted or updated.")
+	cmd.Flags().BoolVar(&options.DefaultConfig, "default-config", false, "if set, uses the value from the config file for hostname or username. If false, values must be entered or will be prompted for.")
```
andrewpmartinez, line 78: I ran this locally to see what this meant. It seems like the original operation eases the common path: opting into using an alternate controller host:port.  The way it works now, the user would have to always specify `--default-config` to use the host:port they configured in the first place.

Not sure if I am missing something or if there is a use case I am not accounting for.

andrewpmartinez: If the goal is to allow the user to use and alternate host:port or username/password with a specific identity adding separate flags for those overrides might work. We already have username/password. Adding `--host <host:port>` should suffice.

plorenz: Here are the three scenarios:

1. You've got a default. You have to confirm it by pressing enter, or you can enter a different value
2. You've got no default. You have to enter a value
3. You've got a default, and specified  --default-config. Your value is used without requiring a prompt.

This came up because both Clint and I have not noticed that the default wasn't expected. This just gives you a chance to confirm that the default is accurate.


```
plorenz@carrot:~/work/nf/ziti$ ze login
Enter controller host[:port] (default https://localhost:1280/edge/management/v1): 
Enter username (default 'admin'): 
Enter password: 
Token: 44e64766-a260-4afc-810a-7a897128e80d
Saving identity 'default' to /home/plorenz/.config/ziti/ziti-cli.json
plorenz@carrot:~/work/nf/ziti$ ze logout
Removing identity 'default' from /home/plorenz/.config/ziti/ziti-cli.json

plorenz@carrot:~/work/nf/ziti$ ze login
Enter controller host[:port]: localhost:1280
Enter username: admin
Enter password: 
Token: a9248596-1079-48e6-ab8b-3ef89880242d
Saving identity 'default' to /home/plorenz/.config/ziti/ziti-cli.json

plorenz@carrot:~/work/nf/ziti$ ze login --default-config
Using controller url: https://localhost:1280/edge/management/v1 from identity 'default' in config file: /home/plorenz/.config/ziti/ziti-cli.json
Using username: admin from identity 'default' in config file: /home/plorenz/.config/ziti/ziti-cli.json
Enter password: 
Token: 04cf144d-7fee-46df-968c-db4e8118d5d5
Saving identity 'default' to /home/plorenz/.config/ziti/ziti-cli.json
plorenz@carrot:~/work/nf/ziti$ 
``

File: ziti/cmd/ziti/cmd/edge/login.go [hunk 3]
```diff
@@ -92,15 +92,25 @@ func (o *loginOptions) Run() error {
...
+		if defaultId := config.Identities[id]; defaultId != nil && defaultId.Url != "" {
+			defaultHost = defaultId.Url
+			if o.DefaultConfig {
+				host = defaultId.Url
+				o.Printf("Using controller url: %v from identity '%v' in config file: %v\n", host, id, configFile)
+			}
+		}
+
+		if host == "" {
+			prompt := "Enter controller host[:port]: "
```
andrewpmartinez, line 105: Prompt asks for `host:port` but then attempts to parse the input as a URL and fails. 

I'm seeing this on windows as well:


```
09-03 16:44:36 (Grubby) ziti [update-deps ≡]> ziti edge login -i weeee
Enter controller host[:port]: localhost:1280
error: invalid controller URL: parse "https://localhost:1280\r": net/url: invalid control character in URL
09-03 16:44:44 (Grubby) ziti [update-deps ≡]> ziti edge login -i weeee
Enter controller host[:port]: https://localhost:1280
error: invalid controller URL: parse "https://localhost:1280\r": net/url: invalid control character in URL
09-03 16:44:53 (Grubby) ziti [update-deps ≡]>
```

andrewpmartinez: For `host:port` validation there is `validateHostPortString` in the edge controller.

ziti/cmd/ziti/cmd/edge/login.go
andrewpmartinez: Maybe I am not seeing the use case work because of the other issue I commented on having.

```
09-03 16:44:53 (Grubby) ziti [update-deps ≡]> ziti edge login
Enter controller host[:port] (default https://localhost:1280/edge/management/v1):
error: invalid controller URL: parse "https://\r": net/url: invalid control character in URL
```

ziti/cmd/ziti/cmd/edge/login.go
plorenz: Added some whitespace trimming which should fix your issue

ziti/cmd/ziti/cmd/edge/login.go
andrewpmartinez: If I could just press enter, I am not sure I like having to confirm the default for scenario 1. I would output that it is being used, its value, and then login.

Having to say yes use the default is striking me as odd.

ziti/cmd/ziti/cmd/edge/login.go
andrewpmartinez: It does thank you!

ziti/cmd/ziti/cmd/edge/login.go
```diff
@@ -139,11 +149,26 @@ func (o *loginOptions) Run() error {
...
+				defaultUsername = defaultId.Username
+				if o.DefaultConfig {
+					o.Username = defaultId.Username
+					o.Printf("Using username: %v from identity '%v' in config file: %v\n", o.Username, id, configFile)
+				}
+			}
+
+			if o.Username == "" {
+				prompt := "Enter username: "
+				if defaultUsername != "" {
```
andrewpmartinez, line 164: It says there is a default but when I press enter it prompts me over and over:

```
09-03 16:54:33 (Grubby) ziti [update-deps ≡]> ziti edge login
Enter controller host[:port] (default https://localhost:1280/edge/management/v1):
Enter username (default 'admin'):
Enter username (default 'admin'):
Enter username (default 'admin'):
```

ziti/cmd/ziti/cmd/edge/login.go
andrewpmartinez: If I want to connect to multiple different controller, the user shouldn't think about the controller host:port, it would be the identity right? There is a "default" which most use for everything - which I think is a problem.

With SSH you have to always specify the host:port you want to go to.

With `ziti edge login` that used to be also true, but now we have the "identity" concept. Should the format be changed to:

`ziti edge login <identity>` ?

andrewpmartinez: The choice of which identity you are using then determines your controller.

This way you can have multiple identities per controller.

The other way, you specify your controller and override which identity you wish.

andrewpmartinez: Mmm I see how you go to where we are. I am not sure I like any of this.

plorenz: I'm going to undo my login changes so it doesn't hold up the rest of the PR. Adding the space trim to prompt, here: https://github.com/openziti/foundation/pull/265

andrewpmartinez (approved):

PR #527, opened 2021-11-02
continue adding arm64 cross-builds for Docker

qrkourier:


qrkourier (commented):
@scareything @plorenz This was necessary because there's still a bug in Qemu that presents when running post-install scripts for the libc-bin package on Bullseye (aliases "stable", Debian 11, and latest). We weren't seeing this for the last few weeks because we were installing the 32bit version of libc-bin on arm64, but now that we have 64bit builds I removed that conditional from the Dockerfile. Now that we're running 64bit packages and `ziti-tunnel` on arm64 we have to use Debian 10 "Buster" until [this bug](https://github.com/moby/qemu/issues/19) is fixed.

plorenz (approved):

PR #531, opened 2021-11-05
fixes missing arg in pfx filter call

andrewpmartinez:


dovholuknf (approved):

ziti/cmd/ziti/cmd/log_filter.go
```diff
@@ -66,6 +67,6 @@ func (o *LogFormatOptions) Run() error {
 	if o.absoluteTime {
 		options.SetAbsoluteTime()
 	}
-	pfxlog.Filter(options)
+	pfxlog.Filter(os.Stdin, options)
```
andrewpmartinez, line 70: os.Stdin was the default, now it is an argument

PR #533, opened 2021-11-08
Add getziti.ps1

dovholuknf:
this adds a very simple script that will:
* download the latest ziti 
* place it into the folder of your choice (defaults to %userprofile%\.ziti\bin)
* adds ziti to your path if it's not there already (and if you 'source' the ps1)

### DIFF
--- a/quickstart/getZiti.ps1
+++ b/quickstart/getZiti.ps1
@@ -31,13 +31,12 @@ if($toDir.Trim() -eq "") {
 
 $zipFile="${toDir}\${name}"
 if($(Test-Path -Path $zipFile -PathType Leaf)) {
-    echo "The file has already been downloading. No need to download again"
+    Write-Output "The file has already been downloading. No need to download again"
 } else {
-    echo "The file does not exist"
     mkdir "${toDir}" -ErrorAction SilentlyContinue
-    echo "Downloading file "
-    echo "    from: ${downloadUrl} "
-    echo "      to: ${zipFile}"
+    Write-Output "Downloading file "
+    Write-Output "    from: ${downloadUrl} "
+    Write-Output "      to: ${zipFile}"
     iwr ${downloadUrl} -OutFile "$zipFile"
 }
 
@@ -46,7 +45,3 @@ Expand-Archive -Path $zipFile -DestinationPath "${toDir}\${version}" -ErrorActio
 if(! $env:Path.Contains("${toDir}\${version}\ziti\")){
     $env:Path+=";${toDir}\${version}\ziti\"
 }
-
-
-
-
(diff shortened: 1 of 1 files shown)


quickstart/getZiti.ps1
```diff
@@ -0,0 +1,32 @@
+$latestFromGitHub=(irm https://api.github.com/repos/openziti/ziti/releases/latest)
```
andrewpmartinez, line 1: Would you mind providing auto help?

```
<#
.SYNOPSIS

Adds a file name extension to a supplied name.

.DESCRIPTION

Adds a file name extension to a supplied name.
Takes any strings for the file name or extension.

.PARAMETER Name
Specifies the file name.

.PARAMETER Extension
Specifies the extension. "Txt" is the default.

.INPUTS

None. You cannot pipe objects to Add-Extension.

.OUTPUTS

System.String. Add-Extension returns a string with the extension
or file name.

.EXAMPLE

PS> extension -name "File"
File.txt

.EXAMPLE

PS> extension -name "File" -extension "doc"
File.doc

.EXAMPLE

PS> extension "File" "doc"
File.doc

.LINK

http://www.fabrikam.com/extension.html

#>
```

quickstart/getZiti.ps1
```diff
@@ -0,0 +1,32 @@
...
+$toDir=$(Read-Host "Where folder should be used for ziti? [default: ${defaultfolder}]")
+if($toDir.Trim() -eq "") {
+    $toDir=("${defaultfolder}")
+}
+
+$zipFile="${toDir}\${name}"
+if($(Test-Path -Path $zipFile -PathType Leaf)) {
+    echo "The file has already been downloading. No need to download again"
+} else {
+    echo "The file does not exist"
```
andrewpmartinez, line 36: As a heads up `echo` is just an alias for `Write-Output`. I prefer to not use echo because if someone has installed a GNU toolkit, echo might not be the echo scripts are looking for.

`Write-Output` has its output pipelined (meaning you can pass it to another command).

There is `Write-Host` which provides output but will not be piped.

quickstart/getZiti.ps1
```diff
@@ -0,0 +1,32 @@
...
+    mkdir "${toDir}" -ErrorAction SilentlyContinue
+    echo "Downloading file "
+    echo "    from: ${downloadUrl} "
+    echo "      to: ${zipFile}"
+    iwr ${downloadUrl} -OutFile "$zipFile"
+}
+
+Expand-Archive -Path $zipFile -DestinationPath "${toDir}\${version}" -ErrorAction SilentlyContinue
```
andrewpmartinez, line 44: You can use `$LASTEXITCODE` to check for command errors.

Highly useful when using silent error actions. Also the default error action can be set at the start of a script. I've seen that used with try/catch styles.

andrewpmartinez (approved):

PR #537, opened 2021-11-09
make express install more clear

dovholuknf:
* removed foolish 'unused' variables and redirect to /dev/null instead 
* add startExpressEdgeRouter function
* don't emit systemd files by default since those won't work on mac/wsl
* reference start functions at the end to show someone how to start/stop ziti locally

output should end up looking like:
image


scareything (approved):

plorenz:
nice!

PR #539, opened 2021-11-10
Update deps/changelog

plorenz:


etc/ctrl.yml
```diff
@@ -134,3 +134,66 @@ mgmt:
   #influxdb:
     #url:                http://localhost:8086
     #database:           ziti
+    #
+    
+
+# web - optional
```
dovholuknf, line 140: was this file just never updated with the new stuff, right?

plorenz: yes, and since it will become required soon with the fabric rest API, I though I should add it in

dovholuknf: oh - this is the "no edge" config file... maybe this file should just be deleted or the other one should be mv'ed onto this file name?

plorenz: I think it makes sense to have both. There's maybe an argument that with edge is the most common case, so edge.yml should be with edge and we could have a ctrl.fabric.only.yml.

plorenz: I don't think it should matter much to people who aren't developing ziti though. It should be easy to create your own config easily.

dovholuknf (approved):

PR #543, opened 2021-11-18
Add support for ziti fabric operations

plorenz:


ziti/cmd/ziti/cmd/edge/login.go
```diff
@@ -186,17 +187,17 @@ func (o *loginOptions) Run() error {
...
 	}
 
-	loginIdentity := &util.RestClientIdentity{
+	loginIdentity := &util.RestClientEdgeIdentity{
 		Url:       host,
 		Username:  o.Username,
 		Token:     o.Token,
 		LoginTime: time.Now().Format(time.RFC3339),
-		Cert:      o.Cert,
+		CaCert:    o.Cert,
```
dovholuknf, line 195: this change sticks out to me - either o.Cert should be o.CaCert or this should not be CaCert?

ziti/cmd/ziti/cmd/fabric/add_identity.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func newAddIdentityCmd(p common.OptionsProvider) *cobra.Command {
+	options := &addIdentityOptions{
+		CommonOptions: p(),
+	}
+
+	cmd := &cobra.Command{
+		Use:   "add-identity <server-url>",
+		Short: "adds a fabric identity for connecting to the Ziti Controller",
```
dovholuknf, line 40: is a fabric identity different than an 'edge' identity different than a 'controller' identity?

dovholuknf (approved):

ziti/cmd/ziti/cmd/fabric/add_identity.go
plorenz: A fabric identity is signed client cert recognized by the controller. An edge identity is an api session token. The controller itself has an identity, which consists of a cert chain.

ziti/cmd/ziti/cmd/edge/login.go
plorenz: Switched o.Cert to o.CaCert. The edge identity only has one cert, whereas the fabric identity has the signed client cert plus the ca cert. So since I had to distinguish them, made sense to make it more consistent

PR #544, opened 2021-11-18
adds ziti-fabric fingerprint command

andrewpmartinez:


ziti-fabric/subcmd/set_fingerprint.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if routerId == "" {
+		pfxlog.Logger().Fatalf("<router-id> must specified")
+	}
+
+	if fingerprint == "" {
+		pfxlog.Logger().Fatalf("<fingerprint> must specified and be a valid hex characters")
+	}
+
+	db, err := bbolt.Open(dbFile, 0666, nil)
```
dovholuknf, line 43: nits: 666 vs 600? or 660? why not use the config file from the controller to find it instead of trusting the user?

ziti-fabric/subcmd/set_fingerprint.go
```diff
@@ -0,0 +1,92 @@
...
+	"strings"
+)
+
+func init() {
+	Root.AddCommand(setRouterFingerprintCmd)
+}
+
+var setRouterFingerprintCmd = &cobra.Command{
+	Use:   "fingerprint <db-file> <router-id> <fingerprint>",
+	Short: "Manually specify the router client fingerprint, the controller must not be running.",
```
dovholuknf, line 19: if the controller is running it'll just block forever?

dovholuknf (approved):
just confirm the test.ctrl.db was intentionally added to the commit. since it's not referenced here i dunno for sure

ziti-fabric/subcmd/set_fingerprint.go
dovholuknf, line 43: i don't suspect you need to close the store but 'should you'?

plorenz (commented):
Do we need the test DB? 

I'd like to deprecate ziti-fabric, once we've got `ziti fabric`. Maybe we want a ziti-db tool, or similar? I've got a debug-db tool that opens a DB and a adds an admin user, so I can debug customer databases. Seems like those could go together. 

In the short term it's fine in ziti-fabric

ziti-fabric/subcmd/set_fingerprint.go
andrewpmartinez: The DB lock can't be acquired and it will fail.

andrewpmartinez:
> just confirm the test.ctrl.db was intentionally added to the commit. since it's not referenced here i dunno for sure

needs to be removed

ziti-fabric/subcmd/set_fingerprint.go
andrewpmartinez: IIRC this is for if the file will be created - in this code it won't do much. Can be whatever you want honestly.

andrewpmartinez: Uhh sure.

plorenz:
You can now update router fingerprints with `ziti fabric update router <nameOrId> --fingerprint <fingerprint>`

PR #546, opened 2021-11-19
log and advertise available update for each Ziti component

qrkourier:


ziti-tunnel/cmd/ziti-tunnel/main.go
```diff
@@ -29,5 +36,59 @@ func init() {
 
 func main() {
 	debugz.AddStackDumpHandler()
+	logReleaseVersionCheck()
 	subcmd.Execute()
 }
+
+func logReleaseVersionCheck() {
```
dovholuknf, line 43: you could make this function take the constant that represents the name of the binary....

or use os.Args[0] and get the name of the executable and then it'd be fully dynamic.. something like:

import "path/filepath" and then use:
```
	exeName := filepath.Base(os.Args[0])
```

then you can put this whole function into the util pacakge and it'd be trivial to use from other CLI commands, right?

dovholuknf: if you go that route though, make sure you test what the behavior is when you rename the exe to something like `ztun` ... i expect this check stops working then. 

also any chance you could show the various outputs of these as screen caps? I'm somewhat too lazy to pull/build test the permutations but i'm interested to see what each look like

qrkourier: warn: current build is unreleased development build

update available: current build is superseded by latest

info: current build is latest

qrkourier: If changed the log level for these messages from INFO and WARN to DEBUG so they will not normally be printed unless `--verbose`.

qrkourier: preview

ziti-fabric-test/main.go
```diff
@@ -48,5 +51,6 @@ func init() {
 
 func main() {
 	rand.Seed(info.NowInMilliseconds())
+	util.LogReleaseVersionCheck(constants.ZITI_FABRIC_TEST)
```
dovholuknf, line 54: probably don't need the check here?

ziti-fabric/main.go
```diff
@@ -18,6 +18,7 @@ package main
```
dovholuknf, line 18: probably can revert this change too?

dovholuknf:
should this get added to `ziti-router` too?

qrkourier:
@plorenz pointed out that it's important to avoid corrupting stdout when `-j` (JSON output) is selected. I'll have to explore ways to switch based on that condition.

qrkourier:
I'm thinking either:
1. notice JSON mode and disable the version check
2. always send version check output to stderr

qrkourier:
@dovholuknf :point_right: [demo video](https://youtu.be/hP-7rEmej_0)

qrkourier:
@dovholuknf Actually, this still isn't quite ready. I think we need to take a closer look at the logrus level. Presently, the new function that checks and logs the update status uses info level because only subcommands have `--verbose` options and those don't take effect until after this function is invoked. 

This means that sending the update log to debug level will never be seen. We could log the message after the subcommand runs by ordering the invocation later in `main()` which I believe would inherit the subcmd log level, but that severely reduces visibility of the message which is intended as a preamble to any troubleshooting efforts.

We could add an option to the root command to control the log level before the subcmd is called e.g. `--verbose`, and we could pass it through to the subcmd if it's set on the root command e.g. `ziti-tunnel --verbose version` (set on root) vs `ziti-tunnel version --verbose` (set on subcmd).

qrkourier:
This is redundant:
```bash
❯ ./build/ziti version
INFO[0000] this build of ziti is unreleased v0.0.0      
NAME             VERSION
ziti             v0.0.0
ziti-controller  not installed
ziti-fabric      v0.0.0
ziti-prox-c      not installed
ziti-router      not installed
ziti-tunnel      v0.18.8
ziti-edge-tunnel v0.17.20
```
This is too much information. It should be debug level...but there's no way for the user to enable debug level.
```bash
❯ ./build/ziti edge
INFO[0000] this build of ziti is unreleased v0.0.0      
Interact with Ziti Edge Components

Usage:
  ziti edge [command]
...
```

qrkourier:
@dovholuknf I know we talked about _not_ changing the formatting for imports in this PR to avoid the "dueling linters" problem, but in this case the changes are suggested by Google's Go language server, not just any linter. Shouldn't we accept these formatting changes as a standard in that case? If not then I'll try to disable the language server's formatting behaviors in my IDE, a task that has proven somewhat obscure.

qrkourier:
> @dovholuknf I know we talked about _not_ changing the formatting for imports in this PR to avoid the "dueling linters" problem, but in this case the changes are suggested by Google's Go language server, not just any linter. Shouldn't we accept these formatting changes as a standard in that case? If not then I'll try to disable the language server's formatting behaviors in my IDE, a task that has proven somewhat obscure.

@dovholuknf This is resolved now that the upstream has adopted the same approach to formatting. My changes are no longer a difference.

PR #558, opened 2021-12-03
Start of ctrl logs parser and support for inspect in ziti fabric

plorenz:


ziti-ops/logs/common.go
```diff
@@ -127,6 +127,17 @@ type JsonLogsParser struct {
...
+	afterTime                   string
+	include                     LogMatcher
+}
+
+func (self *JsonLogsParser) addCommonArgs(cmd *cobra.Command) {
+	cmd.Flags().DurationVarP(&self.bucketSize, "interval", "n", time.Hour, "Interval for which to aggregate log messages")
+	cmd.Flags().IntVarP(&self.maxUnmatchedLoggedPerBucket, "max-unmatched", "u", 1, "Maximum unmatched log messages to output per bucket")
+	cmd.Flags().StringSliceVar(&self.ignore, "ignore", nil, "Filters to ignore")
+	cmd.Flags().StringVarP(&self.beforeTime, "before", "B", "", "Process only messages before this timestamp")
+	cmd.Flags().StringVarP(&self.afterTime, "after", "A", "", "Process only messages after this timestamp")
```
andrewpmartinez, line 140: I don't know why this convention didn't occur to me earlier. For the ziti CLIs we can use upper case letters for root/shared options and lower case for command-specific options. Or vice versa.

andrewpmartinez (approved):

ziti-ops/logs/common.go
plorenz: Oh, yeah. Good idea. I only used those because it matches grep :)

PR #571, opened 2022-01-05
add and document Linux builder

qrkourier:
This documents how developers may build Linux executables for amd64, arm, arm64 with the same parameters used by the OpenZiti project's GitHub Actions workflow.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - main
       - release-v*
-      - builx-action # TODO: prune after initial testing
   pull_request:
   workflow_dispatch:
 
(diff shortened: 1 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -73,7 +72,7 @@ jobs:
 
   tag-and-publish:
     name: Build Linux Binaries and Docker Image, Tag, and Publish
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
dovholuknf, line 75: will this cause a problem with glibc? i think it will? i'd feel better if we kept this at 18

dovholuknf (approved):

.github/workflows/main.yml
dovholuknf: actually would you mind reverting this for now? thx. i forgot about this comment and merged it... :)

PR #582, opened 2022-01-31
Controller.address.move

andrewpmartinez:


ekoby (approved):

plorenz (commented):
autoformat has wreaked havoc here. I stopped used goland for editing release notes b/c I couldn't get the formatter to behave. It's markdown, so probably doesn't matter too much

plorenz (approved):

PR #596, opened 2022-02-10
Revert gox commands until we figure out why docker build is failing

plorenz:


andrewpmartinez:
Very weird. Those commands appear very similar.

plorenz:
> I'm going to guess that the backslashes aren't actually escaping the newlines, so the CC variable is set for a single subprocess that does nothing, and then the `gox` command runs (without CC in its environment)

Yeah, I was suspicious of those backslashes as well

PR #619, opened 2022-02-28
Router allow traversal

camotts:


ziti/cmd/ziti/cmd/edge/create_edge_router.go
```diff
@@ -38,6 +38,8 @@ type createEdgeRouterOptions struct {
 	tags              map[string]string
 	appData           map[string]string
 	cost              uint16
```
plorenz, line 40: I don't think you need both of these flags. If you've got no traversal, and it defaults to false, they can achieve either outcome. Also, you can be explicit and do --no-traversal=true, --no-traversal=false

camotts: I must be going crazy, I thought I checked the explicit setting `--traversal=false` (this was before I added the two flags) and it was not firing the changed events. Testing again and it 100% does. Reverting!

plorenz (commented):
Looks good

plorenz (approved):

PR #621, opened 2022-02-28
Remove local constants to use global constants

gberl002:
Updated default value references to the new global values


gberl002:
This will fail until the dependent project PRs are merged.

dovholuknf (approved):

PR #624, opened 2022-03-01
Release 0.24.12

plorenz:


.dockerignore
```diff
@@ -1 +1 @@
-/release/
+#/release/
```
dovholuknf, line 1: was this intentional? i don't remember it - but figured i'd point it out :)

dovholuknf (approved):

.dockerignore
plorenz: This is a change @qrkourier made to fix the docker release workflow

qrkourier: Yes, it's to workaround a bug in Docker BuildKit. They're already fixed it upstream so we'll inherit that eventually through the GitHub Action we're using.

PR #626, opened 2022-03-02
Fix missing config values

gberl002:


quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -956,6 +956,8 @@ function ziti_createEnvFile {
 
   export ZITI_BIN_ROOT="${ZITI_HOME}/ziti-bin"
 
+  if [[ "${ZITI_EDGE_ROUTER_RAWNAME-}" == "" ]]; then export ZITI_EDGE_ROUTER_RAWNAME="${ZITI_NETWORK}-edge-router"; fi
+
```
gberl002, line 960: Adding this fixes the config generation since the env var was never exported however, it triggers the express install to state that an overridden ZITI_EDGE_ROUTER_RAWNAME value is being used.

I wasn't sure how this should be handled, should we remove the "if value set then state it's overridden" and just expect it to always be set?

dovholuknf (approved):

PR #645, opened 2022-03-14
Fixing advertised port and listener port, added related tests

### LINKED ISSUE #643: expressInstall - controller.web.port wrong
When using expressInstall for 'host it myself' you are supposed to be able to set `export ZITI_EDGE_CONTROLLER_PORT=8441`. This works great for all the ports in the controller config file . See how address is overridden properly (:8441) but the listen interface is wrong (:1280)

```
web:
  # name - required
  # Provides a name for this listener, used for logging output. Not required to be unique, but is highly suggested.
  - name: client-management
    # bindPoints - required
    # One or more bind points are required. A bind point specifies an interface (interface:port string) that defines
    # where on the host machine the webListener will listen and the address (host:port) that should be used to
    # publicly address the webListener(i.e. mydomain.com, localhost, 127.0.0.1). This public address may be used for
    # incoming address resolution as well as used in responses in the API.
    bindPoints:
      #interface - required
      # A host:port string on which network interface to listen on. 0.0.0.0 will listen on all interfaces
      - interface: 0.0.0.0:1280
        # address - required
        # The public address that external incoming requests will be able to resolve. Used in request processing and
        # response content that requires full host:port/path addresses.
        address: bobby.com:8441
```

gberl002:
closes #643


ziti/cmd/ziti/cmd/create_config.go
```diff
@@ -83,6 +83,7 @@ type ControllerTemplateValues struct {
 }
 
 type EdgeControllerValues struct {
+	Port            string
```
dovholuknf, line 86: Port -> AdvertisedPort? also i see `AdvertisedHostPort` still used below. Seems like we need an AdvertisedHost / ListnerHost as well?

dovholuknf (approved):

PR #651, opened 2022-03-18
Add m1 support

### LINKED ISSUE #627: allow quickstart/expressInstall to provide what version of ziti to use
right now the quickstart/expressInstall function/ziti-cli-scripts.sh will get the LATEST version of ziti. if any problems happen with 'latest' you can't currently go back to 'previous' versions. it'd be nice to be able to set the version number of ziti you DO want to use so in the rare times the latest is 'broken' somehow you can move back.

this would also allow one to make two different networks at two different versions for testing reasons etc

gberl002:
closes #627

### DIFF
--- a/quickstart/docker/image/ziti-cli-functions.sh
+++ b/quickstart/docker/image/ziti-cli-functions.sh
@@ -135,8 +135,10 @@ function getLatestZitiVersion {
   fi
 
   ZITI_ARCH="amd64"
-  if [[ "$(uname -a)" == *"arm"* ]]; then
+  if [[ "$(uname -a)" == *"arm"* ]] && [[ "${ZITI_OSTYPE}" != "darwin" ]]; then
     ZITI_ARCH="arm"
+  else
+    echo -e "$(YELLOW "WARN: It has been detected that you are using an Apple computer with ARM architecture. Deployment of Apple ARM architecture distributions is currently unsupported through git, the installer will pull darwin amd distribution instead.")"
   fi
 
   unset ZITI_BINARIES_VERSION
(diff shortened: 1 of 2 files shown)


gberl002:
Since it was a simple change, I incorporated issue #627

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -127,31 +127,42 @@ function checkHostsFile {
   fi
 }
 
-function getLatestZitiVersion {
+function getZitiVersion {
```
dovholuknf, line 130: why this name change? I'm not averse to it but i'd just rather keep these sorts of changes to a minimum. i wonder if anyone else uses this. i think i used it on some blog / article at one point... better to jsut keep it. i'd prefer to provide a "fetch this exact version of ziti" function ( i see below you are allowing for the version override )

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -342,7 +383,37 @@ function ziti_expressConfiguration {
     echo " "
     if [[ "${REPLY}" == [Yy]* ]]; then
       echo -e "$(GREEN "Clearing existing Ziti variables and continuing with express install")"
+
```
gberl002, line 386: This was necessary in order to handle errors "nicely". If you already have ziti version A installed and you try to install version B, the database version mismatches and throws an error when trying to start up the controller.

In trying to keep with the "quickstart" mindset I felt that (even though this is a lot of extra logic) it would make the end user experience easier by not requiring them to read the log and guess what was going wrong. It also makes the end user experience better in the event they want to reinstall but keep the specific version (which would be wiped out by the `unsetZitiEnv` function.

With this being said, feedback on alternative solutions is welcomed.

dovholuknf (approved):

PR #654, opened 2022-03-30
Quickstart rolling image.  Adding support for role attributes on quickstart routers

tburtchell:
1. ubuntu:latest doesn't point to real latest.   Updating to the rolling tag which seems to be more what we want.
2. I needed to be able to add role attributes to edge routers to support a more complex environment.  Adding that ability in case it's useful for others.  Edge routers by default will get a role attribute of the router name.  Set environment variable `ZITI_EDGE_ROUTER_ROLES` to override.

### DIFF
--- a/quickstart/docker/image/Dockerfile
+++ b/quickstart/docker/image/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:latest
+FROM ubuntu:rolling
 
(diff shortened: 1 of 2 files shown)


dovholuknf (approved):
fyi - @gberl001 nice update to allow adding roles when starting the router via docker

PR #662, opened 2022-04-01
Fix docker quickstart

gberl002:


quickstart/docker/simplified-docker-compose.yml
```diff
@@ -19,9 +19,11 @@ services:
       - "/openziti/scripts/run-controller.sh"
 
   ziti-edge-router:
-    image: openziti/quickstart:latest
+    image: "${ZITI_IMAGE}:${ZITI_VERSION}"
     environment:
       - ZITI_EDGE_ROUTER_RAWNAME=ziti-edge-router
+    depends_on:
+      - ziti-controller
```
dovholuknf, line 26: "${ZITI_CONTROLLER_RAWNAME}" ?

quickstart/docker/simplified-docker-compose.yml
```diff
@@ -37,6 +39,8 @@ services:
     environment:
       - ZAC_SERVER_CERT_CHAIN=/openziti/pki/ziti-controller-intermediate/certs/ziti-controller-server.cert
       - ZAC_SERVER_KEY=/openziti/pki/ziti-controller-intermediate/keys/ziti-controller-server.key -it
+    depends_on:
+      - ziti-controller
```
dovholuknf, line 43: can you use a var here too? (ZITI_CONTROLLER_RAWNAME)

quickstart/docker/simplified-docker-compose.yml
```diff
@@ -37,6 +39,8 @@ services:
     environment:
       - ZAC_SERVER_CERT_CHAIN=/openziti/pki/ziti-controller-intermediate/certs/ziti-controller-server.cert
       - ZAC_SERVER_KEY=/openziti/pki/ziti-controller-intermediate/keys/ziti-controller-server.key -it
```
dovholuknf, line 41: this ` -it` appears to be a stray copy/paste error? can you clean that up if yes?

quickstart/docker/buildLocalDev.sh
```diff
@@ -1,9 +1,29 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
+echo "${SHELL}"
+
+myrealpath() {
```
dovholuknf, line 6: comment as to 'why' this function exists in case anyone wonders?

quickstart/docker/simplified-docker-compose.yml
dovholuknf: nevermind -- i see this is the container reference

quickstart/docker/image/run-router.sh
```diff
@@ -12,6 +12,7 @@ if [[ "${ZITI_EDGE_ROUTER_PORT-}" == "" ]]; then export ZITI_EDGE_ROUTER_PORT="3
 if [[ "${ZITI_EDGE_ROUTER_HOSTNAME}" == "" ]]; then export ZITI_EDGE_ROUTER_HOSTNAME="${ZITI_EDGE_ROUTER_RAWNAME}${ZITI_DOMAIN_SUFFIX}"; fi
 if [[ "${ZITI_EDGE_ROUTER_ROLES}" == "" ]]; then export ZITI_EDGE_ROUTER_ROLES="${ZITI_EDGE_ROUTER_RAWNAME}"; fi
 
+generateEnvFile
```
dovholuknf, line 15: not sure about this. i think we need to be careful here - this caused a race condition for the docker-compose environment. i had removed this for that reason.

gberl002: Yeah, I recall this being removed I thought it was a race condition between the controller and router causing the conflict so I had added the depends_on. I added this back because there was an env var not getting set in the controller startup that was needed. I don't recall off hand what that was but I'll take a look today and see if I can work around it and remove this again.

PR #665, opened 2022-04-04
add config hints in ctrl template

qrkourier:


ziti/cmd/ziti/cmd/config_templates/controller.yml
```diff
@@ -119,6 +119,10 @@ web:
         # address - required
         # The public address that external incoming requests will be able to resolve. Used in request processing and
         # response content that requires full host:port/path addresses.
+        #
+        # This ziti CLI login example assumes this interface binding is 127.0.0.1:8443 and "binding: edge-management" is in the list of bound APIs below
```
dovholuknf, line 123: this feels like the wrong file to put `ziti` commands into. i'm curious why you'd want to add this here?

qrkourier: That was the idea from the thread, to show the alignment between these settings and how they influence operation of the APIs configured here.

dovholuknf: We don't need ziti cli examples in the config. it's just too specific for this section. There's numerous other cli commands that need the same API - calling out ziti login just doesn't fit. Also the comment just above covers the config entry adequately.

ziti/cmd/ziti/cmd/config_templates/controller.yml
```diff
@@ -27,7 +27,7 @@ ctrl:
   #connectTimeoutMs:       {{ .Controller.DefaultConnectTimeout.Milliseconds }}
   listener:             tls:{{ .Controller.ListenerAddress }}:{{ .Controller.Port }}
 
-# the endpoint that management tools connect to the controller over.
+# the deprecated GRPC endpoint for legacy management tools (see below for Edge mgmt REST API)
```
dovholuknf, line 30: I'd let someone else make this particular change. @plorenz - can we make this change can call it deprecated?

plorenz: The streaming operations are still unavailable via other mechanisms and most of the operations that have been deprecated have been removed already. Also, it's not GRPC :) You could change the description to something like: 'endpoint that can be used to for some streaming operations, such as streaming circuit events and metrics'.

ziti/cmd/ziti/cmd/config_templates/controller.yml
```diff
@@ -27,7 +27,7 @@ ctrl:
   #connectTimeoutMs:       {{ .Controller.DefaultConnectTimeout.Milliseconds }}
   listener:             tls:{{ .Controller.ListenerAddress }}:{{ .Controller.Port }}
 
-# the endpoint that management tools connect to the controller over.
+# endpoint that can be used to for some streaming operations, such as streaming circuit events and metrics (see 'web' for Edge mgmt REST API)
```
gberl002, line 30: There seems to be a grammar issue here. ("can be used to for"), I think the "to" should be removed.

qrkourier (commented):
ready

qrkourier:
> ready

ziti/cmd/ziti/cmd/config_templates/controller.yml
```diff
@@ -27,7 +27,7 @@ ctrl:
   #connectTimeoutMs:       {{ .Controller.DefaultConnectTimeout.Milliseconds }}
   listener:             tls:{{ .Controller.ListenerAddress }}:{{ .Controller.Port }}
```
plorenz, line 29: I'm going to remove the mgmt section in the next release

qrkourier:
this template hint is moot because the section is being removed

PR #687, opened 2022-04-19
Update CLI. Add stream commands to ziti fabric

plorenz:


CHANGELOG.md
```diff
@@ -11,6 +11,10 @@
 * Bug Fix: Identity property `externalId` was not properly rendering on `GET` and not handled consistently on `PUT` and `PATCH`
 * Enhancement: External JWT Signer Issuer & Audience Validation
 * Enhancement: Add ability to define local interface binding for link and controller dial
+* Enhancement: The mgmt API is now available via websocket. The stream commands are now available on `ziti fabric`
```
andrewpmartinez, line 14: Is this the entire management API or just a subset to support streaming?

ziti/cmd/ziti/cmd/api/json.go
```diff
@@ -89,3 +110,62 @@ func (self *GabsWrapper) StringSlice(path string) []string {
 	}
 	return nil
 }
+
+type Gabs2Wrapper struct {
```
andrewpmartinez, line 114: An aside from this PR's goal is:

I want to reduce the usage of gabs with generated structures. Gabs is great when you don't know the JSON format, but if it is a contract-based object, I'm against using it as it creates scenarios where things can break and no one notices till an end-user reports the issue.

plorenz: Makes sense. For the next pass I'll see about converting to using the generated rest client

CHANGELOG.md
plorenz: So, technically it's all of it, but we've already removed almost everything except streaming operations. I'll remove the last couple of non-streaming operations when I update the fabric tests to use the REST API instead. I'll probably also leave the inspect API since it's convenient to be able to run those asynchronously.

PR #704, opened 2022-05-03
Update to go 1.18

plorenz:


go.mod
```diff
@@ -1,22 +1,6 @@
 module github.com/openziti/ziti
 
-go 1.17
-
-//replace github.com/openziti/foundation => ../foundation
```
dovholuknf, line 5: did you intend to remove all these commented out lines? i think no?

dovholuknf: oh - maybe go 1.18 has a better way? :)

plorenz: Check out go workspaces: https://go.dev/blog/get-familiar-with-workspaces introduced in Go 1.18. I've been using them for a month or so, and they're very convenient. I started off with a single go.work in my root directory but then moved to a go.work per project. So a go.work for ziti if you were working with in-flight work on fabric and edge, might look like:

```
go 1.18

use (
        ./
        ../fabric
        ../edge
)

```

dovholuknf (approved):

PR #723, opened 2022-05-19
get token for ziti-tunnel enroll in container from env or stdin

qrkourier:
The goal of this PR is to update `ziti-tunnel enroll` behaviors in a container to parity with `ziti-edge-tunnel enroll` which already has env and stdin token sources. The env token source in particular enables deploying `ziti-tunnel host` on an edge compute platform that do not afford the ability to supply a token file in the volume that is mounted on the container to persist the identity config file.

### DIFF
--- a/ziti-tunnel/docker/README.md
+++ b/ziti-tunnel/docker/README.md
@@ -10,7 +10,11 @@ when the one-time enrollment token is consumed.
 
 ## Variables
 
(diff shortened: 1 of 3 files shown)


qrkourier (commented):
ready for reviewers

plorenz (approved):

PR #727, opened 2022-05-23
updates changelog

andrewpmartinez:


andrewpmartinez:
> Do we want a `ziti edge re-enroll identity` or similar?

I'm going to add it but this lets the MOP move forward.

PR #757, opened 2022-06-08
Append build number to smoketest version

qrkourier:


.github/workflows/main.yml
```diff
@@ -155,8 +158,13 @@ jobs:
             # fail the job because we could not obtain the current version from ziti-ci
             echo "ERROR: ZITI_VERSION=${ZITI_VERSION}"
             exit 1
```
plorenz, line 160: This needs to include `main` as well

plorenz (approved):

PR #771, opened 2022-06-22
Update deps and changelog

plorenz:


etc/ctrl.with.edge.yml
```diff
@@ -75,10 +75,10 @@ events:
...
 #      - type: fabric.routers
 #      - type: fabric.terminators
-      - type: metrics
-        sourceFilter: .*
-        metricFilter: .*pool.*
-#      - type: fabric.circuits
+#      - type: metrics
+#        sourceFilter: .*
+#        metricFilter: .*pool.*
+      - type: fabric.circuits
```
andrewpmartinez, line 81: Was this intentional?

andrewpmartinez (approved):

etc/ctrl.with.edge.yml
plorenz: Now that we have the config generation, I haven't been keeping these files as tidy. Not sure if it matters. I'll try and tidy it back up on the next update

PR #783, opened 2022-07-07
comment the if out while testing

dovholuknf:


tburtchell (approved):

qrkourier (approved):
Normally I'd delete the line instead of commenting because when it goes to review the question is always 
> Shouldn't we just delete this?

PR #790, opened 2022-07-13
Release 0.26.0

plorenz:


plorenz:
> Change looks good. Version get fail just a fluke?

It's b/c I removed the tag from the last version. Adding a temporary tag until the release is made

plorenz:
> > Change looks good. Version get fail just a fluke?
> 
> It's b/c I removed the tag from the last version. Adding a temporary tag until the release is made

Actually wasn't the removed tag. Turns out this was down to some newish code that relies on getting the current version number in the workflow, and that breaks when we update the minor number. This just happens to be the first time we're bumping the minor since that code was added. I updated ziti-ci correctly report the current build number when we bump the minor.

PR #808, opened 2022-07-22
Simple transfer

camotts:
Added small smokiest using ziti/fablab


.github/workflows/main.yml
```diff
@@ -39,13 +48,22 @@ jobs:
           $(go env GOPATH)/bin/ziti-ci generate-build-info common/version/info_generated.go version
           $(go env GOPATH)/bin/gox -cgo -os=darwin -arch=amd64 -output=$GOX_OUTPUT ./...
```
plorenz, line 50: Can we skip building test artifacts on mac and windows, since they're probably not used in the smoketest?

camotts: Sure can! Left those in to make new tests environments one less thing to think about but easy enough to add back later.

ziti-probe/subcmd/run.go
```diff
@@ -5,14 +5,6 @@ package subcmd
 import (
```
plorenz, line 5: Let's just delete ziti-probe for now. We can resurrect it from git history if we need.

plorenz:
Thinking about it, it might make sense to put the smoketest run to happen concurrently with the windows and mac builds. That we're adding less time to the build and we won't move on to the linux build/publish unless the smoketest passes. Thoughts?

camotts:
> Thinking about it, it might make sense to put the smoketest run to happen concurrently with the windows and mac builds. That we're adding less time to the build and we won't move on to the linux build/publish unless the smoketest passes. Thoughts?

I might be terribly misunderstanding something. Right now I think you're talking about the new fable smoketests? I don't think we can run that concurrently with the windows/Mac build before the linux build. The linux build are the artifacts we are testing against. Or do you mean separate out the publish step?

plorenz:
> > Thinking about it, it might make sense to put the smoketest run to happen concurrently with the windows and mac builds. That we're adding less time to the build and we won't move on to the linux build/publish unless the smoketest passes. Thoughts?
> 
> I might be terribly misunderstanding something. Right now I think you're talking about the new fable smoketests? I don't think we can run that concurrently with the windows/Mac build before the linux build. The linux build are the artifacts we are testing against. Or do you mean separate out the publish step?

We could just build the linux artifacts twice, since building them doesn't take that long. I would guess it's faster to rebuild them in the final step than to upload and download them. 
There's an argument about making sure what you're testing against is the same as what your publishing, but they should be identical since dependencies and code are locked down. Builds should be reproducible.

camotts:
> > > Thinking about it, it might make sense to put the smoketest run to happen concurrently with the windows and mac builds. That we're adding less time to the build and we won't move on to the linux build/publish unless the smoketest passes. Thoughts?
> > 
> > 
> > I might be terribly misunderstanding something. Right now I think you're talking about the new fable smoketests? I don't think we can run that concurrently with the windows/Mac build before the linux build. The linux build are the artifacts we are testing against. Or do you mean separate out the publish step?
> 
> We could just build the linux artifacts twice, since building them doesn't take that long. I would guess it's faster to rebuild them in the final step than to upload and download them. There's an argument about making sure what you're testing against is the same as what your publishing, but they should be identical since dependencies and code are locked down. Builds should be reproducible.

I don't mind building twice if you don't. I agree that it should be the same, if somehow they're different I think we have bigger problems to fix

.github/workflows/main.yml
```diff
@@ -117,6 +126,143 @@ jobs:
           CC=arm-linux-gnueabihf-gcc $(go env GOPATH)/bin/gox -cgo -os=linux -arch=arm -output=$GOX_OUTPUT ./...
           CC=aarch64-linux-gnu-gcc   $(go env GOPATH)/bin/gox -cgo -os=linux -arch=arm64 -output=$GOX_OUTPUT ./...
 
+          mkdir -p test/amd64/linux/bin
+          go build -tags=all,tests -o test/amd64/linux/bin ./...
```
plorenz, line 130: Can drop arm builds

.github/workflows/main.yml
```diff
@@ -117,6 +126,143 @@ jobs:
...
+          CC=arm-linux-gnueabihf-gcc $(go env GOPATH)/bin/gox -cgo -os=linux -arch=arm -output=$GOX_TEST_OUTPUT -tags=all ./...
+          CC=aarch64-linux-gnu-gcc   $(go env GOPATH)/bin/gox -cgo -os=linux -arch=arm64 -output=$GOX_TEST_OUTPUT -tags=all ./...
+
```
plorenz, line 133: Can drop upload artifacts . Can also drop computing the ziti version. That's done below in the publish task

.github/workflows/main.yml
```diff
@@ -117,6 +126,143 @@ jobs:
...
+          elif [[ "${GITHUB_REF}" =~ ^refs/heads/(release-v|main$) ]]; then
+            # Set output parameters for release branches
+            echo "DEBUG: ZITI_VERSION=${ZITI_VERSION}"
+            echo ::set-output name=ziti-version::${ZITI_VERSION}
+          else
+            # Append build / run number for non-release refs
+            ZITI_VERSION="${ZITI_VERSION}-${GITHUB_RUN_ID}"
+            echo "DEBUG: ZITI_VERSION=${ZITI_VERSION}"
+            echo ::set-output name=ziti-version::${ZITI_VERSION}
+          fi
```
plorenz, line 164: Because we're only using the artifacts for test, could just build them right where you want them above in the go build commands

.github/workflows/main.yml
```diff
@@ -117,6 +126,143 @@ jobs:
...
+      - name: Move artifacts to bin subdir
+        run: |
+          mv -v test/amd64/linux/bin/* release/amd64/linux/
+          cd release/amd64/linux
+          mkdir bin
+          shopt -s extglob
+          mv -- !(bin) bin
+          chmod +x bin/*
+
```
plorenz, line 173: Optional, could just have these in a file in the test area and then copy that file to where it should live

.github/workflows/main.yml
```diff
@@ -117,6 +126,143 @@ jobs:
...
+          go install github.com/mitchellh/gox@latest
+          $(go env GOPATH)/bin/gox -cgo -os=linux -arch=amd64 -output=$GOX_OUTPUT ./...
+          CC=arm-linux-gnueabihf-gcc $(go env GOPATH)/bin/gox -cgo -os=linux -arch=arm -output=$GOX_OUTPUT ./...
+          CC=aarch64-linux-gnu-gcc   $(go env GOPATH)/bin/gox -cgo -os=linux -arch=arm64 -output=$GOX_OUTPUT ./...
+
+          mkdir -p test/amd64/linux/bin
+          go build -tags=all,tests -o test/amd64/linux/bin ./...
+          CC=arm-linux-gnueabihf-gcc $(go env GOPATH)/bin/gox -cgo -os=linux -arch=arm -output=$GOX_TEST_OUTPUT -tags=all ./...
+          CC=aarch64-linux-gnu-gcc   $(go env GOPATH)/bin/gox -cgo -os=linux -arch=arm64 -output=$GOX_TEST_OUTPUT -tags=all ./...
+
```
plorenz, line 251: Don't need upload, we're building them here

camotts: Artifacts from a bad copy/paste that I glossed over

.github/workflows/main.yml
camotts: Can definitely move it to a file if that would be preferable. Wasn't sure how often this would be edited

.github/workflows/main.yml
camotts: Forgot to remove that upload. The test does need ziti version, I had left in the second one in because since we we're rebuilding everything anyway just incase. Can remove the second one if you'd like

.github/workflows/main.yml
camotts: dropped

plorenz (approved):

PR #837, opened 2022-09-16
add script to add router quickly

jlin-nf:
Add a quickstart script to create ER external to controller.

### DIFF
--- a/quickstart/docker/image/add-router.sh
+++ b/quickstart/docker/image/add-router.sh
@@ -1,0 +1,182 @@
+#!/bin/bash
+
+# For edge router setup we will need to address some external variables.
+# this section needs to move to doucmentation.
+#export EXTERNAL_IP="$(curl -s eth0.me)"       
+#export ZITI_EDGE_CONTROLLER_IP_OVERRIDE="127.0.0.1"
+#export ZITI_EDGE_ROUTER_IP_OVERRIDE="${EXTERNAL_IP}"
+#export ZITI_EDGE_CONTROLLER_PORT=1280
+#export ZITI_EDGE_CONTROLLER_MANAGEMENT_PORT=6262
+#export ZITI_EDGE_ROUTER_PORT=443
+#export ZITI_EDGE_ROUTER_LAN_PORT="ens3"
+#export ZITI_EDGE_ROUTER_NAME="ER_NEWNAME"
+
+
+# For ease of use, we will introduce an optional parameter for controller ip.
+if [ $# -eq 0 ]; then
+   if [[ "${ZITI_EDGE_CONTROLLER_IP_OVERRIDE}" != "" ]]; then
+      controller_ip=${ZITI_EDGE_CONTROLLER_IP_OVERRIDE}
+   else
+      controller_ip="127.0.0.1"
+   fi
+elif [ $# -eq 1 ]; then
+   # user specified controller ip
+   controller_ip=$1
+else
+   echo ""
+   echo "Usage:"
+   echo "     $0 <controller-ip>"
+   exit
+fi
+
+if [[ "${ZITI_EDGE_CONTROLLER_PORT}" != "" ]]; then
+   ctrl_port=${ZITI_EDGE_CONTROLLER_PORT}
+else
+   ctrl_port="1280"
+fi
+
+if [[ "${ZITI_EDGE_CONTROLLER_MANAGEMENT_PORT}" != "" ]]; then
+   mgmt_port=${ZITI_EDGE_CONTROLLER_MANAGEMENT_PORT}
+else
+   mgmt_port="6262"
+fi
+
+if [[ "${ZITI_EDGE_ROUTER_LAN_PORT}" != "" ]]; then
+   lan_port=${ZITI_EDGE_ROUTER_LAN_PORT}
+else
+   lan_port="ens3"
+fi
+
+if [[ "${ZITI_EDGE_ROUTER_PORT}" != "" ]]; then
+   rport=${ZITI_EDGE_ROUTER_PORT}
+else
+   rport="443"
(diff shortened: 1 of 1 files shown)


quickstart/docker/image/add-router.sh
```diff
@@ -0,0 +1,182 @@
...
+   er_name=${ZITI_EDGE_ROUTER_NAME}
+else
+   er_name="none"
+fi
+
+router_ip=$(ip addr show ${lan_port} | grep "inet\b" | awk '{print $2}' | cut -d/ -f1)
+
+echo Use ${router_ip} as internal IP
+
+cat <<EOT >config.yml
```
dovholuknf, line 66: you should probably be using `ziti create config router edge|private` here instead of emitting the config file. see the ziti-cli-functions.sh for examples if interested like: https://github.com/openziti/ziti/blob/release-next/quickstart/docker/image/ziti-cli-functions.sh#L770

quickstart/docker/image/add-router.sh
```diff
@@ -0,0 +1,182 @@
...
+   # user specified controller ip
+   controller_ip=$1
+else
+   echo ""
+   echo "Usage:"
+   echo "     $0 <controller-ip>"
+   exit
+fi
+
+if [[ "${ZITI_EDGE_CONTROLLER_PORT}" != "" ]]; then
```
dovholuknf, line 32: when you try out `ziti create router config`, it'll do a lot or all of these

quickstart/docker/image/add-router.sh
```diff
@@ -0,0 +1,182 @@
...
+sed -i "s/ctrl_port/$mgmt_port/" config.yml
+
+#update config template with name of lan interface for new router
+sed -i "s/lan_port/$lan_port/" config.yml
+
+if [ -f "ziti/ziti" ]
+then
+   echo "ziti binary found skipping download!"
+else
+   zitilatest=$(curl -s https://api.github.com/repos/openziti/ziti/releases/latest)
```
dovholuknf, line 146: `ziti-cli-functions.sh` provides a `getZiti` and `getLatestZitiVersion`. any improvements should probably land in there

quickstart/docker/image/add-router.sh
```diff
@@ -0,0 +1,182 @@
...
+   echo ${BBlue}curl -OL https://github.com/openziti/ziti/releases/download/v${version}/ziti-linux-amd64-${version}.tar.gz${NC}
+   curl -OL https://github.com/openziti/ziti/releases/download/v${version}/ziti-linux-amd64-${version}.tar.gz
+   #untar ziti
+   echo ''
+   echo ${BBlue}tar zxvf ziti-linux-amd64-${version}.tar.gz${NC}
+   tar zxvf ziti-linux-amd64-${version}.tar.gz
+fi
+
+#login to ziti cli
+ziti/ziti edge login "${controller_ip}:${ctrl_port}" -u "admin" -p "admin" -y
```
dovholuknf, line 159: `ziti-cli-functions.sh`, or the .env file after running the quickstart adds a `zitiLogin` alias that works with the quickstarts. i think it's worthwhile to look into that and how it works. it sources an env file and uses the user/pwd specified in there too. if we ever randomize the user/pwd or let the user supply it during the quickstart, you'll be more future proof i think

quickstart/docker/image/add-router.sh
```diff
@@ -0,0 +1,182 @@
...
+
+if [ -f "ziti/ziti" ]
+then
+   echo "ziti binary found skipping download!"
+else
+   zitilatest=$(curl -s https://api.github.com/repos/openziti/ziti/releases/latest) 
+   version=$(echo "${zitilatest}" | tr '\r\n' ' ' | jq -r '.tag_name' |tr -d v)
+      #download ziti
+   echo ''
+   echo ${BBlue}curl -OL https://github.com/openziti/ziti/releases/download/v${version}/ziti-linux-amd64-${version}.tar.gz${NC}
```
dovholuknf, line 150: i don't see ${BBlue} declared yet? `ziti-cli-function.sh` at the top declares a bunch of ASCI escape codes to use though that you might want to use or you could use the functions in there like:
```
function BLUE {
  echo "${ASCI_BLUE}${1-}${ASCI_RESTORE}"
}
```

example of that being used here: https://github.com/openziti/ziti/blob/release-next/quickstart/docker/image/ziti-cli-functions.sh#L60

quickstart/docker/image/add-router.sh
```diff
@@ -0,0 +1,182 @@
+#!/bin/bash
```
dovholuknf, line 1: i'd recommend you just update the big, ziti-cli-functions.sh if needed. I think much/most maybe all the things you're wanting to do already exists as functions in that script. for example there is already a `createEdgeRouterConfig "${ZITI_EDGE_ROUTER_RAWNAME}"`

quickstart/docker/image/add-router.sh
jlin-nf: yes, Geoff was able to add environment variable to set external IP to SANs (for ziti create config).  Also would need to have the external IP in the advertisement section. (I believe that is been worked on right now).

quickstart/docker/image/add-router.sh
jlin-nf: OK

quickstart/docker/image/add-router.sh
jlin-nf: Will also try to work that in there.

quickstart/docker/image/add-router.sh
jlin-nf: thanks.. i will put this into the main script, so the color etc. are defined.

PR #860, opened 2022-10-12
preallocate memory

florianl:
**What this PR does / why we need it:**

Preallocate memory instead of enforcing an incremental growth. This will result in less work for the garbage collector.


plorenz (approved):
Looks good, thank you. At some point these should be moved to foundation and made generic

PR #871, opened 2022-10-19
update the local dev article

qrkourier:


etc/ctrl.yml
```diff
@@ -12,16 +12,16 @@ v: 3
 # Define all of the paths to the certificates and key required to establish the controller's identity.
 #
 identity:
-  cert:                 etc/ca/intermediate/certs/ctrl-client.cert.pem
-  server_cert:          etc/ca/intermediate/certs/ctrl-server.cert.pem
-  key:                  etc/ca/intermediate/private/ctrl.key.pem
-  ca:                   etc/ca/intermediate/certs/ca-chain.cert.pem
+  cert: ${ZITI_SOURCE}/ziti/etc/ca/intermediate/certs/ctrl-client.cert.pem
```
qrkourier, line 15: I duplicated the value of `identity` from the other provided controller config `ctrl.with.edge.yml`. My goal is to express a convention across the controller configs so that it's simple to guide the reader to initialize their shell environment in a way that works for all of the sample configs. 

There are a couple of downsides to this however, and I'm wondering if we can change the convention: `${ZITI_SOURCE}/ziti/`. Downsides include:
1. This assumes `ZITI_SOURCE` is the _parent_ directory of the checked-out copy. This is a little unexpected and requires explanation. It would be more intuitive if this was the repo copy itself, not the parent dir. There doesn't seem to be any requirement to share data between repos and so it seems like we could optimize for a self-contained experience of this repo alone.
2. This assumes the directory name of this repo's checked out copy is exactly "ziti", which is not a safe assumption. Is there a broader reason today why devs must _always_ check out to the default name "ziti"? If so, then we should explain that reasoning, and if not then we should only operate on files within this repo.

andrewpmartinez: The configs in `ziti/etc` have always been meant for local development that dictates the project be checkout in a specific manner. The way they are set up is because it caused the least amount of friction with our local development when environment variable injection was introduced to the configuration files.

If you want, you can make a set of configurations that are specifically targeted at quickstart, articles, etc. I personally would like you to not change the existing dev configs.

Additionally, the ziti configuration files for the router and controller allow for any environment variable to be used in the `${...}` format. So if you wish, when you create your configuration files for your purposes with whatever environment variables you wish.

qrkourier: Thanks @andrewpmartinez . It not necessary to change anything in `ziti/etc` with this PR and I've reverted the one change I had made there to the fabric-only controller config. 

My main goal with this PR is to add a few pointers and hints to guide the person who has newly cloned this repo and wishes to build from source and run a minimal, local Ziti stack they can use for development.

doc/002-local-dev.md
```diff
@@ -1,184 +1,207 @@
...
+# build the binary without installing in GOPATH
+go build -o ./build/ziti-router ./ziti-router/
+# execute the binary
+./build/ziti-router version
 ```
 
-You will find the built artifacts in `./release`.
+## Crossbuilds
+
+When you push to your repo fork then GitHub Actions will automatically crossbuild for several OSs and CPU architectures. You'll then be able to download the built artifacts from the GitHub UI. The easiest way to crossbuild the Linux exectuables locally is to build and run the crossbuild container. Please refer to [the crossbuild container README](../Dockerfile.linux-build.README) for those steps. For hints on crossbuilding for MacOS and Windows see [the main GitHub Actions workflow](../.github/workflows/main.yml) which defines the steps that are run when you push to GitHub.
```
dovholuknf, line 62: this should all be 'anywhere else'. This isn't relevant to "local development". This all feels way more related to publishing than local dev

doc/002-local-dev.md
```diff
@@ -1,184 +1,207 @@
...
+
+## Run a Local Ziti Stack
+
+Let's get a local Ziti stack up and running now that you have built and installed all the Ziti apps in this repo.
+
+### Initialize the Shell Environment
+
+You'll need to export two environment variables that are employed by the included controller and router configuration YAML files. These are needed in each shell where you'll run a controller or router process. You could manage them all as background processes in a single shell, but it's easier to use separate shells for each if only to have a dedicated view for the log messages of each process.
```
dovholuknf, line 70: this is all bash-centric and i'd personally prefer to just remove all this too. Keep ZITI_SOURCE and ZITI_DATA hidden for now imo.

doc/002-local-dev.md
```diff
@@ -1,184 +1,207 @@
...
 Each configuration file refers to certificate and private keys kept in `etc/ca/intermediate/certs` and
-`etc/ca/intermediate/private/`. The steps for starting the router is to first register the router then 
-start it.
+`etc/ca/intermediate/private/`.
 
 Register:
 
 ```bash
-ziti-fabric create router etc/ca/intermediate/certs/XXX-client.cert.pem
+ziti fabric create router etc/ca/intermediate/certs/SERIAL-client.cert.pem
```
dovholuknf, line 165: should be ziti edge create router imo and it's just easier to copy/paste 1-4 so all i need to do is copy one big block

dovholuknf: and actually - it should just be 1. this is for someone learning how to dev. they don't need 4 routers out of the gate. just givem one. let paul or andrew add permutations/complications like "a full fabric" etc

doc/002-local-dev.md
```diff
@@ -1,184 +1,207 @@
...
 
-There is only one example configuration file for an edge router: 
+There is only one example configuration file for an edge router:
 `etc/edge.router.yml`. Additional configuration files can be created by copying and altering the
 file. Specifically the identity section needs to point to unique file locations that do not collide with other identity
-file and the listening ports need to not be in use.
+files, and the listening ports need to be available to bind.
 
-### Authenticate w/ the Ziti Edge API
+## Ziti Edge REST API Primer
```
dovholuknf, line 193: i don't want you to put it there, but this really belongs in the edge repo imo. not here.  Alternaltively you can just remove this and all the rest from here down since its' all about using ziti not dev'ing on ziti. it doesn't belong

doc/002-local-dev.md
qrkourier: We can do that. It means adding a new controller config that doesn't use these two env vars. It might also mean a new .gitignore statement to allow `ZITI_DATA` to be a relative path inside this repo's checked-out copy.

dovholuknf (approved):

doc/002-local-dev.md
qrkourier: Moved to BUILD.md

doc/002-local-dev.md
qrkourier: We decided to generate the controller and router configs with `ziti` CLI instead of using the checked-in configs

PR #880, opened 2022-10-27
Added function for the same

avinilcode:
Please let review and let me know

### DIFF
--- a/quickstart/docker/image/ziti-cli-functions.sh
+++ b/quickstart/docker/image/ziti-cli-functions.sh
@@ -1471,4 +1471,17 @@ function getFileOverwritePermission() {
   fi
 }
 
+function checkIfPortForZitiEdgeControllerPortIsListening(){
+  ports=()
+  while IFS= read -r line; do
+    ports+=( "$line" )
+  done < <("sudo lsof -i:1280")
+  if ["${#ports[*]}" == 0]
+  then
+   return 0
+  else
+   return 1
+  fi
+}
+
 set +uo pipefail
--- a/quickstart/docker/image/ziti-cli-functions.sh
+++ b/quickstart/docker/image/ziti-cli-functions.sh
@@ -1471,7 +1471,7 @@ function getFileOverwritePermission() {
   fi
 }
 
-function checkIfPortForZitiEdgeControllerPortIsListening(){
+function isZitiEdgeControllerPortListening(){
   ports=()
   while IFS= read -r line; do
     ports+=( "$line" )
--- a/quickstart/docker/image/ziti-cli-functions.sh
+++ b/quickstart/docker/image/ziti-cli-functions.sh
@@ -255,6 +255,11 @@ function getZiti {
 }
 
 function checkPrereqs {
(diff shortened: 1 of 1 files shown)


quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -1471,4 +1471,17 @@ function getFileOverwritePermission() {
   fi
 }
 
+function isZitiEdgeControllerPortListening(){
```
dovholuknf, line 1474: let's make this a more generic "isPortAvailable" that takes a parameter of the port to check. Probabaly need to turn that logic around then as well and need to pass the port to be checked and not hardcode the 1280.  Also need to add this to the `checkPrereqs` function. where it does something like check "if !isPortAvailable print error and exit"

avinilcode: ok

avinilcode: Please review now

gberl002:
@avinilcode Thanks for working on this for us, this is a feature I've been wanting but haven't had time to implement it so I'm excited to see this work.

However, this doesn't seem to work for me. I tried on MacOS 12.5 and on Ubuntu 20.04. The first issue was that the syntax for the isPortAvailable call `!isPortAvailable: command not found`. I added a space between the exclamation and function call and that fixed that issue.

However, after that fix, I 'm seeing this output on both mac and linux
```
sudo lsof -i:1280: command not found
[0: command not found
Port 1280 is not available hence exiting
```
So, it seems that it doesn't like this line `if ["${#ports[*]}" == 0]` and I'm not sure why the first line as that should work and it does when running that command straight from the terminal.

Have you tested this locally to confirm it works for you and, if so, what OS are you testing this with?

avinilcode:
> @avinilcode Thanks for working on this for us, this is a feature I've been wanting but haven't had time to implement it so I'm excited to see this work.
> 
> However, this doesn't seem to work for me. I tried on MacOS 12.5 and on Ubuntu 20.04. The first issue was that the syntax for the isPortAvailable call `!isPortAvailable: command not found`. I added a space between the exclamation and function call and that fixed that issue.
> 
> However, after that fix, I 'm seeing this output on both mac and linux
> 
> ```
> sudo lsof -i:1280: command not found
> [0: command not found
> Port 1280 is not available hence exiting
> ```
> 
> So, it seems that it doesn't like this line `if ["${#ports[*]}" == 0]` and I'm not sure why the first line (the lsof not found) as that should work and it does when running that command straight from the terminal.
> 
> Have you tested this locally to confirm it works for you and, if so, what OS are you testing this with?

@gberl002 I am not able to run this, i tried the following "./ziti-cli-functions.sh isPortAvailable 1280" but there no output. Can you tell how to run it. And thanks for testing on my behalf.

gberl002:
@avinilcode since you were unable to run this locally, I went ahead and added this on a separate branch. That branch has since been merged so this is no longer needed. Thank you for taking a look at this.

PR #881, opened 2022-10-27
added exception if LAN_INTERFACE is not set with Tproxy option

dariuszSki:


dariuszSki:
@dovholuknf is this going to be approved anytime soon?

gberl002:
@dariuszSki I spoke with @scareything about the lanIf and setting up tproxy. He kindly pointed me to the documentation in the [release notes for 0.19.8](https://github.com/openziti/ziti/releases/tag/v0.19.8) which explains that `lanIf` is not required, if left blank it's simply not specified but you can put a value in there to specify an interface so it's totally optional.

It would be better if `ZITI_EDGE_ROUTER_LAN_INTERFACE` was a flag option on the CLI command, this way it's easily documented right in the list of command options.

With that said, we don't need this check for the lanIf values. So, I am going to close this PR. I'll rework this today to make it a flag option and put that PR up soon.

dariuszSki:
@gberl002 ok, sounds good. Will you rework all the changes in this branch to add tproxy switch to ```ziti edge config edge``` or just lanif?

PR #902, opened 2022-11-15
add function to create external ER

jlin-nf:
Setup environmental variables, then call:

source /dev/stdin <<< "$(wget -qO- https://raw.githubusercontent.com/openziti/ziti/release-next/quickstart/docker/image/ziti-cli-functions.sh)"; addRouter

### DIFF
--- a/quickstart/docker/image/ziti-cli-functions.sh
+++ b/quickstart/docker/image/ziti-cli-functions.sh
@@ -44,6 +44,11 @@ function BLUE {
 function zitiLogin {
   "${ZITI_BIN_DIR-}/ziti" edge login "${ZITI_EDGE_CTRL_ADVERTISED}" -u "${ZITI_USER-}" -p "${ZITI_PWD}" -c "${ZITI_PKI_OS_SPECIFIC}/${ZITI_EDGE_CONTROLLER_ROOTCA_NAME}/certs/${ZITI_EDGE_CONTROLLER_INTERMEDIATE_NAME}.cert"
 }
+
+function zitiLoginNoCert {
+  "${ZITI_BIN_DIR-}/ziti" edge login "${ZITI_EDGE_CTRL_ADVERTISED}" -u "${ZITI_USER-}" -p "${ZITI_PWD}" -y
+}
+
 function cleanZitiController {
   checkEnvVariable ZITI_HOME
   retVal=$?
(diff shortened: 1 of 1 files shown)


gberl002:
Just as an overall comment, I would prefer to see this utilize more of the existing code and functions in the script. Possibly just modifying the `createEdgeRouterConfig` since it is already used to create an edge router. You could modify that function to also check for all the environment variables since it won't always be run during express install but, during express install it should zip right past them since they'd already be set up, this way it's one reusable function.

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -379,6 +384,35 @@ function generateEnvFile {
   echo -e "environment file sourced from: $(BLUE "${ENV_FILE}")"
 }
 
+function generateRouterEnvFile {
```
gberl002, line 387: What would this file be used for in the future? Once the router is created, I'm not sure why we would need to refer to the env variables specific to a router unless we were going to recreate it but I'm not sure why someone would want to do that.

jlin-nf:
> Just as an overall comment, I would prefer to see this utilize more of the existing code and functions in the script. Possibly just modifying the `createEdgeRouterConfig` since it is already used to create an edge router. You could modify that function to also check for all the environment variables since it won't always be run during express install but, during express install it should zip right past them since they'd already be set up, this way it's one reusable function.

If i reuse or modify the existing function, it is very possible i will break something that is out that.

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -525,6 +559,180 @@ function ziti_expressConfiguration {
...
+  echo -e "        Controller API port: ${ZITI_EDGE_CONTROLLER_PORT}"
+  echo -e "    Edge Router IP and Port: ${ZITI_EDGE_ROUTER_IP_OVERRIDE}:${ZITI_EDGE_ROUTER_PORT}"
+  echo -e "                  Interface: ${main_interface}"
+  echo -e "                  ZITI-HOME: ${ZITI_HOME}"
+  echo -e "       ZITI Binary location: ${ZITI_BIN_DIR-}"
+  echo -e "                router-name: ${ername}"
+  echo -e ""
+}
+
+function ziti_expressRouter {
```
gberl002, line 624: This feels like going through the quickstart process again, I mentioned this in a main PR comment but would it be possible to modify `createEdgeRouterConfig` to create the router config rather than a new express install style function, I'm worried that with the reused variables and duplicated code is going to introduce a lot more points of failure.

jlin-nf: it was make to be similar to quickstart minus the creation of the controller.

the idea is the directory structures and etc. are similar to quickstart, so we can find config/env quickly.  It is also not lump into quickstart, in case there is bug, we can fix them easier.  At some point, we can try to consolidate functions if we choose to.

quickstart/docker/image/ziti-cli-functions.sh
gberl002: But for ziti login we already have `zitiLogin` and we have the credentials for that in the quickstart `.env` file, if we're now storing it here as well we're duplicating data and if something changes (like their password) and they forget to update it in one of the env files it could create a headache for a new user to troubleshoot.

And really, the only thing needed to run the router is the path to the yaml file for the router, it seems less complicated to just modify `startExpressEdgeRouter` to take a file path parameter or to create a more generic function `startRouter <path/name.yaml>`

quickstart/docker/image/ziti-cli-functions.sh
gberl002: The quickstart does many different things, downloading ziti, setting up a PKI, creating a controller, router, etc. I don't think all of this is necessary for just creating a router.

With this being essentially copied code, if there is a bug that we find we'll have to remember to fix it in both places. In addition, even just making an enhancement, we'll have to review all of this code too.

For instance, this entire first if statement is a copy of what's in express install. If we were going to use this in two places it should definitely be in a separate function and just referenced by each place calling that function. However, I'm not suggesting we do that as I don't think this first if statement, which checks if expressInstall was run, pertains to someone wanting to add a single new router. This would clear all of their env variables and they'd have to later re-source the main quickstart `.env` file which seems like a step backward.

quickstart/docker/image/ziti-cli-functions.sh
jlin-nf: this file will be on a node outside of controller.  this is for external edge-router.  the login function are different since it does not use the certs when login to the controller.

the password issue is something we will need to work out next. not sure there is a easier way to handle password change (from the initial passwd). and i am not sure about your comment wrt two places.  the zitiLogin uses environment variable ZITI_USER and ZITI_PWD, so if these were changed by user, when i dump the env file, it should save correctly whether it is on the original environment file or the new one i created for router.

quickstart/docker/image/ziti-cli-functions.sh
jlin-nf: this code  is adding router outside of the vm. so there should be one env file.

gberl002:
Closing as this is a duplicate, I took over work and unfortunately could not retain this code as this is a forked repository.

The new PR is https://github.com/openziti/ziti/pull/961

PR #903, opened 2022-11-15
Issue-706 Added port check to quickstart

### LINKED ISSUE #706: Add port check to quickstart
It would be nice to have a command to check if all the necessary quickstart related ports are available so the install doesn't fail. This could be either a command (that would be applied to `ziti-doc` repo) that's copy/pasted or even better, added to `expressInstall` and maybe a reference on how to change the port (in a case where the user cannot free it for the install).

gberl002:
Added ability to set listener bind port
Added tests for new port variable
Fixed bug in quickstart test
Fixed typo in quickstart test
fixed bug in quickstart env file test
closes #706


gberl002:
I would like to better understand the rules for committing go.mod and go.sum changes.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/Jeffail/gabs/v2 v2.6.1
 	github.com/MakeNowJust/heredoc v1.0.0
 	github.com/blang/semver v3.5.1+incompatible
+	github.com/docker/docker v20.10.21+incompatible
```
dovholuknf, line 10: i feel like we need a way to do these sorts of dependencies without checking them in. or we need a whole new project

gberl002: Yeah, I left a comment, I'm not sure how changing dependencies is managed so I would like to learn about that process.

quickstart/docker/image/run-with-ziti-cli.sh
```diff
@@ -29,7 +29,7 @@ if [[ -f "${initFile}" ]]; then
 fi
 
 # give the controller scripts time to start and create the ziti environment file if running in docker compose 
-until $(test -f "${ZITI_HOME}/ziti.env"); do echo "waiting for ziti.env..."; sleep 1; done
+until $(test -f "${ZITI_HOME}/ziti.env"); do echo "waiting for ziti.env..."; sleep 2; done
```
dovholuknf, line 32: just too much output, eh? on average how many times do you _actually_ see this? dunno - i don't see the need to add another second (on average) but... i mean  it's only a second so `¯\_(ツ)_/¯`

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -279,6 +279,36 @@ function checkPrereqs {
...
+  fi
+
+  envVar="${1-}"
+  if [[ -n "$ZSH_VERSION" ]]; then
+    envVarValue="${(P)envVar}"
+  elif [[ -n "$BASH_VERSION" ]]; then
+    envVarValue="${!envVar}"
+  else
+    echo -e "Unknown/Unsupported shell, cannot check port status"
+    return 1
```
dovholuknf, line 295: should this just return 0 in this case? i feel like you shouldn't fail for this particular reason. i'd just print a warning. i'd also print it in orange/yellow/red or something to get attention and say 'whatever, don't mess it up' from here.

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -279,6 +279,36 @@ function checkPrereqs {
   fi
 }
 
+function _portCheck {
+  if [[ "${1-}" == "" ]] || [[ "${2-}" == "" ]]; then
```
dovholuknf, line 283: i feel like it'd be helpful to inform the users that the configured ports are being checked for open-ness. it's also only checking that they are available LOCALLY, not 'remotely' (yet) but checking that the 'actual' firewall is open to the 'advertised' host would be a logical next addition here imo

ziti/cmd/ziti/cmd/create_config_environment.go
```diff
@@ -185,6 +186,7 @@ func NewCmdCreateConfigEnvironment() *cobra.Command {
 		"%-40s %-50s %s\n"+
 		"%-40s %-50s %s\n"+
 		"%-40s %-50s %s\n"+
+		"%-40s %-50s %s\n"+
```
dovholuknf, line 189: i feel like there's probably a 'more manageable' way of doing this?

ziti/cmd/ziti/cmd/create_config_router_edge_test.go
```diff
@@ -189,6 +190,51 @@ func TestBlankEdgeRouterNameBecomesHostname(t *testing.T) {
 
 }
 
+func TestDefaultZitiEdgeRouterListenerBindPort(t *testing.T) {
+	expectedDefaultPort := "10080"
```
dovholuknf, line 194: why isn't this a constant from somewhere?

dovholuknf: though - testing for it to be 10080 hardcoded like this is a good indicator that "doc" probably needs to change. so if that's why- makes sense

ziti/cmd/ziti/cmd/create_config_router_fabric_test.go
```diff
@@ -109,3 +110,48 @@ func TestFabricRouterOutputPathDoesNotExist(t *testing.T) {
 
 	assert.EqualError(t, err, expectedErrorMsg, "Error does not match, expected %s but got %s", expectedErrorMsg, err)
 }
+
+func TestDefaultZitiFabricRouterListenerBindPort(t *testing.T) {
+	expectedDefaultPort := "10080"
```
dovholuknf, line 115: but. i don't know if i would hardcode it more than one time (see above comment)

ziti/cmd/ziti/constants/constants.go
```diff
@@ -36,8 +36,8 @@ const (
 
 // Config Template Constants
 const (
-	DefaultListenerBindPort  = 10080
-	DefaultGetSessionTimeout = 60 * time.Second
+	DefaultZitiEdgeRouterListenerBindPort = "10080"
```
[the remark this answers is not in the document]
gberl002: It was mainly for future-proofing and to be more in line with the other environment variables, in case there was ever more than one type of router listener port. The environment variable that represents this value (when the default isn't used) is `ZITI_EDGE_ROUTER_LISTENER_BIND_PORT` and most other values in the portion of the config this is used start with `ZITI_EDGE_ROUTER_LISTENER...`

ziti/cmd/ziti/cmd/create_config_router_edge_test.go
gberl002: I believe the main reason I started hard coding values that are actually constants stored elsewhere was because some of these constants are stored in other projects so, if it was changed in another project, this test would fail allowing us to review any changes that might need to be made in this project... it was a discussion from a long time ago but I _think_ that was the main reason.

ziti/cmd/ziti/constants/constants.go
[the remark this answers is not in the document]
gberl002: Would you prefer changing the Go variable, both, or something else.

Again, I changed it only to resemble the env var that it represents, previously there was not env var which is why it was much more succinct.

dovholuknf (approved):

PR #909, opened 2022-11-21
quickstart shell function tweaks

qrkourier:
* use the short hostname because it's tidier and might fix a problem on older Linuxes.
    Arrived at this while troubleshooting a problem where the controller times out looking for its BoltDB. The path to ctrl.db was valid, but I guessed it might be too long for something in CentOS. Everything worked with the short hostname. The output of `$(hostname)` was a long EC2 FQDN. 
* use `pkill` instead of `killall` because it's more prevalent (`killall` missing in CentOS 7)


quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -138,14 +140,15 @@ function verifyZitiVersionExists {
 
   unset ZITI_BINARIES_VERSION
 
-  ziticurl="$(curl -s https://api.github.com/repos/openziti/ziti/releases/tags/"${ZITI_VERSION_OVERRIDE}")"
+  ziticurl="$(curl -s https://${GITHUB_TOKEN:+${GITHUB_TOKEN}@}api.github.com/repos/openziti/ziti/releases/tags/"${ZITI_VERSION_OVERRIDE}")"
```
qrkourier, line 143: Expands to empty string if not defined, otherwise expands as `${GITHUB_TOKEN}@`

qrkourier: I added this because I keep running into the GitHub rate limit which breaks `expressInstall`.

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -69,7 +71,7 @@ function startZitiController {
 }
 
 function stopZitiController {
-  killall ziti-controller
+  pkill ziti-controller
```
dovholuknf, line 74: really we need a 'mult-platform way' of doing this better than killall or pkill. Are either of those better/worse for Fedora (and any other *nix variant)? Is there some better way to do this sort of thing in general? capturing a pid on start for example ?

quickstart/docker/image/ziti-cli-functions.sh
dovholuknf: > I added this because I keep running into the GitHub rate limit which breaks expressInstall.

this is the sort of comment that fits best in the code imo

quickstart/docker/image/ziti-cli-functions.sh
qrkourier: Oh dang, you're right. `pkill` is missing in fedora:36. `killall` didn't exist in CentOS 7 is why I changed it. You're right, `kill` is the only truly universal because it's a BASH builtin. It's not hard to add PID handling here. I can do that.

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -998,16 +1001,16 @@ function ziti_createEnvFile {
...
-      export ZITI_NETWORK="$(hostname)"
+      export ZITI_NETWORK="${ZITI_HOSTNAME}"
     fi
   fi
 
   echo "ZITI_NETWORK set to: ${ZITI_NETWORK}"
 
   if [[ "${ZITI_USER-}" == "" ]]; then export ZITI_USER="admin"; fi
-  if [[ "${ZITI_PWD-}" == "" ]]; then export ZITI_PWD="admin"; fi
+  if [[ "${ZITI_PWD-}" == "" ]]; then export ZITI_PWD="$(tr -dc _A-Z-a-z-0-9 < /dev/urandom | head -c${1:-32})"; fi
```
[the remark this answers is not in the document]
qrkourier: No, but the quickstart doc they're presumably following is clear about it being stashed in the .env file and used by the `zitiLogin` command.

quickstart/docker/image/ziti-cli-functions.sh
qrkourier: I'll revert this change unless someone can test it with macOS bash v3.

quickstart/docker/image/ziti-cli-functions.sh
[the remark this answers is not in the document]
qrkourier:
```bash
  if [[ "${ZITI_PWD-}" == "" ]]; then 
    ZITI_PWD="$(tr -dc _A-Z-a-z-0-9 < /dev/urandom | head -c32)"
    echo -e "Do you want to keep the generated admin password '$ZITI_PWD'? (Y/n)"
    # shellcheck disable=SC2162
    read ZITI_PWD_REPLY
    if [[ -z "${ZITI_PWD_REPLY}" || ${ZITI_PWD_REPLY} =~ [yY] ]]; then
      echo "INFO: using ZITI_PWD=${ZITI_PWD}"
    else
      echo "Type the preferred admin password and press <enter>"
      read -r ZITI_PWD
    fi
  fi
```

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -420,7 +471,7 @@ function ziti_expressConfiguration {
 
       # Stop any devices currently running to avoid port collisions
       stopAllEdgeRouters
-      stopZitiController
+      stopZitiController "${ZITI_EXPRESS_CONTROLLER_PID}"
```
qrkourier, line 474: Note to self: global, not param

gberl002 (approved):

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -62,30 +64,46 @@ function initializeController {
...
+  if [[ -n ${ZITI_EXPRESS_EDGE_ROUTER_PID:-} ]]; then
+    # shellcheck disable=SC2015
+    kill "${ZITI_EXPRESS_EDGE_ROUTER_PID}" && {
+      echo "INFO: stopped router"
+    } || {
+      echo "ERROR: something went wrong with stopping the router(s)" >&2
+      return 1
+    }
+  else
+    echo "ERROR: you can only stop a router process that was started with startExpressEdgeRouter" >&2
```
dovholuknf, line 104: probably should rename this function now to `stopExpressEdgeRouter`. also, that name is inconsistent and it's bugged me for a while. Maybe also change `startExpressEdgeRouter` to `startEdgeRouter` ? or ... i dunno something to clean/true these ideas up?

quickstart/docker/image/ziti-cli-functions.sh
dovholuknf: did you want to keep the change or revert it? my point is that IF you keep it. you should add your comment INTO the script:

> Expands to empty string if not defined, otherwise expands as ${GITHUB_TOKEN}@

dovholuknf: what i mean is, if you ever feel the need to comment on code like this - i think that's a pretty clear indicator to take a moment and consider if that comment belongs "in" the code. in this case - i feel like it definitely would be useful to have in the script

quickstart/test/quickstart_test.go
```diff
@@ -300,12 +300,15 @@ func TestQuickStartEnvFile(t *testing.T) {
...
-This is a manually run test that will, with the default values, confirm the docker-compose ziti network is running as
-expected. The values can be edited to confirm other ziti networks but will require an http server on the back end.
+This is a manually run test that will, with the default values except the admin password, confirm the docker-compose
+ziti network is running as expected. The values can be edited to confirm other ziti networks but will require an http
+server on the back end.
 */
 func TestSimpleWebService(t *testing.T) {
 
 	// Wait for the controller to become available
+	zitiAdminUsername := "admin"
```
dovholuknf, line 310: @gberl002 can you just peek at this and make sure you're ok with this approach too, please

dovholuknf:
I assume there's a few bits of doc here/there to change, right? also i feel like this actually is important enough to add to the overall ziti CHANGELOG.md

qrkourier:
Affirmative. https://github.com/openziti/ziti-doc/pull/248/files has complementary hosted quickstart and ZAC quickstart doc changes. That's the PR where the hosted quickstart file's line endings were rewritten, so the diff is messed up. You can still see it with `git diff`:

```bash
# repo: openziti/ziti-doc
git checkout hosted-quickstart-tweaks
git diff --ignore-cr-at-eol main -- docusaurus/docs/quickstarts/network/hosted.md
```

quickstart/docker/image/ziti-cli-functions.sh
qrkourier: Good idea. I trued up the start/stop function names.

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -2,8 +2,10 @@
 
 set -uo pipefail
 
+ZITI_HOSTNAME="$(hostname -s)"
```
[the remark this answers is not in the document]
qrkourier: Now I see what you mean. I replaced `ZITI_HOSTNAME` with `DEFAULT_ZITI_NETWORK` and checked over a few scenarios. Assumption include that `expressInstall` is run first.

dovholuknf (approved):

PR #913, opened 2022-12-02
Keep certain files in the container 

### LINKED ISSUE #912: Binaries not updated in docker-compose env with new image
Because docker-compose creates a shared volume which contains the binaries, when a new image is pulled it does not update the binaries as docker leaves mapped files alone.

We need a way to update the binaries when a new image is released and `docker-compose pull` is used without tearing down the volume.

gberl002:
Closes #912

### DIFF
--- a/quickstart/README.md
+++ b/quickstart/README.md
@@ -41,7 +41,7 @@ the scripts in these folders before running them.
 ## Express
 
 By far the easiest way to establish an environment quickly is to simply run the express install script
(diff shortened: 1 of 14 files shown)


dovholuknf (approved):
lgtm - what kind of doc do we need for migration from 'the old compose' to this version? also will there be updates to the doc too?

gberl002:
Doc updates PR ready https://github.com/openziti/ziti-doc/pull/259

PR #914, opened 2022-12-02
Update HA docs and developer setup instructions. Fix chain generation

plorenz:


ziti/pki/store/store.go
```diff
@@ -37,14 +37,15 @@ type Store interface {
 	// Returns an error if it failed to store the bundle.
 	Add(string, string, bool, []byte, []byte) error
 
-	// Chain concats an intermediate cert and a newly signed certificate bundle and adds the chained cert to the store.
+	// Chain concats an signing cert and a newly signed certificate bundle and adds the chained cert to the store.
```
dovholuknf, line 40: nit: an -> a

ziti/pki/store/store.go
```diff
@@ -37,14 +37,15 @@ type Store interface {
...
 	//
 	// Args:
-	//  The intermediate CA name.
+	//  The signing CA name.
+	//  The destination CA name.
```
dovholuknf, line 44: what's a 'destination CA'?

ziti/pki/store/local.go
```diff
@@ -217,18 +217,23 @@ func (l *Local) writeBundle(caName, name string, isCa bool, key, cert []byte) er
...
 		}
 	}
 
-	caPath := filepath.Join(l.Root, caName, LocalCertsDir, caName+".cert")
-	caIn, err := os.Open(caPath)
+	caChainPath := filepath.Join(l.Root, caName, LocalCertsDir, caName+".chain.pem")
+	caIn, err := os.Open(caChainPath)
 	if err != nil {
-		return fmt.Errorf("failed to open CA: %v: %v", caPath, err)
+		// no chain path, likely means we're using the root as the CA, so fallback to the cert
```
dovholuknf, line 231: this feels like a fail/panic situation to me? i kinda feel like this should be 'right' or not and fail without making inferences?

ziti/cmd/pki_create_intermediate.go
```diff
@@ -123,6 +123,11 @@ func (o *PKICreateIntermediateOptions) Run() error {
 		return errors.Wrap(err, "cannot sign")
 	}
 
+	// Concat the newly-created intermediate cert with the signing cert to create an intermediate.chain.pem file
+	if err := o.Flags.PKI.Chain(signer, req); err != nil {
+		return errors.Wrap(err, "cannot sign")
```
dovholuknf, line 128: probably a different error here? "cannot chain"?

ziti/pki/store/store.go
plorenz: Fixed

ziti/pki/store/store.go
plorenz: Under which CA directory the chain should be stored. For intermediates we should store the chain with the intermediate CA, but server certs are stored in the signing CA, so we should store it in the signing CA dir.

ziti/pki/store/local.go
plorenz: The cases are relatively well defined. If the root is the signer, there will be no chain, if the signer is an intermediate there will be. Either way, if there's no chain, we just use the signer as the chain, which was the previous behavior.

ziti/cmd/pki_create_intermediate.go
plorenz: Fixed

doc/ha/overview.md
```diff
@@ -1,42 +1,229 @@
-# HA Setup for Development
+# OpenZiti HA
 
-**NOTE: HA is a work in progress and not yet usable for anything other than developing HA**
+This document gives a brief overview of how OpenZiti HA works and how it differs from running
+OpenZiti in non-HA mode.
```
plorenz, line 5: Maybe add a note about certs, why to use different intermediates, spiffe ids, what's expected in the cert chain, etc

plorenz: done

camotts (approved):
Looks good!

PR #918, opened 2022-12-07
Setting ZITI_BIN_ROOT explicitly for docker environments

### LINKED ISSUE #917: ZITI_BIN_ROOT is incorrect in docker env
When using the docker environment, the `generateEnv` function is called, this function incorrectly sets the ZITI_BIN_ROOT for docker environments.

This doesn't pose an issue per se since ZITI_BIN_ROOT is not used in the normal startup of docker but it could pose an issue down the road or if certain functions are called manually after the docker environment is spun up.

gberl002:
Removing overwrite of ZITI_BIN_ROOT
closes #917

### DIFF
--- a/quickstart/docker/image/ziti-cli-functions.sh
+++ b/quickstart/docker/image/ziti-cli-functions.sh
@@ -1116,8 +1116,6 @@ function ziti_createEnvFile {
   export ZITI_SIGNING_ROOTCA_NAME="${ZITI_SIGNING_CERT_NAME}-root-ca"
   export ZITI_SIGNING_INTERMEDIATE_NAME="${ZITI_SIGNING_CERT_NAME}-intermediate"
 
-  export ZITI_BIN_ROOT="${ZITI_HOME}/ziti-bin"
-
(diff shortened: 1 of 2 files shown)


quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -1116,8 +1116,6 @@ function ziti_createEnvFile {
   export ZITI_SIGNING_ROOTCA_NAME="${ZITI_SIGNING_CERT_NAME}-root-ca"
   export ZITI_SIGNING_INTERMEDIATE_NAME="${ZITI_SIGNING_CERT_NAME}-intermediate"
 
-  export ZITI_BIN_ROOT="${ZITI_HOME}/ziti-bin"
```
gberl002, line 1119: In `generateEnvFile` there is already a call to check if ZITI_BIN_ROOT is set, if not then it sets it, otherwise it leaves it alone. This just overwrites it every time.

qrkourier (approved):

PR #940, opened 2023-01-05
added commands for new raft interactions

camotts:


plorenz (approved):
Looks good overall. I'll probably go in and rename join to add and leave to remove, since join and leave both imply local action, when it's more of a global action. I can do that in a follow up pass, though.

PR #949, opened 2023-01-13
Use docker buildx to build/push Quickstart image

scareything:
The only way I've found to successfully push Multiplatform images is with `docker buildx build --push`. The major downside of using buildx is that it's cumbersome for local development.

To the best of my knowledge in order to use an image that was built with buildx locally, you must push and then pull the image. I think this is a worthwhile tradeoff for mutliplatform representation on the Quickstart image.

### DIFF
--- a/quickstart/docker/image/Dockerfile
+++ b/quickstart/docker/image/Dockerfile
@@ -1,3 +1,15 @@
+FROM ubuntu:rolling as fetch-ziti-bins
+
+# optional arg to specify which version to fetch when local "ziti-bin" directory is not present
+ARG ZITI_VERSION_OVERRIDE
+
+COPY . /docker.build.context
+RUN apt update && apt install jq curl -y && \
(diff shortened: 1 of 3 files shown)


qrkourier:
I arrived at the same conclusion about where to send the output from BuildKit. Possibly of  value for reference:

Shell scripts for building multi-platform tunneler containers for testing:
* https://github.com/openziti/ziti/blob/main/ziti-tunnel/docker/buildx.sh
* https://github.com/openziti/ziti-tunnel-sdk-c/blob/main/docker/buildx.sh

GH Actions workflows that run BuildKit for tunneler container releases:
* https://github.com/openziti/ziti-tunnel-sdk-c/blob/main/.github/workflows/publish-containers.yml
* https://github.com/openziti/ziti/blob/main/.github/workflows/main.yml#L258

qrkourier (approved):

PR #956, opened 2023-01-18
updates to compose exposing more ports/variables

dovholuknf:
trying to support using docker compose in simplified or not as "host it anywhere"


scareything (approved):
I just ran into this during my own experience with the docker quickstart, and I simply updated the .env file. I didn't know `docker-compose` supported the default value syntax that you used here. nice!

PR #961, opened 2023-01-19
Added function for creating additional routers

gberl002:

### DIFF
--- a/quickstart/docker/image/ziti-cli-functions.sh
+++ b/quickstart/docker/image/ziti-cli-functions.sh
@@ -45,7 +45,7 @@ function BLUE {
 }
 
 function zitiLogin {
-  "${ZITI_BIN_DIR-}/ziti" edge login "${ZITI_EDGE_CTRL_ADVERTISED}" -u "${ZITI_USER-}" -p "${ZITI_PWD}" -c "${ZITI_PKI_OS_SPECIFIC}/${ZITI_EDGE_CONTROLLER_ROOTCA_NAME}/certs/${ZITI_EDGE_CONTROLLER_INTERMEDIATE_NAME}.cert"
+  "${ZITI_BIN_DIR-}/ziti" edge login "${ZITI_EDGE_CTRL_ADVERTISED_HOST_PORT}" -u "${ZITI_USER-}" -p "${ZITI_PWD}" -y 2>&1
 }
 function cleanZitiController {
   checkEnvVariable ZITI_HOME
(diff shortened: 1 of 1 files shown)


quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -1514,27 +1519,32 @@ function setOs {
 }
 
 function checkEnvVariable() {
-
+  _error=false
```
scareything, line 1522: if we are assuming bash (or zsh), it would be a good idea to declare this as a `local`.

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -1552,4 +1562,32 @@ function getFileOverwritePermission() {
   fi
 }
 
+function addRouter {
+  router_name="${1-}"
```
scareything, line 1566: another candidate for `local`

qrkourier:
On the subject of variable scope, I noted that we're using `export` liberally, or perhaps even as a rule. This shouldn't be necessary and may have unintended effects if the child processes are not the intended audience of the variables. Rather, we should only `export` when required.

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -1552,4 +1562,49 @@ function getFileOverwritePermission() {
...
+  fi
+
+  # Check expected and set up environment variables
+  checkEnvVariable ZITI_USER ZITI_PWD ZITI_CTRL_ADVERTISED_ADDRESS ZITI_EDGE_CONTROLLER_PORT
+  retVal=$?
+  if [[ "${retVal}" != 0 ]]; then
+    return 1
+  fi
+  export ZITI_EDGE_CTRL_ADVERTISED_HOST_PORT="${ZITI_CTRL_ADVERTISED_ADDRESS}":"${ZITI_EDGE_CONTROLLER_PORT}"
+  if [[ "${router_type-}" == "" ]]; then export router_type="public"; fi
```
gberl002, line 1585: Remove this duplicated code that will never execute

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -1552,4 +1562,49 @@ function getFileOverwritePermission() {
...
+  # Create router config
+  if [[ "private" == "${router_type}" ]]; then
+    createPrivateRouterConfig "${router_name}"
+  else
+    createEdgeRouterConfig "${router_name}"
+  fi
+
+  # Enroll the router
+  "${ZITI_BIN_DIR-}/ziti" router enroll "${ZITI_HOME}/${router_name}.yaml" --jwt "${ZITI_HOME}/${router_name}.jwt" &> "${ZITI_HOME}/${router_name}.enrollment.log"
+    retVal=$?
```
gberl002, line 1601: Why is the following code indented?

gberl002:
This feature was added during quickstart cleanup

PR #971, opened 2023-01-24
add container images for ctrl, router, tunnel

qrkourier:


scareything (changes requested):
Now that entry point scripts are present I definitely think it makes sense to give each image it's own subdirectory.

scareything (approved):

docker-images/controller/Dockerfile
```diff
@@ -0,0 +1,20 @@
+ARG ZITI_VERSION="latest"
+# this builds docker.io/openziti/ziti-controller
+FROM docker.io/openziti/ziti-cli:${ZITI_VERSION}
+
+# This build stage grabs artifacts that are copied into the final image.
+# It uses the same base as the final image to maximize docker cache hits.
+
+### Required OpenShift Labels 
+LABEL name="openziti/ziti-controller" \
```
scareything, line 9: should the directory name match the image label? just a thought.

qrkourier: Ah, yes. I see your point. Use the same basename from the label instead of abbreviating the folder names. I'll make this change in the target branch where all the image-related changes are aggregating. That'll save me from some merge conflicts.

PR #988, opened 2023-02-03
organize and update cross-build container

qrkourier:


docker-images/cross-build/Dockerfile
```diff
@@ -4,8 +4,9 @@ FROM debian:bullseye-slim
 # reproducible builds for downstream forks for Ziti contributors 
 #
 
-ARG latest_golang=1.17.2
-ARG go_distribution_file=go${latest_golang}.linux-amd64.tar.gz
+ARG TARGETARCH
+ARG golang_version=1.19.5
```
dovholuknf, line 8: is there any way to 'just take the latest'? this is gonna be a constant battle to keep up?

docker-images/cross-build/README.md
```diff
@@ -16,18 +17,23 @@ LATEST_GOLANG=$(curl -sSfL "https://go.dev/VERSION?m=text" | /bin/grep -Po '^go(
...
-    --file=Dockerfile.linux-build \
-    --build-arg latest_golang=${LATEST_GOLANG} \
     --build-arg uid=$UID \
-    --build-arg gid=$GID .
+    --build-arg gid=$GID \
+    ./docker-images/cross-build/
 ```
 
+<!--
+    I pruned this arg because it doesn't currently match the go.dev URL, and
```
dovholuknf, line 26: this comment is strange to see here? what's this about?

docker-images/cross-build/Dockerfile
qrkourier: This line sets a default value in case the build arg is not supplied. The correct Go version is the one declared in go.mod. I've updated the README for this Dockerfile to automate providing the correct Go version as this build arg and verified that it works with the currently-declared Go version, 1.19.

docker-images/cross-build/README.md
qrkourier: I found a way to solve the issue I was commenting about, so I removed the comment.

qrkourier:
@dovholuknf ready for review

dovholuknf (approved):
lgtm

PR #990, opened 2023-02-03
update local dev tutorial

qrkourier:


doc/002-local-dev.md
```diff
@@ -7,56 +7,40 @@ This local development README guides you to install the Ziti stack built from th
 You will need to [install a version of Go](https://go.dev/) that is as recent as the version used by this project. Find the current minimum version by running this command to inspect `go.mod`.
 
 ```bash
-grep -Po '^go\s+\K\d+\.\d+$' go.mod
+grep -Po '^go\s+\K\d+\.\d+(\.\d+)?$' go.mod
```
qrkourier, line 10: This improves the Perl regexp that matches the semver in our go.mod file so that it's more resilient to change. Without this improvement, the regexp will fail to match if we start expressing the Go version as a patch-level semver. This allows the patch level to be optional. Go.dev's latest release is 1.20, for example, without the patch level of the semver. They appear to alternate between minor level and patch level in their download URL patterns and associated download directory structure, hence the need for flexibility at every level when refering to the Go release version.

doc/002-local-dev.md
```diff
@@ -7,56 +7,40 @@ This local development README guides you to install the Ziti stack built from th
...
 ## Build and Install All Applications
 
-This repo contains several Go applications. The easiest way to build and install all to `${GOPATH}/bin` is:
+This repo contains several Go applications, e.g. ziti-controller, ziti-router. These applications were reorganized as subcommands of the `ziti` CLI application, so it's no longer necessary to build the discrete binaries. Only the `ziti` CLI is necessary. The easiest way to build and install `ziti` in `${GOPATH}/bin` is:
 
 ```bash
-# build and install all apps
-go install ./...
+# build and install ziti CLI
+go install ./ziti
```
qrkourier, line 19: Now `ziti` is the only package that most devs want to build because it has subsumed the others, e.g. `ziti-controller`. This example will install `${GOPATH}/bin/ziti` instead of littering GOPATH with the vestigial artifacts that are no longer needed by most devs.

dovholuknf (approved):

Issue #832, opened 2022-09-13
Perhaps "Zitify" is not the best marketing verb


jfxio:
Hi. I listened to Clint Dovholuk's DeveloperWeek Cloud's talk. Ziti looks useful, we'll be giving it a whirl, but please, please on your website, change "Zitify your application". It scans as "Zit-ify your application". I don't want to turn my application into a giant puss-filled pimple ;-)

dovholuknf:
bahahahahahahahah - the funniest part is that you're not the first person to notice that. Sounds like __something__ must be done. :)

ekoby:
we use ziti-enabled or ziti-embedded more now

PR #1000, opened 2023-02-09
#832 minor issue resolved regarding verb

dsukruth:
Minor change of words in the documentation.


dovholuknf:
thanks for the PR. "zitify" is actually a term that we use to mean "add ziti to something". I added the term to the glossary at https://docs.openziti.io/docs/reference/glossary#zitification-zitified-zitify 

unfortunately that means i'm going to close your PR with "wontfix", but thank you for the PR! :)

PR #1008, opened 2023-02-13
make version checking optional

dovholuknf:
This PR makes it optional to check the version of the software running and prevents any kind of "phone home looking" type of activity (even though it's only checking for the version, not actually "phoning home")


qrkourier (approved):
This solves the immediate problems of surprising network requests and output. Ideally, the version checker function is additionaly exposed through a sub-command with meaningful exit codes and output, e.g., `ziti check-version`. That can wait for future enhancement if it ever seems worth doin'.

plorenz (approved):

gberl002 (approved):

PR #1025, opened 2023-03-01
Fix #920 Rerunning quickstart

### LINKED ISSUE #920: Rerunning quickstart will always result in error
When reinstalling quickstart you are prompted to reset env variables. If you go ahead with the install the environment variables are reset prior to trying to stop the router and controller which will always fail since it now relies on the PID which is stored in an env variable.

The variables should be cleared after the controller and router are shut down. Or, the PID of the processes could be stored in an environment variable that does not start with `ZITI_`

snippet of the issue
```
      # Silently clear ziti variables
      unsetZitiEnv "-s"

      if [[ "${specifiedVersion}" != "" ]]; then
        export ZITI_VERSION_OVERRIDE="${specifiedVersion}"
      fi

      # Stop any devices currently running to avoid port collisions
      stopRouter
      stopController
```

dsukruth:


gberl002 (approved):

gberl002:
I've approved this PR, but I can't merge it because we require that commits are signed. Would you mind setting yourself up with signed commits? Sorry for the inconvenience, but definitely thank you for your contribution.

https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

You will have to either amend this commit or remove it and recommit with the signed commit. I'll leave this PR open in the meantime.

dsukruth:
Okay understood @gberl002

gberl002:
@dsukruth sorry to bug you about this again, I see you got signed commits working. The branch won't let me merge unless all the commits are signed and only your most recent one is. 

The easiest solution is probably just delete this branch and create a new one with only signed commits.

gberl002:
Cannot merge due to unsigned commits

Issue #1031, opened 2023-03-10
Quickstart on Raspberry Pi 4 running 32 bit does not run


gberl002:
When the binary zip download is attempted, there is an erroneous newline character in the URL for the zip, causing a failure.

Example of the newline error

```
tar: /home/gberl001/.ziti/quickstart/raspberrypi/ziti-bin/ziti-linux-arm-0.27.5.tar.gz\nziti-linux-arm64-0.27.5.tar.gz: Cannot open: No such file or directory.
```

I believe this is because a search for arm also returns arm64 so we need to clarify starts with but not followed by 64.

PR #1033, opened 2023-03-10
Issue 1031 arm64 on arm

gberl002:
When downloading the zip, looking for begins with "arm" captures both "arm" and "arm64"


gberl002:
closes #1031

dovholuknf (approved):

PR #1066, opened 2023-03-30
publish container images on release-next

qrkourier:
This changes the build to enable open-source contributors and team members to consume the bleeding-edge builds from release-next as a container image in Hub, i.e. `:release-next`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - release-v*
+      - release-next
   pull_request:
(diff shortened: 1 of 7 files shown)


plorenz (approved):

qrkourier:
Thanks @dovholuknf for pointing out this won't work as written because I erroneously assumed the Docker Hub login name is the same value as the Hub organization. This is true for my personal Hub account where I tested the branch, but not for OpenZiti which is shared with multiple login names. I'll have to rework this branch slightly.

PR #1079, opened 2023-04-08
Add ziti edge login support for certs/ext-jwt-signer (via file)

dovholuknf:
Initial update to allow certificate-baed login for `ziti`.
* changed the 'cert' flag to'ca', it's more accurate anyway imo. also removed the 'c' short flag to be repurposed to the cert-based auth
* made struct loginOptions, prettyPrintResponse(), outputJson() public
* moved EdgeControllerLogin to just 'login' and made it private since it wasn't used elsewhere


CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 ## What's New
 
 * CLI additions for `ziti` to login with certificates or external-jwt-signers
+* NOTE: ziti edge login flag changes:
+  * `-c` flag has been changed to map to `--client-cert`
```
qrkourier, line 7: It's difficult to estimate how much user pain we can expect from this change. I'm sure there will be some, but the new options are more clear. `--cert` was pretty vague, so that's an improvement.

I suggest that we have a uniform long option name, e.g. `--ca-cert` for all this CLI's sub-commands, or make it a global option. For example, `--ca-cert` is  already used by this CLI. The uniformity seems more important than reverse compatibility.

plorenz (commented):
Can we bump to 0.28, please, since this is a bit of a breaking change?

plorenz (approved):

PR #1088, opened 2023-04-15
tweak build workflow for forks 

### LINKED ISSUE #1087: re-enable CI in forks
The Linux build never runs because the Fablab job is skipped in forks. I think we want to skip Fablab in forks because it wouldn't succeed without cloud credentials anyway, but there's not currently any other way to get a Linux build which is a prerequisite for container builds which are prerequisite to deploying and testing the artifacts in a container-based environment.

Based on what I've read so far, it might be possible to embellish the job conditions to run the Linux build job when Fablab is skipped.

Relevant discussion: https://github.com/orgs/community/discussions/45058

qrkourier:
resolves #1087

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,7 +29,8 @@ env:
 jobs:
   mac-os-build:
     name: Build Mac OS binaries
-    if: github.repository_owner == 'openziti'
+    # allow fors to opt-out of time-consuming macOS builds
+    if: vars.ZITI_SKIP_MACOS_BUILD != 'true'
     runs-on: macos-11
     steps:
       - name: Git Checkout
@@ -60,7 +61,8 @@ jobs:
 
   windows-build:
     name: Build Windows binaries
-    if: github.repository_owner == 'openziti'
+    # allow fors to opt-out of time-consuming Windows builds
+    if: vars.ZITI_SKIP_WINDOWS_BUILD != 'true'
     runs-on: windows-2019
     steps:
       - name: Git Checkout
@@ -92,6 +94,7 @@ jobs:
 
   fablab-smoketest:
     name: Fablab Smoketest
+    # not applicable to forks
     if: github.repository_owner == 'openziti'
     runs-on: ubuntu-20.04
     steps:
@@ -157,6 +160,8 @@ jobs:
 
   publish:
     name: Publish Binaries
+    # - always() allows evaluating further conditional expressions even if
+    #   needed jobs were skipped
     if: |
       always()
       && (needs.mac-os-build.result == 'success' || needs.mac-os-build.result == 'skipped')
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -202,19 +209,21 @@ jobs:
...
       - name: Download windows release artifact
+        if: needs.windows-build.result == 'success'
         uses: actions/download-artifact@v3
         with:
           name: windows-release-${{ github.run_id }}
           path: release/
 
       - name: Install Jfrog CLI
-        uses: jfrog/setup-jfrog-cli@v2
+        uses: jfrog/setup-jfrog-cli@v3
```
dovholuknf, line 226: this one change seems minor but it worries me. sometimes things like this have unintended consequences.

.github/workflows/main.yml
```diff
@@ -155,6 +157,11 @@ jobs:
 
   publish:
     name: Publish Binaries
+    if: |
+      always()
```
dovholuknf, line 161: i'd like to just understand what and why you're doing this. maybe we can chat?

.github/workflows/main.yml
qrkourier: This is a response to GH's deprecation notice for Node.js v12. We should try v3 before they fully decommission. We used this change to the jfrog CLI setup action in the checks run on this PR, so we know that it works.

.github/workflows/main.yml
qrkourier: https://github.com/openziti/ziti/issues/1087

qrkourier: @dovholuknf  There's a bit more context and a detailed explanation of this workaround in this issue: https://github.com/openziti/ziti/issues/1087

scareything: maybe a comment would be in order here, explaining the workaround? I'd expect most folks who look at this critically would at least be confused by the `always()`, and might even remove it.

qrkourier: Comments added to reduce confusion for future maintainers, and I gave the reasons for skipping certain builds some more thought. 

#### TL;DR

I've refactored this CI workflow to always build Windows and macOS, even on forks, and allow forks to opt-out of those builds by setting a GH Actions repo var.

#### Explanation

I agree it makes sense to continue building Windows and macOS by default, as long as they can succeed, despite being time-consuming. 

I didn't specifically mention that I wanted to skip some builds because they were time-consuming when we talked because I was attempting to generalize a solution for ziti and zrok. The case I did mention where a signing cert is required is applicable to zrok's branch builds and ziti release builds, not ziti branch builds.

I've changed the conditional to allow forks to opt out of time-consuming builds by setting an optional var in their repo's GH Actions "vars." Ziti and zrok (different PR for zrok in the works) builds that require a signing cert will only run if repo owner is "openziti."

```
    if: vars.ZITI_SKIP_MACOS_BUILD != 'true'
```

With these vars set, the fork's CI run looks like this example in my fork where I'm now able to iterate quickly and obtain a container image with the Linux build: https://github.com/qrkourier/ziti/actions/runs/4766536885

scareything (approved):

PR #1027, opened 2023-03-01
Adding default REST API service to quickstart

gberl002:
Added express install tests
Added tests for REST API service
Added ability to provide a .env file to tests
Added go.mod to test to keep dependencies out of ziti builds

Issue #1013, opened 2023-02-13
docker env password is renewed at each `docker-compose up`


gberl002:
If a docker-compose environment is spun up, then stopped (leaving the volume). If that compose environment is started again, a new password is generated and stored in the `ziti.env` file. This causes subsequent logins to fail due to the incorrect password being used.

The environment should use the existing password if one exists during start up. The way the current quickstart script is set up, it will not prompt for a new password if one is already set so a quick solution to this might simply be sourcing `ziti.env` before running `generateEnvFile` in `run-controller.sh`. A more robust solution probably wouldn't be much more difficult though.

PR #1092, opened 2023-04-18
Adding ability to set ZITI_PWD in docker

gberl002:
Fixing password re-generation on container down and up cycles Closes #1027


quickstart/docker/.env
```diff
@@ -12,6 +12,9 @@ ZITI_CONTROLLER_RAWNAME=ziti-controller
 #ZITI_CONTROLLER_HOSTNAME=advertised.address
 #ZITI_CTRL_PORT=8440
 
+# Leave password blank to have a unique value generated or set the password explicitly
+ZITI_PWD=admin
```
dovholuknf, line 16: i'd prefer to keep it generated

quickstart/docker/image/run-controller.sh
```diff
@@ -7,8 +7,11 @@ export ZITI_EDGE_CONTROLLER_RAWNAME=ziti-edge-controller
 
 . "${ZITI_SCRIPTS}/ziti-cli-functions.sh"
 
-generateEnvFile
-. ${ZITI_HOME}/ziti.env
+# access-control.init is created by ziti-controller-init-container after initial controller startup
+if [ ! -f "${ZITI_HOME}/access-control.init" ]; then
```
dovholuknf, line 11: does this work with "just docker"? aka not compose too?

quickstart/docker/.env
gberl002: Oops, that was intended to be blank.

quickstart/docker/simplified-docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
     environment:
       - ZITI_EDGE_IDENTITY_ENROLLMENT_DURATION=${ZITI_EDGE_IDENTITY_ENROLLMENT_DURATION}
       - ZITI_EDGE_ROUTER_ENROLLMENT_DURATION=${ZITI_EDGE_ROUTER_ENROLLMENT_DURATION}
+      - ZITI_PWD=${ZITI_PWD}
```
dovholuknf, line 13: don't forget the other compose file

dovholuknf (approved):

quickstart/docker/image/run-controller.sh
gberl002: It does, assuming they follow the directions. If they don't run the init-container (step 2 in "With Docker" setup) then it would not.

I toyed with the idea of seeing if the .env file was empty or not to determine the password generation. Would you prefer that method? Or, we can have a file generated by the controller, just like `access-control.init` and just keep the same style as above with the controller-based file instead.

gberl002:
closes #1013

PR #1095, opened 2023-04-21
get Docker login name from var or secret

qrkourier:


.github/workflows/publish-docker-images.yml
```diff
@@ -39,7 +39,9 @@ jobs:
       - name: Login to Docker Hub
         uses: docker/login-action@v2
         with:
-          username: ${{ secrets.DOCKER_HUB_API_USER }}
+          # it is preferable to obtain the username from a var so that
+          # recurrences of the same string are not masked in CI output
+          username: ${{ vars.DOCKER_HUB_API_USER || secrets.DOCKER_HUB_API_USER }}
```
dovholuknf, line 44: i don't see the point of this change. what did you need this output to be in the clear for?

qrkourier: The login name string is masked when it happens to be the same as some other string in the CI output. Like GitHub, the Hub login name is typically the same as the Hub organization. This means that unnecessarily masking the login name causes the Hub organization too to be masked, which makes the logs less useful.

qrkourier: This change gives the repo owner a choice to have more useful logs by not masking the login name while changing nothing for existing repo owners who prefer to store their Hub login as a GH Actions secret.

qrkourier: @dovholuknf At a bare minimum, this change improves my experience running Ziti CI in a fork.

dovholuknf: I understand it helps you somehow. I still don't understand why or how or why it's necessary. it's just one more subtlety that can/will cause confusion at some point... but ok ¯\_(ツ)_/¯

dovholuknf (approved):

PR #1109, opened 2023-05-11
add user ziggy in container image

qrkourier:


docker-images/ziti-cli/Dockerfile
```diff
@@ -43,12 +43,17 @@ COPY --from=bitnami-kubectl /opt/bitnami/kubectl/bin/kubectl /usr/local/bin/
 RUN mkdir -p -m0755 /licenses
 COPY ./LICENSE /licenses/apache.txt
 
+RUN groupadd --gid 1001 ziggy \
```
scareything, line 46: I'm not keen on pinning the uid/gid to a value that's likely to already exist on the host.

I understand that the goal of this change is to facilitate automatically configuring ziti autocompletion, which requires a home directory. Could we continue using the user `nobody`, and populate the completion files in `nobody`'s home directory?

dovholuknf (approved):

Issue #1123, opened 2023-05-24
cannot update config types with ziti cli


scareything:
I encountered the following problems when attempting to update a config-type with the cli:

1. The CLI is attempting to update "configs" instead of "config-types".
    ```
        $ ziti edge show config-type ziti-tunneler-client.vX
        {
            "$id": "http://edge.openziti.org/schemas/ziti-tunneler-client.v1.config.json",
            "additionalProperties": false,
            "properties": {
                "hostname": {
                    "type": "string"
                },
                "port": {
                    "maximum": 65535,
                    "minimum": 0,
                    "type": "integer"
                }
            },
            "required": [
                "hostname",
                "port"
            ],
            "type": "object"
        }

        $ ziti edge update config-type ziti-tunneler-client.vX -f ~/tmp/client.v1.json
        error: no configs found with id or name ziti-tunneler-client.vX
    ```
2. After fixing the above issue, the update failed because the schema was rejected:
    ```
        $ cat ~/tmp/client.v1.json
        {
        "$id": "http://edge.openziti.org/schemas/ziti-tunneler-client.v1.config.json",
        "additionalProperties": false,
        "properties": {
            "hostname": {
                "type": "string"
            },
            "port": {
                "maximum": 65535,
                "minimum": 2,
                "type": "integer"
            }
        },
        "required": [
            "hostname",
            "port"
        ],
        "type": "object"
    }
    $ ziti edge update config-type ziti-tunneler-client.v1 -f ~/tmp/client.v1.json
    $ error: error updating config-types/f2dd2df0-9c04-4b84-a91e-71437ac229f1 instance in Ziti Edge Controller at https://localhost:1280/edge/management/v1. Status code: 400 Bad Request, Server returned: {
    "error": {
        "cause": {
            "field": "schema",
            "reason": "schema in body must be of type object",
            "value": "\u003cnil\u003e"
        },
        "code": "COULD_NOT_VALIDATE",
        "message": "The supplied request contains an invalid document or no valid accept content were available, see cause",
        "requestId": "gX0v8ayZ6q"
    },
    "meta": {
        "apiEnrollmentVersion": "0.0.1",
        "apiVersion": "0.0.1"
    }
    }
    ```
    This was fixed by putting the schema in the `schema` path of the api request instead of the `data` path.

PR #1124, opened 2023-05-24
use correct entity type when updating config-types

scareything:
fixes https://github.com/openziti/ziti/issues/1123


plorenz (approved):
For future reference, the correct way to reference the issues is with `Fixes #<issue number>` (or similar). See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

Issue #1132, opened 2023-05-31
Update router yml to use wss instead of ws


gberl002:
The protocol format of `ws` is no longer supported. Router configs need to be updated to use `wss` instead.

In `router.yml` there is a listener line with `ws` hardcoded, it needs to be updated to `wss`.

```
{{ if .Router.IsFabric }}#{{ end }}    address: {{ if .Router.IsWss }}ws{{ else }}tls{{end}}:0.0.0.0:{{ .Router.Edge.Port }}
```

Due to this, when starting a WSS router the following error will appear.
```
FATAL fabric/router.(*Router).startXgressListeners: error listening [edge] (transport.ws not supported. use transport.wss)
```

PR #1133, opened 2023-05-31
Update router yml to use wss instead of ws

tungbq:
Fixes: #1132


gberl002:
Testing this now; thanks for working on this.

gberl002 (approved):

PR #1175, opened 2023-06-23
Adding missing public attribute

gberl002:


dovholuknf (approved):
well i can see how that happened. a comment might be in order... but lgtm

PR #1224, opened 2023-07-24
enable to run controller and router listeners on a single port

### LINKED ISSUE #1135: Edge Router: Support multiple protocols on the same listener port
I would like to be able to use the same port for multiple protocols.  For example, when I host an edge router today I need up to three ports:
80 - fabric link listener
443 - edge TLS listener
7001 - edge WSS listener

I would dearly love to be able to use only one port for all three types of traffic.  Two ports - 80 for link and 443 for all edge would be good enough, too.

ekoby:
[fixes #1135]
technical notes:
uses TLS ALPN feature to select proper handler for incoming connection

the following application protocols are defined:
- ziti-ctrl - control channel
- ziti-link - link channel
- ziti-edge - edge connection (also accepts ''(blank) for backward compatibility)
- Controller REST API endpoint uses standard protocols: 'h2', 'http/1.1', and ''(blank) for SDK clients
- Edge Router WSS endpoint uses standard protocols: 'h2', 'http/1.1'


andrewpmartinez (commented):
lgtm assuming we don't want to handle router backward compatibility differently.

andrewpmartinez (approved):

PR #1230, opened 2023-08-01
Fix docker race condition

gberl002:


quickstart/docker/image/run-router.sh
```diff
@@ -14,12 +14,6 @@ fi
 
 . ${ZITI_HOME}/ziti.env
 
-# wait for the controller to come online
-_wait_for_controller
```
dovholuknf, line 18: why are you removing this? this is important...

qrkourier (approved):

Issue #1219, opened 2023-07-19
IP_OVERRRIDE values are not making it into the SANs list


gberl002:
There are various environment variables with the suffix `IP_OVERRIDE` which should be making it into the SANs list, however, they are not.

gberl002:
Closed by #1235

PR #1235, opened 2023-08-04
Issue 1219 ZITI_EDGE_CTRL_IP_OVERRIDE not in SANs

gberl002:
Also added a check for a DNS appearing as an IP, will warn user and not add that IP as a DNS in the SANs


qrkourier (approved):

gberl002:
closes #1219

PR #1244, opened 2023-08-11
Build and publish linux package ziti-cli

qrkourier:
Testing repo for RedHat family distros

```
sudo tee -a /etc/yum.repos.d/openziti.repo >/dev/null <<EOF
[OpenZiti Test]
name=OpenZitiTest
baseurl=https://packages.openziti.org/zitipax-openziti-rpm-stable/testing/$basearch
enabled=1
gpgcheck=0
gpgkey=https://packages.openziti.org/zitipax-openziti-rpm-stable/testing/$basearch/repodata/repomd.xml.key
repo_gpgcheck=1
EOF

sudo dnf update

sudo dnf install ziti-cli
```

Testing repo for Debian family distros

```
curl -sSLf https://get.openziti.io/tun/package-repos.gpg \
  | sudo gpg --dearmor --output /usr/share/keyrings/openziti.gpg

sudo tee -a /etc/apt/sources.list.d/openziti.list >/dev/null <<EOF
deb [signed-by=/usr/share/keyrings/openziti.gpg] https://packages.openziti.org/zitipax-openziti-deb-stable testing main
EOF

sudo apt update

sudo apt install ziti-cli

```


.github/nfpm-configs/nfpm-linux-ziti-cli.yaml
```diff
@@ -0,0 +1,20 @@
...
+maintainer: ${ZITI_MAINTAINER}
+description: >
+  The ziti-cli package provides the ziti executable binary as a command line
+  interface for Ziti.
+vendor: ${ZITI_VENDOR}
+homepage: ${ZITI_HOMEPAGE}
+license: Apache-2.0
+contents:
+- src: ./release/ziti
+  dst: /usr/bin/ziti
```
sabedevops, line 18: Should this install to `/opt/openziti/bin` instead?

qrkourier: Only if we symlink it from a dir that's typically in the executable search PATH because the CLI is user facing and interactive, not a daemon. That's more a UX perspective than a rigid adherence to FHS, but I can't imagine it'd go over well asking every user to modify their PATH.

.github/workflows/publish-linux-packages.yml
```diff
@@ -0,0 +1,118 @@
+name: Publish Linux Packages
+
+on:
+  workflow_call:
+    inputs:
+      ziti-version:
+        description: semver string generated by 'ziti-ci get-current-version'
```
qrkourier, line 7:
```suggestion
        description: generated by 'ziti-ci get-current-version'
```

.github/workflows/publish-linux-packages.yml
```diff
@@ -0,0 +1,118 @@
...
+        uses: actions/download-artifact@v3
+        with:
+          name: linux-release-${{ github.run_id }}
+          path: release/
+
+      - run: mv -v ./release/${{ matrix.arch.gox }}/linux/ziti ./release/ziti
+
+      - name: Build Package
+        id: nfpm
+        # uses: burningalchemist/action-gh-nfpm@v1
```
qrkourier, line 56:
```suggestion
        uses: burningalchemist/action-gh-nfpm@v1
```

.github/workflows/publish-linux-packages.yml
```diff
@@ -0,0 +1,118 @@
...
+        uses: netfoundry/action-gh-nfpm@main
```
qrkourier, line 57:
```suggestion
        # uses: netfoundry/action-gh-nfpm@main
```

.github/workflows/publish-linux-packages.yml
qrkourier: They accepted and released the patch that motivated me to fork the nfpm action so we can revert back to the upstream

.github/nfpm-configs/README.md
```diff
@@ -0,0 +1,6 @@
+
```
plorenz, line 1: I don't love this living under .github, but we don't currently have a build directory, or something similar. If we accumulate more things like, we should consolidate them in one top level place along with the docker stuff.

plorenz (approved):

.github/nfpm-configs/README.md
qrkourier: I agree it makes more sense to reserve `.github` for GitHub. :slightly_smiling_face: I was thinking "only GH will run this," but it's also possible someone wants to run it locally to test a package change:

```bash
ZITI_VERSION=0.30.0 \
ZITI_HOMEPAGE="https://openziti.io" \
ZITI_VENDOR=netfoundry \
ZITI_MAINTAINER="Maintainers <developers@openziti.org>" \
GOARCH=amd64 \
    nfpm pkg --packager deb --target  . --config linux-packages/nfpm-linux-ziti-cli.yaml
```

I've moved those files out of ".github" dir to a new top-level dir "linux-packages."

sabedevops: Should the top level be something like dist, or packages with a subfolder for linux?

qrkourier: "packages" might be confusing because Go has "packages" too, and "dist" doesn't have a clear meaning by itself. How about "build/dist-packages/linux?" Then I can send another pull request to nest "build/docker-images." That'll keep further install/distribution packages organized like "build/dist-packages/homebrew."

sabedevops: Sounds good to me. Nice work on all this @qrkourier

.github/workflows/main.yml
```diff
@@ -381,3 +381,19 @@ jobs:
...
+  # call on release-next and main branches to publish linux packages to
+  # "testing" and "release" package repos in Artifactory
+  call-publish-linux-packages:
+    # always() re-enables evaluating conditionals in forks even if Windows or
+    # macOS builds were skipped
+    if: |
+      always()
+      && needs.publish.result == 'success'
+      && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/release-next')
+    name: Publish Release Linux Packages
```
qrkourier, line 394:
```suggestion
    name: Publish Linux Packages
```

.github/workflows/publish-linux-packages.yml
```diff
@@ -0,0 +1,126 @@
...
+on:
+  workflow_call:
+    inputs:
+      ziti-version:
+        description: generated by 'ziti-ci get-current-version'
+        type: string
+        required: true
+
+# cancel older, redundant runs of same workflow on same branch
+concurrency:
```
qrkourier, line 12: We experienced a deadlock with the Jenkins smoke test reusable workflow. It may be necessary to move this concurrency group to the job level or remove it and rely on the calling workflow's concurrency group.

build/dist-packages/README.md
```diff
@@ -0,0 +1,6 @@
+
+# nfpm-configs
+
+These are `nfpm` configuration files. `nfpm` generates Linux packages, e.g.,
+RPM. These files references environment variables set in GitHub Actions workflow
```
qrkourier, line 5:
```suggestion
RPM. These files reference environment variables set in GitHub Actions workflow
```

qrkourier: Explain how to build the package with standalone nfpm

qrkourier (commented):
I plan to revisit this Monday 21st.

PR #1260, opened 2023-08-23
further reduce unnecessary trust from quickstart

dovholuknf:


qrkourier (approved):
I got the expected full chain from the router's TLS server cert presentation.

PR #1262, opened 2023-08-23
use openziti/hello-world container image

qrkourier:


dovholuknf (approved):

dovholuknf:
this might/probably will need some doc updates. i don't think we can merge it until doc changes happen too

qrkourier:
Doc updates completed in https://github.com/openziti/ziti-doc/pull/655#discussion_r1312070208

PR #1272, opened 2023-08-30
quickstart compose configurable port forward interface

### LINKED ISSUE #1268: docker compose quickstart should not bind local ports for all attached interface addresses
The intention appears to be binding a local port like `127.0.0.1:1280`, but the Compose project binds the port on _all_ attached interfaces, which is the default behavior for Docker.

This doesn't work on my computer because some of the ports used by the Compose project are already bound on some attached interfaces, such as those used by virtual machine and container environment bridge interfaces.

There's no apparent benefit to blindly attempting to bind the ports on _all_  interfaces, so I propose that we explicitly bind _some_ loopback address like 127.0.0.1 or 127.2.171, for the giggles.

qrkourier:
continue forwarding ports from 0.0.0.0 by default and allow overriding the interface address, e.g., 127.0.21.71
resolves #1268

### DIFF
--- a/quickstart/docker/.env
+++ b/quickstart/docker/.env
@@ -7,6 +7,8 @@ ZITI_VERSION=latest
 ZITI_USER=admin
 ZITI_PWD=
 
+ZITI_INTERFACE=0.0.0.0
+# ZITI_GO_VERSION=
 # controller name, address/port information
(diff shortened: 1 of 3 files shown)


dovholuknf (approved):
probably shoulda done it to the other compose file too?

PR #1274, opened 2023-08-30
optimize quickstart Dockerfile

qrkourier:


quickstart/docker/image/Dockerfile
```diff
@@ -2,25 +2,42 @@ FROM ubuntu:rolling as fetch-ziti-bins
...
 
 COPY . /docker.build.context
-RUN apt update && apt install jq curl -y && \
+RUN apt-get update \
+    && apt-get --yes install \
+        jq \
+        curl \
+    && apt-get --yes autoremove \
+    && apt-get clean autoclean \
+    && rm -fr /var/lib/apt/lists/{apt,dpkg,cache,log} /tmp/* /var/tmp/*
```
scareything, line 14: how much difference do these changes make in the final image? I'd expect no difference at all, since these steps happen in a stage from which only the /ziti-bin directory is retained (via the COPY on line 60)

qrkourier: Correct, it does not affect the final image, and I don't feel strongly about keeping those lines in the earlier build stage. 

I included that storage-efficiency convention in the earlier build stage because I guessed it's easier to duplicate the convention everywhere an `apt-get install` is used than it is to explain why those commands aren't necessary except in the final stage of a multi-stage Dockerfile.

PR #1275, opened 2023-08-30
provide quickstart test module and a container image for running tests

qrkourier:


quickstart/test/go.mod
dovholuknf, line 1: i thought we decided to just use the go.mod/go.sum at the root of the checkout?

dovholuknf (approved):

PR #1280, opened 2023-08-31
quickstart: ignore only the correct location of the ziti-bin directory

qrkourier:
There was an inconsistency between the scripts that build the quickstart image that could cause the image to be built with unintended contents of the ziti-bin dir. The only correct location of the dir is quickstart/docker/image/ziti-bin because that is where it can be included in the Docker build context.

These changes are intended to highlight the location of any misplaced ziti-bin dirs within devs' working copies by ignoring only the correct location of ziti-bin.

### DIFF
--- a/quickstart/.gitignore
+++ b/quickstart/.gitignore
@@ -1,2 +1,3 @@
-ziti-bin
+# only ignore the correct location of the ziti-bin dir to highlight any misplace files
+/docker/image/ziti-bin
 test/gotester.json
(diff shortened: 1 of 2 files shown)


dovholuknf:
I'd rather we keep the less specific / bigger hammer.

qrkourier:
We'll accept the remote risk of unintentionally shipping a container image with the old contents of quickstart/docker/image/ziti-bin after mistakenly building the intended binaries in quickstart/docker/ziti-bin.

I see what I can do to mitigate then when I work on the release automation.

PR #1282, opened 2023-08-31
Quickstart function env race condition

qrkourier:
rebuild the env file atomically


dovholuknf (approved):
i like the atomic update/creation of the file. seems like a nice improvement.

gberl002 (approved):

PR #1285, opened 2023-09-01
add a ci job to run the Go quicker quickstart

qrkourier:
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -137,6 +137,7 @@ jobs:
     name: Fablab Smoketest
     # not applicable to forks. shouldn't run on release build
     if: github.repository_owner == 'openziti' && github.ref != 'refs/heads/main'
+    needs: [ quickstart ]
     runs-on: ubuntu-20.04
     steps:
       - name: Git Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -223,6 +224,31 @@ jobs:
...
+        uses: actions/download-artifact@v3
+        with:
+          name: linux-release-${{ github.run_id }}
+          path: release/
+
+      - name: Run Go Quickstart
+        run: |
+          for ZITI in ./release/amd64/linux/ziti; do 
+            chmod +x "$ZITI"
+            "$ZITI" edge quickstart
```
dovholuknf, line 249: this CI job will never complete. This command runs a controller and router in the foreground. It'll never exit until you ctrl-c (or exit etc). I don't think this test, as it is written, should be in the path.

qrkourier: I'll adust this branch to reflect what I learn about the test as it's being developed. I hadn't successfully run it locally yet, so don't yet understand the rationale of the quickstart command. 

Now it sounds like it's intended to run the controller and possibly a router in the foreground, waiting for input.

In that case, we probably need two processes: the quickstart in the background with a readiness check, and the quickstart_test in the foreground.

PR #1286, opened 2023-09-01
don't sleep if smoketest cancelled

qrkourier:


.github/workflows/main.yml
```diff
@@ -206,22 +206,31 @@ jobs:
...
+      # allow time for investigation unless the workflow is cancelled or the smoketest succeeded or was skipped
       - name: Sleep If Failed
-        if: needs.fablab-smoketest.result != 'success' && needs.fablab-smoketest.result != 'skipped'
+        if: needs.fablab-smoketest.result != 'success' && needs.fablab-smoketest.result != 'skipped' && needs.fablab-smoketest.result != 'cancelled'
         run: |
           sleep 30m
 
+      # release cloud resources if the smoketest succeeded, failed, or was cancelled; unnecessary if skipped
       - name: Teardown Test Environment
-        if: always() && needs.fablab-smoketest.result != 'skipped'
```
qrkourier, line 215: `always()` is workflow-level so this check is redundant to the job-level check at L205 and so never influences whether this step runs.

plorenz (approved):

PR #1296, opened 2023-09-05
Fix router config overwrite on docker compose up

gberl002:
Using router config file existence to determine whether or not to create a new config file.


quickstart/docker/image/run-router.sh
```diff
@@ -30,10 +30,11 @@ if [[ "${_ZITI_ROUTER_NAME}" != "" ]]; then
   export ZITI_ROUTER_NAME="${_ZITI_ROUTER_NAME}"
   echo "ZITI_ROUTER_NAME set to: ${ZITI_ROUTER_NAME}"
 fi
-_UNIQUE_NAME="${ZITI_HOME}/${ZITI_ROUTER_NAME}-${HOSTNAME}.init"
-if [ ! -f "${_UNIQUE_NAME}" ]; then
-  echo "system has not been initialized. initializing..."
-  "${ZITI_BIN_DIR-}/ziti" edge login ${ZITI_CTRL_EDGE_ADVERTISED_ADDRESS}:${ZITI_CTRL_EDGE_ADVERTISED_PORT} -u $ZITI_USER -p $ZITI_PWD -y
+
+"${ZITI_BIN_DIR-}/ziti" edge login ${ZITI_CTRL_EDGE_ADVERTISED_ADDRESS}:${ZITI_CTRL_EDGE_ADVERTISED_PORT} -u $ZITI_USER -p $ZITI_PWD -y
```
dovholuknf, line 34: shouldn't the login be skipped if the router config exists? seems like there's no need to login if you don't need to make a config to me?

quickstart/docker/image/run-router.sh
```diff
@@ -53,21 +54,20 @@ if [ ! -f "${_UNIQUE_NAME}" ]; then
...
+    echo " Found existing config file ${_CONFIG_PATH}, not creating a new config."
+    echo " If you would like to overwrite this config, delete the file and restart the container."
+fi
+
+found=$("${ZITI_BIN_DIR-}/ziti" edge list edge-routers 'name = "'"${ZITI_ROUTER_NAME}"'"' | grep -c "${ZITI_ROUTER_NAME}")
+if [[ found -gt 0 ]]; then
+  echo "----------  Found existing edge-router ${ZITI_ROUTER_NAME}...."
 else
-  echo "system has been initialized. starting the process."
+  "${ZITI_BIN_DIR}/ziti" edge create edge-router "${ZITI_ROUTER_NAME}" -o "${ZITI_HOME}/${ZITI_ROUTER_NAME}.jwt" -t -a "${ZITI_ROUTER_ROLES}"
```
dovholuknf, line 66: this should be done in the "if the config doesn't exist" block ?

quickstart/docker/image/run-router.sh
gberl002: The login is used later to check if the router was enrolled, if not, it enrolls.

Currently the two (check for config and enrollment) were done in one go but I figured there may be a situation where a container goes down before enrollment or enrollment fails for some reason. Separating them out will check for a config but also check for enrollment even if the config is there.

dovholuknf: Right. I don't think that's the flow we want though. If the file exists, i think we expect it's been created in the overlay like the old logic.

quickstart/docker/image/run-router.sh
gberl002: See previous comment, I can put them back in one `if`, just figured this solution was more flexible.

dovholuknf: I'd rather them delete the container/file if they want to recreate the router so yah, i think one if is best

dovholuknf (approved):

PR #1299, opened 2023-09-06
Run the quicker quickstart before fablab

qrkourier:


quickstart/docker/.env
```diff
@@ -9,6 +9,8 @@ ZITI_PWD=
 
 ZITI_INTERFACE=0.0.0.0
 # ZITI_GO_VERSION=
+# GOPATH=
```
dovholuknf, line 12: i'd rather see the test just 'cat' these lines onto the bottom of the .env file instead of having them in the 'default' .env

dovholuknf (approved):

PR #1301, opened 2023-09-07
run quickstart test as step in linux-build job

qrkourier:


.github/workflows/main.yml
plorenz, line 1: :smiley:  :+1:

PR #1312, opened 2023-09-15
get edge changes for filtering intercepts by src ip

scareything:


plorenz:
edge repo has merged up to the ziti repo. If we want this change we'll need to bring that branch up from the ziti repo

PR #1314, opened 2023-09-18
bump lib versions

qrkourier:
adopt the router extend patch


andrewpmartinez (changes requested):
PR title indicates an edge version bump, but I do not see an update go.mod nor go.sum file that would indicate the dep version was changed.

Was this just to generate the changelog?

qrkourier:
Thanks for the review. The original goal was only to bump the edge version. Along the way, I learned that we typically bump all libraries with a procedure [like this](https://github.com/openziti/ziti/pull/1316#discussion_r1332014869).

Now I've run that script to refresh all library versions.

plorenz:
superceded by #1361 which needed to update deps as well.

PR #1317, opened 2023-09-19
always create fablab instance archive

qrkourier:


.github/workflows/main.yml
```diff
@@ -209,6 +212,7 @@ jobs:
   fablab-smoketest-teardown:
     name: Teardown SmokeTest
     if: always()
```
plorenz, line 214: The timeout is shorter than the sleep, so the sleep will always trigger the timeout

qrkourier: Good catch.

plorenz (approved):

PR #1322, opened 2023-09-20
update snapshot cmd

plorenz:
- Update snapshot db command for fabric changes
- Update deps. Allow echo server to be interrupted.


qrkourier (approved):
:coin:

dovholuknf (approved):
linter failure but lgtm

PR #1331, opened 2023-09-25
add some flags to the quickstart

dovholuknf:
Adds some flags to `ziti edge quickstart` based on @ekoby wanting to try it in CI. still might need a 'stop' command
```
      --already-initialized     Specifies the PKI does not need to be created and the db does not need to be initialized. default: false
      --ctrl-address string     Sets the advertised address for the control plane and API
      --ctrl-port int16         Sets the port to use for the control plane and API
  -h, --help                    help for quickstart
      --home string             Sets the directory the environment should be installed into. defaults to a temporary directory
  -p, --password string         Password to use for authenticating to the Ziti Edge Controller. default: admin
      --persistent              Prevents the environment from being destroyed when shutting down. default: false
      --router-address string   Sets the advertised address for the integrated router
      --router-port int16       Sets the port to use for the integrated router
  -u, --username string         Username to use when creating the Ziti Edge Controller. default: admin
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,12 @@
   following flags:
   
   ```
-      --already-initialized     Specifies the PKI does not need to be created and the db does not need to be initialized. default: false
+      --already-initialized     Specifies the PKI does not need to be created and the db does not need to be initialized. Recommended to be combined with --persistent. If --persistent is not specified the environment will be destroyed on shutdown! default: false
       --ctrl-address string     Sets the advertised address for the control plane and API
       --ctrl-port int16         Sets the port to use for the control plane and API
   -h, --help                    help for quickstart
-      --home string             Sets the directory the environment should be installed into. defaults to a temporary directory
+      --home string             Sets the directory the environment should be installed into. Defaults to a temporary directory. If specified, the environment will not be removed on exit.
   -p, --password string         Password to use for authenticating to the Ziti Edge Controller. default: admin
-      --persistent              Prevents the environment from being destroyed when shutting down. default: false
       --router-address string   Sets the advertised address for the integrated router
       --router-port int16       Sets the port to use for the integrated router
   -u, --username string         Username to use when creating the Ziti Edge Controller. default: admin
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
   following flags:
   
   ```
+      --already-initialized     Specifies the PKI does not need to be created and the db does not need to be initialized. default: false
```
ekoby, line 12: does this imply `--persistent`? maybe mention this

dovholuknf: i'll clarify. it doesn't imply --persistent at this time, but if you don't supply --persistent it'll tear it all down on exit. so if you want to keep your pet instance. you'd have to:

* ziti edge quickstart --persistent
* (shut it down)
* ziti edge quickstart --persistent --already-initialized

ekoby: actually I think --home should imply persistence otherwise it always start in a new temp dir, right?

dovholuknf: if --home is not supplied it'll use a tempdir. if --home is supplied, it could default to --persistent, sure but i'm leery of adding too implicit rules around it too quickly though at this time. I'm leaning towards keeping it all explicit

dovholuknf: ok. you convinced me. i'll remove --persistent entirely and use --home instead. I thoght you were thinking to keep --persistent too but I can see removing the --persistent flag in favor of the user specifying it

CHANGELOG.md
```diff
@@ -5,8 +5,34 @@
 * `ziti edge quickstart`](https://github.com/openziti/ziti/issues/1298). You can now
   download the `ziti` CLI and have a functioning network with just one command. The
   network it creates is ephemeral and is intended to be torn down when the process exits.
-  It is intended for quick evaluation and testing of an overlay network.
-
+  It is intended for quick evaluation and testing of an overlay network. It supports the
+  following flags:
+  
+  ```
+      --already-initialized     Specifies the PKI does not need to be created and the db does not need to be initialized. Recommended to be combined with --persistent. If --persistent is not specified the environment will be destroyed on shutdown! default: false
```
ekoby, line 12: no more `--persistent`

ekoby (approved):
just cleanup usage

CHANGELOG.md
dovholuknf: good catch! thx

PR #1418, opened 2023-10-06
use generic debian package meta distro "debian"

qrkourier:
This sets the `deb.distribution` property in the Artifactory API and in the Debian packages metadata to "debian." When distributing package variants for various distributions, this property reflects the target distro, e.g., `buster`, `wheezy`, etc.

For this repo channel where we're distributing the `ziti` CLI, I mean to imply broad compatibility of artifacts in this channel with Debian-based distros, including Debian, Ubuntu, etc.

The `ziti` CLI is unlikely to demand distro-specific package variants.

If we end up distributing package variants for `openziti-controller`, we'll set `deb.distribution` appropriately so that `apt` can self-select the appropriate distro variant.

### DIFF
--- a/.github/workflows/publish-linux-packages.yml
+++ b/.github/workflows/publish-linux-packages.yml
@@ -93,7 +93,7 @@ jobs:
         shell: bash
         run: >
           jf rt upload
-          ./release/${{ matrix.package_name }}*.${{ matrix.nfpm_packager }}
+          ./release/${{ matrix.package_name }}*.rpm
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-linux-packages.yml
```diff
@@ -93,7 +93,7 @@ jobs:
         shell: bash
         run: >
           jf rt upload
-          ./release/${{ matrix.package_name }}*.${{ matrix.nfpm_packager }}
```
qrkourier, line 96: There's only one possible value and it's more readable this way.

plorenz (approved):

PR #1419, opened 2023-10-08
display the package, architecture, and format in the list of checks

qrkourier:


qrkourier:
will bundle with another PR

PR #1444, opened 2023-10-16
Db creation 2

padibona:
Should be a better PR for db-creation things, including some fixes for newer ziti versions


padibona:
Squashed

PR #1449, opened 2023-10-18
sdk hosting test

plorenz:
- Add fablab test for SDK hosting
- update for fablab changes


controller/apierror/messages.go
```diff
@@ -194,4 +194,8 @@ const (
 	EnrollmentExistsCode    string = "ENROLLMENT_EXISTS"
 	EnrollmentExistsMessage string = "ENROLLMENT_EXISTS"
 	EnrollmentExistsStatus  int    = http.StatusConflict
+
+	ServerTooManyUpdatesCode    string = "SERVER_TOO_MANY_UPDATES"
```
andrewpmartinez, line 198: It appears this rate limits any mutation command and the wording on this might be confusing to some.

"I issued a delete not an update".

It might be better to change this to "NewTooManyRequests" with the same changes to the `Code` and `Message` text. At the risk of sounding more generic at least, it won't confuse external API users who

I wold also suggest changing `ServerTooManyUpdatesMessage` not only to `ServerTooManyRequestsMessage` but also to say "Too many requests to alter state have been issues. Please slow your request rate or try again later."

andrewpmartinez (changes requested):
The OpenAPI Specs all need to be updated w/ the new HTTP Status code that can return. Depending on the client, if an invalid code is returned, it might panic or return a useless "unexpected HTTP status code" error with no way to see the actual response. At best, you get the raw response and hit a generic handler in the end client's code.

This is also not backward compatible w/ older generated clients. So, a version change is also suggested. This also will affect the GoSDK.

ziti/cmd/edge/create_identity.go
```diff
@@ -183,15 +183,18 @@ func runCreateIdentity(o *createIdentityOptions) error {
 
 	if o.jwtOutputFile != "" {
 		id := result.S("data", "id").Data().(string)
-		if err := getIdentityJwt(o, id, o.Options.Timeout, o.Options.Verbose); err != nil {
+		enrollmentType := "ott"
+		if o.username != "" {
+			enrollmentType = "updb"
+		}
+		if err = getIdentityJwt(&o.Options, id, o.jwtOutputFile, enrollmentType, o.Options.Timeout, o.Options.Verbose); err != nil {
```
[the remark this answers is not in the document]
plorenz: fixed

controller/apierror/messages.go
plorenz: fixed

plorenz:
> The OpenAPI Specs all need to be updated w/ the new HTTP Status code that can return. Depending on the client, if an invalid code is returned, it might panic or return a useless "unexpected HTTP status code" error with no way to see the actual response. At best, you get the raw response and hit a generic handler in the end client's code.
> 
> This is also not backward compatible w/ older generated clients. So, a version change is also suggested. This also will affect the GoSDK.

There's a PR up for updating the edge-api project and the fabric API has been updated here. Should we bump the version to 0.31.0?

andrewpmartinez:
> > The OpenAPI Specs all need to be updated w/ the new HTTP Status code that can return. Depending on the client, if an invalid code is returned, it might panic or return a useless "unexpected HTTP status code" error with no way to see the actual response. At best, you get the raw response and hit a generic handler in the end client's code.
> > This is also not backward compatible w/ older generated clients. So, a version change is also suggested. This also will affect the GoSDK.
> 
> There's a PR up for updating the edge-api project and the fabric API has been updated here. Should we bump the version to 0.31.0?

I would say yes.

plorenz:
> > > The OpenAPI Specs all need to be updated w/ the new HTTP Status code that can return. Depending on the client, if an invalid code is returned, it might panic or return a useless "unexpected HTTP status code" error with no way to see the actual response. At best, you get the raw response and hit a generic handler in the end client's code.
> > > This is also not backward compatible w/ older generated clients. So, a version change is also suggested. This also will affect the GoSDK.
> > 
> > 
> > There's a PR up for updating the edge-api project and the fabric API has been updated here. Should we bump the version to 0.31.0?
> 
> I would say yes.

Version has been bump to 0.31

andrewpmartinez (approved):

PR #1454, opened 2023-10-20
clarify CLI usage hints about CSV values

### LINKED ISSUE #1048: cli needs more syntax hints
Example 1: the `default []` hint implies the expected argument to `--role-attributes` is a list, but does not explain how the user should express the list. A description like "Comma-separated role attributes of the new identity" would do the trick. This is needed in output for self-help modes like `ziti edge create TYPE SUBTYPE --help`. 

Example 2: the `list` command accepts a filter, but the filter syntax and capabilities are not obvious. This is a powerful feature and a link to or name of the (standard?) query language would be supremely helpful. For example, it was a minor stumbling block for me to understand that the query must have a `key="double-quoted value"`, and that it's necessary to escape the double-quotes from the shell, so in practice you end up sending `'key="quoted value"'` or `key=\"quoted value\"`, which is functional but awkward.

qrkourier:
resolves #1048

### DIFF
--- a/ziti/cmd/edge/create_edge_router.go
+++ b/ziti/cmd/edge/create_edge_router.go
@@ -59,7 +59,7 @@ func NewCreateEdgeRouterCmd(out io.Writer, errOut io.Writer) *cobra.Command {
 
 	// allow interspersing positional args and flags
 	cmd.Flags().SetInterspersed(true)
-	cmd.Flags().StringSliceVarP(&options.roleAttributes, "role-attributes", "a", nil, "Role attributes of the new edge router")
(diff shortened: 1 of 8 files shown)


plorenz (commented):
How would you feel about writing out '... specified as a list of comma separated values'? I think what you've got is already better, but I had a mental stumbling block, where I overthought CSV. I thought maybe it meant a CSV file, or something more complicated than was intended. Thoughts?

qrkourier:
@plorenz handful of usage hint tweaks here to correct a common point of confusion about valid input for options that expect comma-separated values as a string argument

plorenz (approved):

PR #1466, opened 2023-10-25
makes index tracker tests more timing friendly

andrewpmartinez:
- adds fresh fixtures for each test
- calls out the intent of each test


plorenz (approved):
Looks good other than the linter issue

Issue #1460, opened 2023-10-24
Controller crash on nil pointer deference during - now at different location


arunsworld:
Stack trace:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1cc4311]

goroutine 13310556 [running]:
github.com/openziti/channel/v2.(*channelImpl).rxer.func1()
	/go/pkg/mod/github.com/openziti/channel/v2@v2.0.99/impl.go:308 +0x3f
panic({0x2113340?, 0x4196650?})
	/usr/local/go/src/runtime/panic.go:914 +0x21f
github.com/openziti/ziti/controller/sync_strats.(*InstantStrategy).ReceiveClientHello(0xc00204ab00, 0xc002e390e0, 0xc002514000)
	/go/src/github.com/openziti/ziti/controller/sync_strats/sync_instant.go:500 +0x1051
github.com/openziti/ziti/controller/handler_edge_ctrl.(*helloHandler).HandleReceive(0xc001d6c3a0, 0xc000ac6330, {0x2bc82e0, 0xc00273f500})
	/go/src/github.com/openziti/ziti/controller/handler_edge_ctrl/hello.go:58 +0x4a8
github.com/openziti/channel/v2.(*channelImpl).rxer(0xc0003e3680?)
	/go/pkg/mod/github.com/openziti/channel/v2@v2.0.99/impl.go:357 +0x571
created by github.com/openziti/channel/v2.(*channelImpl).startMultiplex in goroutine 56
	/go/pkg/mod/github.com/openziti/channel/v2@v2.0.99/impl.go:288 +0xc7
```

Branch: fix-router-connect-panic
Part of the longevity test to check if issue #1423 was fixed.

I see that there is null checking above for logging but in line 500 it's unchecked.

If I may also suggest, a lot of the code I've looked at is very "Java". Setters and Getters and self. if it were more idiomatic Go, it would be easier on Go devs and leverage the simplicity and semantic principles foundational in the language.

plorenz:
Hi @arunsworld ,
   Would you be willing to run a build from a debug branch, so we can get some more information? I've spent a few hours digging through the code and I can't see a way for the VersionInfo to be nil there. 

[...]

PR #1483, opened 2023-10-31
Add additional logging. Remove unnecessary router lookups. May addres…

plorenz:
…s #1460

### DIFF
--- a/controller/handler_ctrl/accept.go
+++ b/controller/handler_ctrl/accept.go
@@ -59,64 +59,75 @@ func (self *CtrlAccepter) Bind(binding channel.Binding) error {
 	ch := binding.GetChannel()
 
 	log := pfxlog.Logger().WithField("routerId", ch.Id())
-
 	// Use a new copy of the router instance each time we connect. That way we can tell on disconnect
 	// if we're working with the right connection, in case connects and disconnects happen quickly.
 	// It also means that the channel and connected time fields don't change and we don't have to protect them
-	if r, err := self.network.GetReloadedRouter(ch.Id()); err == nil {
-		if ch.Underlay().Headers() != nil {
-			if versionValue, found := ch.Underlay().Headers()[channel.HelloVersionHeader]; found {
-				if versionInfo, err := self.network.VersionProvider.EncoderDecoder().Decode(versionValue); err == nil {
-					r.VersionInfo = versionInfo
-				} else {
-					return errors.Wrap(err, "could not parse version info from router hello, closing router connection")
-				}
+	r, err := self.network.GetReloadedRouter(ch.Id())
+	if err != nil {
+		return err
+	}
+	if r == nil {
+		return errors.Errorf("no router with id [%v] found, closing connection", ch.Id())
+	}
+
+	if ch.Underlay().Headers() != nil {
+		if versionValue, found := ch.Underlay().Headers()[channel.HelloVersionHeader]; found {
+			if versionInfo, err := self.network.VersionProvider.EncoderDecoder().Decode(versionValue); err == nil {
+				r.VersionInfo = versionInfo
+				log = log.WithField("version", r.VersionInfo.Version).
+					WithField("revision", r.VersionInfo.Revision).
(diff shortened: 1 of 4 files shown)


controller/handler_ctrl/accept.go
```diff
@@ -59,64 +59,75 @@ func (self *CtrlAccepter) Bind(binding channel.Binding) error {
...
-							WithField("costTags", listener.GetCostTags()).
-							Debug("router listener")
-					}
-				}
+	r, err := self.network.GetReloadedRouter(ch.Id())
+	if err != nil {
+		return err
+	}
+	if r == nil {
+		return errors.Errorf("no router with id [%v] found, closing connection", ch.Id())
```
dovholuknf, line 70: "closing connection" is not actually happening here and is misleading imo.

controller/handler_ctrl/accept.go
```diff
@@ -59,64 +59,75 @@ func (self *CtrlAccepter) Bind(binding channel.Binding) error {
...
 		} else {
 			return errors.New("no version info header, closing router connection")
 		}
 
-		r.Control = ch
-		r.ConnectTime = time.Now()
-		if err := binding.Bind(newBindHandler(self.heartbeatOptions, r, self.network, self.xctrls)); err != nil {
-			return errors.Wrap(err, "error binding router")
+		r.Listeners = nil
+		if val, found := ch.Underlay().Headers()[int32(ctrl_pb.ContentType_ListenersHeader)]; found {
```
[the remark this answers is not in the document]
dovholuknf: (i also see now that is how it always was, originally i thought tihs was new code so... nevermind prolly)

controller/handler_ctrl/accept.go
```diff
@@ -59,64 +59,75 @@ func (self *CtrlAccepter) Bind(binding channel.Binding) error {
...
+				r.SetLinkListeners(listeners.Listeners)
+				for _, listener := range listeners.Listeners {
+					log.WithField("address", listener.GetAddress()).
+						WithField("protocol", listener.GetProtocol()).
+						WithField("costTags", listener.GetCostTags()).
+						Debug("router listener")
+				}
+			}
+		} else {
+			log.Info("no advertised listeners")
```
dovholuknf, line 104: is this really an info type of message?

controller/handler_ctrl/accept.go
```diff
@@ -59,64 +59,75 @@ func (self *CtrlAccepter) Bind(binding channel.Binding) error {
...
+	if err := binding.Bind(newBindHandler(self.heartbeatOptions, r, self.network, self.xctrls)); err != nil {
+		return errors.Wrap(err, "error binding router")
+	}
 
-		self.network.ConnectRouter(r)
+	if self.traceHandler != nil {
+		binding.AddPeekHandler(self.traceHandler)
 	}
+
+	log.Info("accepted new router connection")
```
dovholuknf, line 128: why lose the existing, presumably useful, `r.Id`? is the `[r/Id]` not useful?

dovholuknf (approved):
nothing looks like it shoudl stop the pr, just some questions here/there

controller/handler_ctrl/accept.go
plorenz: changed to 'not accepting' so it's more accurate.

controller/handler_ctrl/accept.go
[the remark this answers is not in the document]
plorenz: ctrl_pb.ContentType_ListenersHeader is a ContentType which is an alias for an int32, so the cast is fine.

controller/handler_ctrl/accept.go
plorenz: There's already a logger field for routerId added at the top, so this was redundant.

controller/handler_ctrl/accept.go
plorenz: It was Warn and I dropped it to Info. Dropped it further to Debug.

PR #1518, opened 2023-11-17
add some steps to HA dev setup

qrkourier:


doc/ha/dev-setup.md
```diff
@@ -70,32 +70,54 @@ ziti pki create server --pki-root ./pki --ca-name ctrl3 --dns localhost --ip 127
...
+    ```bash
+    ziti controller run ctrl1.yml
+    ziti controller run ctrl2.yml
+    ziti controller run ctrl3.yml
+    ```
+
+    All three are configured with `minClusterSize` of 3, so they will wait to be joined to a raft cluster.
+    The ctrl1.yml config file has the other two controllers as bootstrap members, so when it starts the first
+    controller will start trying form the raft cluster.
+
```
plorenz, line 84: Once this: https://github.com/openziti/ziti/pull/1524 is merged, the `cluster add` commands should no longer be necessary, so we can probably either take it out or make a note that you could use this method instead of the bootstrapMembers config option.

qrkourier: I moved those down to the bottom as a troubleshooting note.

plorenz (approved):

PR #1524, opened 2023-11-20
Retry bootstrap members until bootstrapped. Fixes #1523

### LINKED ISSUE #1523: Bootstrap members not working
If raft bootstrap members are configured in the config file, and they aren't available on startup, the controller doesn't retry. Should retry until we're successful or until the cluster has been bootstrapped.

plorenz:


plorenz:
recreating

PR #1526, opened 2023-11-20
Tweaks ha docs based on feedback.

plorenz:


plorenz:
> I'd sent some changes for one of the same files, `doc/ha/dev-setup.md`, in [this other PR](https://github.com/openziti/ziti/pull/1518). Was `ha-staging` the wrong destination branch for those?

Either is ok. The ha-staging branch will need to pull changes from release-next regularly, so they should get picked up from either spot.

qrkourier (approved):

PR #1527, opened 2023-11-21
HA controller config

ekoby:
also makes quickstart run with HA controller

### DIFF
--- a/ziti/cmd/create/config_templates/controller.yml
+++ b/ziti/cmd/create/config_templates/controller.yml
@@ -9,6 +9,17 @@ v: 3
 
 db:                     "{{ .ZitiHome }}/db/ctrl.db"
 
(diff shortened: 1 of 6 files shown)


ziti/cmd/create/config_templates/controller.yml
```diff
@@ -211,6 +223,8 @@ web:
         # These settings are per API. The example below is for the 'edge-api' and contains both optional values and
         # required values.
         options: { }
+      - binding: edge-oidc
```
dovholuknf, line 226: should this be if'ed the same way that lines 11-15 are?

dovholuknf (approved):

ziti/cmd/create/config_templates/controller.yml
ekoby: I am pretty sure you can have oidc without HA

PR #1529, opened 2023-11-21
clarify a log warning about mismatched router ports behind an LB

qrkourier:


router/internal/edgerouter/config.go
```diff
@@ -310,7 +310,7 @@ func parseEdgeListenerOptions(index int, address string, edgeListenerMap map[int
 			}
 
 			if advertisePort != addressPort {
-				pfxlog.Logger().Warnf("port in [listeners[%d].options.advertise] must equal port in [listeners[%d].address] for edge binding but did not. Got [%d] [%d]", index, index, advertisePort, addressPort)
+				pfxlog.Logger().Warnf("advertisement [listeners[%d].options.advertise] port [%d] does not match the listening port in [listeners[%d].address][%d]. This is expected if Ziti Router is behind a load balancer routing the advertised port to the listening port.", index, advertisePort, index, addressPort)
```
dovholuknf, line 313: let's keep it simple and just state the port is not the same. adding in the load balancer bit is meaningless to someone that knows they changed the port and for someone that didn't realize it, it's meaningless for them too

qrkourier: Okay. My main goal is to stop saying "must equal," but maybe we should stop warning the user entirely, and delete this check and log message when there's no reason to expect that these ports match in a deployed environment where ports are likely to be forwarded or L4 load balanced.

dovholuknf: Yes, I agree with that for sure. It might be intentional.

qrkourier: Demoted from WARNING to DEBUG and minimized to merely state the fact of the mismatch.

dovholuknf: it can't be a debug, it needs to at least be an INFO

dovholuknf (approved):

PR #1534, opened 2023-11-22
add loopback IP SAN to quickstart

qrkourier:


dovholuknf (approved):

ekoby:
also add IPv6 loopback

qrkourier:
> also add IPv6 loopback

There seems to be an issue with the representation of the `::1` value in the resulting  x509 certificate.

got

```
            X509v3 Subject Alternative Name: 
                IP Address:127.0.0.1, IP Address:0:0:0:0:0:0:0:6
```

expected

```
            X509v3 Subject Alternative Name: 
                IP Address:127.0.0.1, IP Address: ::1
```

I ran:

```
ziti pki create server --allow-overwrite --ip 127.0.0.1,'::6'
```

Nonetheless, I'll modify this branch to also add IP SAN `::1`, so it'll work when we fix the way its parsed.

PR #1536, opened 2023-11-24
add minimal docker quickstart

qrkourier:


qrkourier:
Video tour of this branch: https://youtu.be/XlscY29K03s?feature=shared

ziti/cmd/edge/quickstart.go
```diff
@@ -42,6 +42,7 @@ import (
 type QuickstartOpts struct {
 	Username           string
 	Password           string
+	InitOnly           bool
```
dovholuknf, line 45: I am reluctant to turn this into a 'kitchen sink' type of command that allows one to update/configure everything. With that type of lens, I don't know who would want to "InitOnly" or why. What's the driver here, who would use this and why?

qrkourier: A simple set of command-line options is sufficient for what I have in mind, as long as I can edit the generated configs, and other state files, and restart the quickstart. That's where the main use case for `docker compose up` comes in.

The same approach may have a lesser value for someone who's developing Ziti itself, so I added a `--build` option to support that.

The main value seems to be running a mostly-default, minimal, local Ziti network, so I can use it with whatever I'm building on the platform, like zrok.

qrkourier: > who would want to "InitOnly"

Oh, I needed this to automate bringing the quickstart because there is no way to express the intent "initialize if not initialized and run." 

Without `--init-only`, I could only say "fail if already initialized, else initialize and run."

dovholuknf: for me, I'm not sold that this is a good enough idea. we'll need to discuss

qrkourier: Resolving this one because I removed `--init-only` after talking it over with @dovholuknf. That means it's necessary to wait until the router has synchronized API sessions and the Docker project is simpler because now there's no init-only step, which was a separate init container.

qrkourier:
This compose file provides a local-only Ziti network by default, and can be made available to the LAN or globally by setting `EXTERNAL_DNS`.

For example, if your Docker host's IP address is 192.168.21.71, and the Docker host is accepting 1280,3022/tcp on that IP address, this provides the Ziti network to all hosts that can route to that IP.

```bash
export ZITI_QUICK_TAG=release-next                                                          
docker compose pull &&
    EXTERNAL_DNS=ziti.192.168.21.71.sslip.io \
        docker compose up --detach
```

The `release-next` tag override is necessary until the next Ziti release.

dist/cloudfront/get.openziti.io/routes.yml
```diff
@@ -38,3 +38,7 @@
 - get: /zdew/
   raw: /openziti/desktop-edge-win/main/release-streams/
   file: latest.json
+
+- get: /minimal/
```
dovholuknf, line 42: are we capable of using something like `/dock/minimal` here? (two path segments)? I'd like to see this under the existing `dock`

quickstart/docker/minimal/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM debian:bookworm-slim
```
dovholuknf, line 1: why isn't this extending ziti-cli? seems like it should be?

FROM ${ZITI_CLI_IMAGE}:${ZITI_CLI_TAG}

quickstart/docker/minimal/compose.yml
```diff
@@ -0,0 +1,50 @@
...
+      ZITI_CMD+=" --ctrl-address ${EXTERNAL_DNS:-127.0.0.1}"\
+      " --router-address ${EXTERNAL_DNS:-127.0.0.1}"\
+      " --password ${ZITI_PWD:-admin}"
+      echo "DEBUG: run command is: ziti $${@} $${ZITI_CMD}"
+      exec ziti "$${@}" $${ZITI_CMD}
+    command: -- edge quickstart --home /persistent
+    user: ${ZIGGY_UID:-1000}
+    environment:
+      HOME: /persistent
+      PFXLOG_NO_JSON: "true"
```
dovholuknf, line 40: Maybe change this to `${PFXLOG_NO_JSON:-true}` ?

quickstart/docker/minimal/compose.yml
```diff
@@ -0,0 +1,50 @@
+services:
+  initialize:
+    image: busybox
+    command: chown -Rc ${ZIGGY_UID:-1000} /persistent
+    user: root
+    environment:
+      HOME: /persistent
+      # PFXLOG_NO_JSON: "true"
+    volumes:
+    - ./persistent:/persistent
```
dovholuknf, line 10: this is compose, can the example make a volume and reference that instead of a mount like this?

quickstart/docker/minimal/compose.yml
```diff
@@ -0,0 +1,50 @@
...
+  quickstart:
+    depends_on:
+      initialize:
+        condition: service_completed_successfully
+    image: ${ZITI_QUICK_IMAGE:-docker.io/openziti/ziti-cli}:${ZITI_QUICK_TAG:-latest}
+    restart: unless-stopped
+    build:
+      context: ${ZITI_SRC_ROOT:-../../../}
```
dovholuknf, line 18: what's all this for? as an example, this is confusing

dist/cloudfront/get.openziti.io/routes.yml
qrkourier: I removed the additional shortcut route. The existing directory shortcut `get.openziti.io/dock` to `/quickstart/docker` in this repo already provides a getter URL like get.openziti.io/dock/minimal/compose.yml`.

quickstart/docker/minimal/Dockerfile
qrkourier: The purpose of this Dockerfile is to support a secondary use case for the minimal quickstart: local development environment running the local build of the `ziti` executable from the `/build/ziti` path at the top-level of the checked-out Git working copy.

This means the Dockerfile is only used for running the local build of `ziti`, and has no relationship to the `ziti-cli` container image, which is used for delivering CI builds of `ziti`.

The only requirements for this Linux container image are that it has a small footprint and it has a version of GLIBC that is fairly new, because the `zrok` binary that is built locally will be linked to the `libc` installed on the developer's workstation. The `ziti-cli` image, in contrast, has a `ziti` binary that is intentionally linked to an older `libc` for broader compatibility. This means the locally-built `ziti` binary will likely not be executable inside the `ziti-cli` container image.

quickstart/docker/minimal/compose.yml
qrkourier: The primary use case I have in mind for this minimal quickstart example is an immediately up-and-running Ziti network to support Ziti edge development, i.e., things that run on Ziti, like edge SDKs and zrok.

We could provide a commented example of a named volume without changing the use case. One of my goals for the primary use case is a bind-mounted persistence directory to enable the developer to adjust the Ziti controller and router settings directly without the fuss of piping commands through Docker or starting over from scratch with new env var settings.

michaelquigley: I use this daily.

dovholuknf: I'd like people who download it to get a volume mount for their stuff. if they want to change it to a directory mount like this, i'd like to let them decide that.

quickstart/docker/minimal/compose.yml
qrkourier: Added the following explanation to [the applicable section of the README](https://github.com/openziti/ziti/blob/minimal-docker-quickstart/quickstart/docker/minimal/README.md?plain=1#L47).

> By adding this `--build` option to the `up` command, the container image is built from the Dockerfile with your locally built `ziti` binary instead of pulling the default `openziti/ziti-cli` container image from Docker Hub. In the `compose.yml`, the Docker build context is defined with environment variable `ZITI_SRC_ROOT` which defaults to `../../../` (three levels up from this directory at the top level of a Git working copy of the source repo).

dovholuknf (approved):

dovholuknf:
it probably should be changed to 'dev' or somethign like that too not 'minimal' imo

quickstart/docker/minimal/compose.yml
qrkourier: I removed from this branch the additional shortcut route that I'd proposed like `get.openziti.io/minimal/compose.yml`. The existing directory shortcut `get.openziti.io/dock` to `/quickstart/docker` in this repo already provides a getter URL like `get.openziti.io/dock/minimal/compose.yml`. 

As for named vs. bind volume mount, using a named volume works but requires an extra step that every adopter must perform before they can enjoy an instant Ziti network. I've documented the extra step in a comment inside the `compose.yml` like so.

```yaml
    volumes:
    # store the quickstart state on the Docker host in the same directory as this compose.yml file
    - ./persistent:/persistent
    # store the quickstart state in a named volume; you must comment the bind mount above, uncomment the named mount
    # below, and set permissions on the named volume if you choose this alternative, e.g.:
    #       docker compose run --entrypoint='' --user=0 quickstart chmod -Rc 1777 /persistent
    # - persistent:/persistent
```

The bind mount is the best default configuration for persistence because it doesn't require the extra step to set permissions on the named volume (set sticky bit as root so any UID can mkdir and thenceforth manage their own files).

qrkourier:
> it probably should be changed to 'dev' or somethign like that too not 'minimal' imo

How about "naked ziti," as in pasta that's not dressed with sauce?

PR #1539, opened 2023-11-27
Retry bootstrap members until bootstrapped. Fixes #1523

### LINKED ISSUE #1523: Bootstrap members not working

plorenz:


controller/raft/raft.go
```diff
@@ -694,6 +695,34 @@ func (self *Controller) addConfiguredBootstrapMembers() error {
 	return nil
 }
 
+func (self *Controller) retryBootstrapMember(bootstrapMember string) {
```
ekoby, line 698: do we need upper limit on how long to try?

plorenz: I feel  like no. The system is unusable until it's bootstrapped, so we'll keep trying until they stop us or we get bootstrapped. Once we're bootstrapped we'll stop, even if the configured cluster members were never reached.

PR #1542, opened 2023-11-29
Prototype zac xweb handler

dovholuknf:
with this xweb implemenation, an SPA such as the new ziti console coming would be able to be put into some directory locally and served from the controller. This makes it even easier to setup a controller, router and zac for development type work.

If this looks good, I'll add it to the changelog, 

One would:

* update the controller config with this API binding:
  ```
      - binding: zac
        options: { "location":"C:\\Users\\clint\\Downloads\\zac" }
        -- or --
      - binding: zac
        options: { "location": "/tmp/zac-files" }
  ```
* start the controller (for example: `ziti edge quickstart --home /tmp/quickstart-debug --already-initialized`
* access ZAC at https://localhost:1280/zac

Example:
zac-via-ctrl


controller/controller.go
```diff
@@ -247,6 +248,10 @@ func (c *Controller) initWeb() {
 		logrus.WithError(err).Fatalf("failed to create metrics api factory")
 	}
 
+	if err := c.xweb.GetRegistry().Add(zac.NewZitiAdminConsoleFactory()); err != nil {
+		logrus.WithError(err).Fatalf("failed to create myXweb factory")
```
andrewpmartinez, line 252: Probably don't want this error message to be "myXweb factory"

common/zac/handler.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Falls back to a supplied index (indexFile) when either condition is true:
+// (1) Request (file) path is not found
+// (2) Request path is a directory
+// Otherwise serves the requested file.
+func (h *spaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	r.URL.Path = strings.TrimPrefix(r.URL.Path, h.contextRoot)
+	p := filepath.Join(h.content, filepath.Clean(r.URL.Path))
+
+	if info, err := os.Stat(p); err != nil {
```
andrewpmartinez, line 105: This order of operations will allow an attack to use this endpoint to see what files on the host are accessible by the user hosting the controller as os.Stat is operating on `p` which isn't protected against for parent traversal or absolute paths. `p` is assumed to be a relative path.

`http.ServeFile` will respond as follows:
- path exists and is readable - http.Server will handle and then return a 400 bad request
- path exits and is not readable - redirects to index

To address this URLs should be cleaned before os.Stat is called by 
1) returning an error for any path with ".." in it
2) return an error for any path starting with `/` or `<driveLetter>:`

andrewpmartinez (changes requested):
See comments in `ServeHTTP` as critical.

The suggestion of making this a generic API factory is not as critical but still suggested.

common/zac/handler.go
andrewpmartinez: It looks like you are joining to the root of the static files, so that should handle absolute root traversal. Just not `..`

common/zac/handler.go
```diff
@@ -0,0 +1,120 @@
+/*
```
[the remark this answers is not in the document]
andrewpmartinez: Options are parsed as part of the overall configuration which is just YAML. The options sections are parsed as `map[any]any` so as long as it fits in YAML, it can be put in the options as long as the code that reads it understands it.

ziti/cmd/create/config_templates/controller.yml
```diff
@@ -215,3 +215,5 @@ web:
         options: { }
       - binding: fabric
         options: { }
+      - binding: zac
```
qrkourier, line 218: `binding: console`, because initialisms make poor names when a meaningful name is available and unambiguous because they require definition or explanation.

plorenz (approved):

common/zac/handler.go
dovholuknf: Excellent catch. I'd tried most of my tooling and went out of my way to find a utility that would actually send the `..` in the http request. curl wouldn't do it and all the browsers I tried wouldn't either. go is already using that containsDotDot function to protect people. Proof is here shown below. httppie. this HTTP response is returned from `func ServeFile(w ResponseWriter, r *Request, name string) {` in`go/pkg/mod/golang.org/toolchain@v0.0.1-go1.21.5.linux-amd64/src/net/http/fs.go`
```
$ http -v --verify=no https://192.168.1.83:1280/zac/..
GET /zac/.. HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: 192.168.1.83:1280
User-Agent: HTTPie/2.4.0



HTTP/1.1 400 Bad Request
Content-Encoding: gzip
Content-Length: 41
Content-Type: text/plain; charset=utf-8
Date: Wed, 28 Feb 2024 02:58:46 GMT
X-Content-Type-Options: nosniff

invalid URL path
```

If you still feel strongly about adding an additional check in our code, I can. I'm going to consider this resolved and I'll directly follow up wit you to see if this meets your needs or if you want the extra safety.

PR #1548, opened 2023-12-01
Support filter/sort/page on circuits and links. Fixes #1547

### LINKED ISSUE #1547: Support filtering, sorting and paging circuits and links
Circuits and links aren't filterable, sortable or pageable because they're stored in memory, not in bbolt. Support the same querying primitives as we do for entities stored in bolt.

plorenz:


controller/network/link_controller.go
```diff
@@ -31,18 +32,65 @@ type linkController struct {
...
 		linkTable:      newLinkTable(),
 		idGenerator:    idgen.NewGenerator(),
 		initialLatency: initialLatency,
 	}
+
+	result.store = objectz.NewObjectStore[*Link](func() objectz.ObjectIterator[*Link] {
+		return IterateCMap[*Link](result.linkTable.links)
+	})
+
+	result.store.AddStringSymbol("id", func(entity *Link) *string {
```
andrewpmartinez, line 54: It seems odd that these are outside of the store creation. Does this type of store have multiple instances?

andrewpmartinez (approved):

controller/network/link_controller.go
plorenz: Yes, there isn't a store type per entity here. The store is generic, it just needs symbols and an entity iterator.

PR #1575, opened 2023-12-11
Add support adding/removing members and transferring leadership via REST API. Fixes #1544 and fixes #1543

### LINKED ISSUE #1543: Support add/remove raft peer via REST
Note that REST is not available until after the cluster is bootstrapped, so this won't work for all cases

### LINKED ISSUE #1544: Support transfer raft leadership via REST

plorenz:


andrewpmartinez:
Fixes #1544 and fixes #1543

controller/apierror/messages.go
```diff
@@ -198,4 +198,12 @@ const (
 	ServerTooManyRequestsCode    string = "SERVER_TOO_MANY_REQUESTS"
 	ServerTooManyRequestsMessage string = "Too many requests to alter state have been issued. Please slow your request rate or try again later."
 	ServerTooManyRequestsStatus  int    = http.StatusTooManyRequests
+
+	ServerNotRunningInClusterModeCode    string = "SERVER_NOT_RUNNING_IN_CLUSTER_MODE"
```
andrewpmartinez, line 202: Cluster mode? HA mode? I have a capability in the edge section called `HA_CONTROLLER`.

I think we should try to be consistent.

plorenz: Makes sense, I'll change it to HA

plorenz: Done

PR #1584, opened 2023-12-13
Fix stall when checking acks after checking window size. Fixes #1583

### LINKED ISSUE #1583: xgress: Potential data stall due when processing acks after checking window size
If we check the window size, and decide to block incoming data because the window size has been exceeded, then process acks before entering the select loop, we can have shrunk the outstanding data size, but still think the window size has been exceeded and not allow more data to flow through.

plorenz:


michaelquigley (approved):
:ship:

PR #1587, opened 2023-12-14
Fix stall cause by ack arriving before payload is received by link send buffer. Fixes #1586

### LINKED ISSUE #1586: If ack is received before payload is processed by link send buffer, a stall can result
If ack is received before payload is processed by link send buffer, a stall can result. The ack will be viewed as a duplicate ack, since we don't have a record of the payload yet. The payload will then need to be retransmitted.

plorenz:


michaelquigley (approved):
Seems reasonable... :+1: :ship:

PR #1688, opened 2024-01-17
test the all-in-one quickstart

qrkourier:


quickstart/docker/image/Dockerfile
```diff
@@ -19,6 +18,7 @@ RUN apt-get update \
     #   just build the Dockerfile \
     # to use a specific version of ziti, specify ZITI_VERSION_OVERRIDE then  \
     #   build the Dockerfile
+COPY . /docker.build.context
```
qrkourier, line 21: after APT so we don't have to wait for it every time the build context changes when iterating locally

quickstart/docker/image/run-router.sh
```diff
@@ -72,5 +72,5 @@ fi
 unset ZITI_USER
 unset ZITI_PWD
 
-"${ZITI_BIN_DIR}/ziti" router run "${ZITI_HOME}/${ZITI_ROUTER_NAME}.yaml" > "${ZITI_HOME}/${ZITI_ROUTER_NAME}.log"
+"${ZITI_BIN_DIR}/ziti" router run "${ZITI_HOME}/${ZITI_ROUTER_NAME}.yaml"
```
qrkourier, line 75: use Docker's built in log facility for both stdout and stderr instead of splitting stderr to Docker log and stdout to a file

quickstart/docker/image/ziti-cli-functions.sh
```diff
@@ -40,12 +40,12 @@ function _wait_for_controller {
 }
 
 function _wait_for_public_router {
-  local advertised_host_port="${ZITI_ROUTER_NAME}:${ZITI_ROUTER_PORT}"
+  local advertised_host_port="${ZITI_ROUTER_ADVERTISED_ADDRESS}:${ZITI_ROUTER_PORT}"
```
qrkourier, line 43: this only worked when the entity name and domain name of the router happened to match

quickstart/test/compose-test.zsh
```diff
@@ -39,9 +39,9 @@ TESTDIR="$(mktemp -d -t "${BASENAME%.*}.${DATESTAMP}.XXX")"
 if [[ -z "${ZITI_QUICK_DIR:-}" ]]; then
     ZITI_QUICK_DIR="$(realpath "${DIRNAME}/..")"
 fi
-# if unset, set ZITI_QUICK_IMAGE_TAG to this run's dirname
-if [[ -z "${ZITI_QUICK_IMAGE_TAG:-}" ]]; then
-    ZITI_QUICK_IMAGE_TAG=$(basename "${TESTDIR}")
+# if unset, set ZITI_QUICK_TAG to this run's dirname
```
qrkourier, line 42: adopt the name used by the all-in-one quickstart for consistency

quickstart/test/compose-test.zsh
```diff
@@ -68,35 +68,37 @@ if [[ -n "${ZITI_QUICK_DIR:-}" ]]; then
...
+    echo "ERROR: ZITI_QUICK_TAG is not set, try running with --local" >&2
     exit 1
 fi
 
 # rename the simplified Compose file to the default Compose project file name
 mv ./simplified-docker-compose.yml ./compose.yml
 
 # learn the expected Go version from the Go mod file so we can pull the correct container image
-ZITI_GO_VERSION="$(awk '/^go[[:space:]]+/ {print $2}' ./go.mod)"
+: ${ZITI_GO_VERSION:="$(awk '/^go[[:space:]]+/ {print $2}' ./go.mod)"}
```
qrkourier, line 95: use parent var if defined

quickstart/test/compose-test.zsh
```diff
@@ -68,35 +68,37 @@ if [[ -n "${ZITI_QUICK_DIR:-}" ]]; then
...
 # make this var available in the Compose project
 sed -E \
-    -e  "s/^(#[[:space:]]+)?(ZITI_PWD)=.*/\2=${ZITI_PWD}/" \
-    -e  "s/^(#[[:space:]]+)?(ZITI_INTERFACE)=.*/\2=${ZITI_INTERFACE:-127.0.0.1}/" ./.env > ./.env.tmp
+    -e  "s/^(#[[:space:]]*)?(ZITI_PWD)=.*/\2=${ZITI_PWD}/" \
```
qrkourier, line 98: set router entity name and router address in compose env, and fix regex that was failing to match commented lines

quickstart/test/compose-test.zsh
```diff
@@ -153,7 +155,6 @@ docker compose exec ziti-controller \
     '
         # TODO: re-add cert checks to above test suite after https://github.com/openziti/ziti/pull/1278
         # zsh /persistent/check-cert-chains.zsh;
-docker compose run quickstart-test
+docker compose --profile test run --rm quickstart-test
```
qrkourier, line 158: move test to a compose profile so it only runs when deliberately invoked, not every time the project is `up`'d

qrkourier:
I moved this from read to draft because I found a problem with the default router entity name convention and an opportunity to simplify.

qrkourier:
I'm diagnosing the failed check.

ziti/cmd/edge/quickstart_automated_test.go
```diff
@@ -15,7 +15,6 @@ import (
 func TestEdgeQuickstartAutomated(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	_ = os.Setenv("ZITI_CTRL_EDGE_ADVERTISED_ADDRESS", "localhost") //force localhost
-	_ = os.Setenv("ZITI_ROUTER_NAME", "quickstart-router")
```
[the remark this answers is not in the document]
qrkourier: I misunderstood this. It needs to be set here because this is the automated test which is not designed to accept vars from parent env, but rather must define all vars needed to test in a self-contained manner.

quickstart/docker/all-in-one/README.md
```diff
@@ -16,13 +19,18 @@ This is the primary use case for this project: running the `ziti edge quickstart
...
 
     ```bash
-    docker compose up --detach
+    docker compose up
     ```
 
-3. Modify the state in `./persistent/`, and bounce the container.
+3. Modify configuration and bounce the container.
+
+    If you set `ZITI_HOME=./persistent`, then you would modify the configs in `./persistent/` on the Docker host.
```
dovholuknf, line 27: can we reword this. something like:
> By default the compose file uses will place files into a ./persistent folder adjacent to where you run docker compose from. Set `ZITI_HOME` to change the location where the image stores the persistet files.

something like that?

quickstart/test/compose-test.zsh
dovholuknf: > use parent var if defined

i'd rather see an echo statement here regardless of how it was found. ideally it should tell you if it's getting it from the environment or if it determining it by reading the go.mod on its own

dovholuknf (approved):

quickstart/test/compose-test.zsh
qrkourier: This change was moved to [another branch](https://github.com/openziti/ziti/pull/1749) when this branch's changes were split to make them easier to review.

quickstart/docker/all-in-one/README.md
qrkourier: That would be accurate if inverted like this:

- default: named volume managed by Compose, 
- alternative: bind volume from Docker host's filesystem

> By default, the Compose project places files into `/persistent` folder in a volume that's mounted on the container and managed by Compose. Set `ZITI_HOME` to mount any folder from the Docker host that's writeable by the logged-in user on the container instead of using the Docker-managed volume.

quickstart/docker/all-in-one/Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM debian:bookworm-slim
 
-COPY ./build/ziti /usr/local/bin/
+ARG ARTIFACTS_DIR=./build
```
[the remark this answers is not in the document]
qrkourier: This container image is much newer than the ziti-cli image and is needed because it can run the ziti binary that was built on the Docker host, which typically has a much newer glibc.

Put another way, if we use the ziti-cli image instead of this new Dockerfile, then we also must build the ziti binary in that image (or one with the same vintage of glibc), which is slower and impedes the developer who is using this all-in-one Compose project to run the locally-built ziti binary.

PR #1712, opened 2024-01-29
Ziti edge service helper

gberl002:
Create the configs, service, and service policies for an endpoint using one cli command.

Example usages

Create dial and bind setup for `localhost:1280` using `tcp`, the service is called `myservice`, with an endpoint at `myservice.ziti` (defaulted to servicename.ziti)
```ziti edge secure myservice tcp:localhost:1280```

Create dial and bind setup for `localhost:1280` using tcp, the service is called `myervice`, with a custom endpoint at `myendpoint.ziti`
`ziti edge secure myservice tcp:localhost:1280 --endpoint myendpoint.ziti`

Create dial and bind setup for `127.0.0.1:1280` using "tcp,udp" protocols, the service is called `myservice`, with an endpoint at `myservice.ziti`. When not provided, allowed protocols defaults to both `tcp` and `udp`, and the assumed address is `127.0.0.1`. At a minimum, port must always be provided.
`ziti edge secure myservice 1280`

### CODE REVIEW

File: ziti/cmd/edge/quickstart.go [hunk 1]
```diff
@@ -383,3 +384,18 @@ func waitForController(ctrlUrl string, done chan struct{}) {
 	}
 	done <- struct{}{}
 }
+
+func IsPortListening(host string, port int, timeout time.Duration) bool {
```
dovholuknf, line 388: function doesn't seem to be used?

qrkourier (changes requested):

dovholuknf:
closing the pr for now

PR #1716, opened 2024-01-29
link mgmt race conditions

### LINKED ISSUE #1709: Fix link management race conditions found by chaos testing
1. Fix issue where messages from router can be processed after router has disconnected or reconnected
2. Fix issue where async processing of router link notification can happen after async notification of link fault by making router link processing synchronous
3. Ensure we're comparing routers by id, in case instances have changed
4. Keep trying to send faults to controllers until they've been disconnected for a while. This avoids issues with controllers missing faults if they're only briefly disconnected
5. Ensure only one dial for a link can happen at a given time, since link status can changed due to incoming links
6. Cleanup partial links if first part of splink link fails bind
7. Fix issue with missed link notification when notification of replaced link clears flag for notification of new link
8. Fix router cache set on error

### LINKED ISSUE #1715: Ensure controller raft peers don't end up with duplicate connections
Ensure controller raft peers don't end up with duplicate connections

plorenz:
- Fix link management race conditions found by chaos testing. Fixes #1709
- Ensure controller raft peers don't end up with duplicate connections. Fixes #1715

### DIFF
--- a/zititest/zitilab/chaos/chaos.go
+++ b/zititest/zitilab/chaos/chaos.go
@@ -18,6 +18,7 @@ package chaos
 
 import (
 	"fmt"
+	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/fablab/kernel/model"
 	"math/rand"
 	"time"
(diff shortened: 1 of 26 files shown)


zititest/zitilab/chaos/chaos.go
```diff
@@ -92,4 +122,8 @@ func ValidateUp(run model.Run, spec string, concurrency int, timeout time.Durati
 			time.Sleep(time.Second)
 		}
 	})
+	if err == nil {
+		pfxlog.Logger().Infof("all %v components for spec '%s' are running", len(components), spec)
```
andrewpmartinez, line 126: I still dislike %v due to its lack of complaints when refactoring. In this scenario %d would work just fine.

Not demanding a change, just venting.

andrewpmartinez (approved):

Issue #1513, opened 2023-11-15
DNS service failure should not cause a router restart


mikegorman-nf:
Currently, if the DNS injection test fails, the router process will restart.  The system should log messages regarding the failure, but continue if possible.  This will allow connectivity to be preserved to access the system and attempt to remedy the issue.  A process restart is very unlikely to change the failed state.

dariuszSki:
@plorenz I found the bug related to this fix. Here are the details. The resolver test fails, and the router doe snot error out. But when I try to reach the service and initiate tcp accept log, it causes a sig fault.

`│ [   2.166]    INFO ziti/tunnel/dns.NewDnsServer: dns server running at 127.0.0.1:53                           │
│ [   2.166]    INFO ziti/tunnel/dns.(*resolver).AddHostname: adding ziti-tunnel.resolver.test = 19.65.28.94 to │
│  resolver                                                                                                     │
│ [   2.206]   ERROR ziti/router/xgress_edge_tunnel.(*tunneler).Start: {error=[system resolver test failed: fai │
│ led to resolve ziti-tunnel.resolver.test: lookup ziti-tunnel.resolver.test on 192.168.4.1:53: no such host    │
│                                                                                                               │
│ ziti-tunnel runs an internal DNS server which must be first in the host's                                     │
│ resolver configuration. On systems that use NetManager/dhclient, this can                                     │
│ be achieved by adding the following to /etc/dhcp/dhclient.conf:                                               │
│                                                                                                               │
│     prepend domain-name-servers 127.0.0.1:53;                                                                 │
│                                                                                                               │
│ ]} failed to start DNS resolver                                                                               │
│ [   2.749]    INFO ziti/tunnel/intercept.SetDnsInterceptIpRange: dns intercept IP range: 100.64.0.1 - 100.127 │
│ .255.254                                                                                                      │
│ [   2.807]
...
│ [  29.935]    INFO ziti/tunnel/intercept/tproxy.(*tProxy).acceptTCP: received connection: 10.20.96.100:98 --> │
│  10.20.96.0:45404                                                                                             │
│ panic: runtime error: invalid memory address or nil pointer dereference                                       │
│ [signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x30e7019]                                      │
│                                                                                                               │
│ goroutine 227 [running]:                                                                                      │
[...]

PR #1727, opened 2024-02-01
Optimize service add

scareything:
IP address selection for intercepted hostnames was taking a significant amount of time, especially for large numbers of services (1000's). This PR optimizes IP assignment:

* All local addresses were being looked up before getting a new IP. this was not necessary at all
* Assign the entire DNS IP range to the loopback device at startup instead of assigned each IP as it is allocated.
* re-use returned DNS IPs if available


scareything:
Inspired by Robert's changes at https://github.com/openziti/ziti/tree/optimize_service_add

router/xgress_edge_tunnel/tunneler.go
```diff
@@ -70,6 +70,16 @@ func (self *tunneler) Start(notifyClose <-chan struct{}) error {
 	log := pfxlog.Logger()
 	log.WithField("mode", self.listenOptions.mode).Info("creating interceptor")
 
+	resolver, err := dns.NewResolver(self.listenOptions.resolver)
+	if err != nil {
+		pfxlog.Logger().WithError(err).Error("failed to start DNS resolver")
```
plorenz, line 75: There's an issue related to this code: https://github.com/openziti/ziti/issues/1513
Do you think you could pop in a quick fix for that here.. just assign a default resolver if the selected resolver initialization fails?

plorenz (approved):

PR #1738, opened 2024-02-09
Get smoketest logs and store in actions

plorenz:


qrkourier (approved):
Should be fine as long as we're not logging secrets.

PR #1767, opened 2024-02-21
just log error when dns server self-test fails

### LINKED ISSUE #1513: DNS service failure should not cause a router restart

scareything:
fixes #1513


dovholuknf (approved):
lgtm

PR #1769, opened 2024-02-22
add freebsd function defs

qrkourier:


common/profiler/cpu_freebsd.go
plorenz, line 1: It looks like we could replace cpu_linux, cpu_darwin and cpu_freebsd with cpu_unix. Can you give that a try and see if it works?

tunnel/intercept/tproxy/tproxy_freebsd.go
```diff
@@ -0,0 +1,26 @@
...
+
+package tproxy
+
+import (
+	"github.com/openziti/ziti/tunnel/intercept"
+	"github.com/pkg/errors"
+)
+
+func New(config Config) (intercept.Interceptor, error) {
+	return nil, errors.New("tproxy not supported on darwin")
```
plorenz, line 25: Should be freebsd, not darwin

plorenz (approved):

PR #1782, opened 2024-02-28
don't add bind and dial permissions to services for admins

### LINKED ISSUE #1781: admin identities have bind and dial permissions to services
Admin identities have bind and dial permissions to any/all services. This causes tunnelers to intercept and host services when run with an admin identity, which can lead to potentially confusing issues like connection feedback.

Admin identifies would be more predicable and practical if they didn't automatically have bind and dial permissions to services. Instead these permissions can be granted with policies as desired by the end-user.

scareything:
fixes #1781


controller/model/edge_service_manager.go
```diff
@@ -143,10 +133,14 @@ func (self *EdgeServiceManager) ReadForIdentityInTx(tx *bbolt.Tx, id string, ide
```
plorenz, line 133: Can we merge ReadForNonAdminIdentityInTx into ReadForIdentityInTx, as the 'NonAdmin' name is no longer accurate and the first method is now pretty short?

controller/model/edge_service_manager.go
```diff
@@ -259,14 +250,8 @@ func (result *ServiceListResult) collect(tx *bbolt.Tx, ids []string, queryMetaDa
...
-		if !result.isAdmin && result.identityId != "" {
-			service, err = result.manager.ReadForNonAdminIdentityInTx(tx, key, result.identityId)
-		} else {
-			service, err = result.manager.readInTx(tx, key)
-			if service != nil && result.isAdmin {
-				service.Permissions = []string{db.PolicyTypeBindName, db.PolicyTypeDialName}
-			}
-		}
+		// service permissions for admin & non-admin identities will be set according to policies
+		service, err = result.manager.ReadForIdentityInTx(tx, key, result.identityId, result.configTypes)
```
scareything, line 254: ☝️ this will now happen when `result.identityId` == "", and I'm not really sure what that means or if it's ok.

plorenz: I'm not sure it can happen, but we could short-circuit it in queryServices. If the `identityId` is `""` we could return the service list result before we run the query.

plorenz:
Can you add a note to the changelog as well, please?

PR #1786, opened 2024-02-29
consolidated ha changes

andrewpmartinez:


common/event_cache.go
```diff
@@ -0,0 +1,243 @@
...
+		return nil
+	}
+
+	if cache.index != nil {
+		if *cache.index >= event.Index {
+			//already past this index
+			return nil
+		}
+
+		if *cache.index >= event.Index {
```
plorenz, line 73: This condition can't ever be hit since the same condition is checked above and causes a return

common/event_cache.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+}
+
+func (cache *LoggingEventCache) WhileLocked(callback func(uint64, bool)) {
+	cache.lock.Lock()
+	defer cache.lock.Unlock()
+
+	callback(cache.currentIndex())
+}
+
```
plorenz, line 106: LoggingEventCache and ForgetfulEventCache methods are mingled

common/event_cache.go
```diff
@@ -0,0 +1,243 @@
...
+	lock  sync.Mutex
+	index *uint64
+}
+
+func NewForgetfulEventCache() *ForgetfulEventCache {
+	return &ForgetfulEventCache{}
+}
+
+func (cache *ForgetfulEventCache) SetCurrentIndex(index uint64) {
+	cache.index = &index
```
plorenz, line 46: Should this lock?

controller/db/controller_store.go
```diff
@@ -0,0 +1,109 @@
+/*
```
plorenz, line 1: I'm a little hesitant to store controller information in the db, especially the 'isOnline' and 'lastJoined'. Won't the different controllers have different opinions on that, depending on who is connected to who and when?

controller/env/appenv.go
```diff
@@ -82,31 +82,37 @@ const (
 )
 
 type AppEnv struct {
```
plorenz, line 84: BoltStores field isn't used anywhere

controller/env/broker.go
```diff
@@ -45,6 +46,9 @@ const (
 	EnrollmentExtendRouterVerifyRequestType = int32(edge_ctrl_pb.ContentType_EnrollmentExtendRouterVerifyRequestType)
 )
```
plorenz, line 48: Not used anywhere

controller/env/broker.go
```diff
@@ -69,16 +73,26 @@ func NewBroker(ae *AppEnv, synchronizer RouterSyncStrategy) *Broker {
 	broker.ae.GetStores().ApiSession.GetEventsEmitter().AddListener(db.EventFullyAuthenticated, broker.apiSessionFullyAuthenticated)
 	broker.ae.GetStores().ApiSessionCertificate.AddEntityEventListenerF(broker.apiSessionCertificateCreated, boltz.EntityCreatedAsync)
 	broker.ae.GetStores().ApiSessionCertificate.AddEntityEventListenerF(broker.apiSessionCertificateDeleted, boltz.EntityDeletedAsync)
+	broker.ae.GetStores().Controller.AddEntityEventListenerF(broker.controllerAdded, boltz.EntityCreated)
+	broker.ae.GetStores().Controller.AddEntityEventListenerF(broker.controllerAdded, boltz.EntityUpdated)
 
 	ae.HostController.GetNetwork().AddRouterPresenceHandler(broker)
 	ae.HostController.GetEventDispatcher().AddClusterEventHandler(broker)
```
plorenz, line 81: Is this doing something?

controller/db/controller_store.go
plorenz: Ah, i got further and saw that this is gated to the leader, so I'm less concerned :+1:

controller/handler_edge_ctrl/hello.go
```diff
@@ -26,10 +26,10 @@ import (
 
 type helloHandler struct {
 	appEnv   *env.AppEnv
```
plorenz, line 28: msg added to params, but not used

controller/oidc_auth/parse.go
```diff
@@ -0,0 +1,233 @@
...
+
+type FieldTranslator interface {
+	Translate(string, ...string) (string, bool)
+}
+
+func MapToStruct(m map[string][]string, dst interface{}) error {
+	translator, _ := dst.(FieldTranslator)
+	return mapToStruct(0, m, dst, translator)
+}
+
```
plorenz, line 66: There's a mapstructure library that we already include that does this. I'm not sure if it covers this use case exactly, but might be something to check out.

controller/sync_strats/sync_instant.go
```diff
@@ -89,8 +102,107 @@ type InstantStrategy struct {
...
+
+	//ca create/delete/update
+	strategy.caHandler = &constraintToIndexedEvents[*db.Ca]{
+		indexProvider: strategy.indexProvider,
+		createHandler: strategy.CaCreate,
+		updateHandler: strategy.CaUpdate,
+		deleteHandler: strategy.CaDelete,
+	}
+	strategy.ae.GetStores().Ca.AddEntityConstraint(strategy.caHandler)
+
```
plorenz, line 195: comment ca -> revocation

router/config.go
```diff
@@ -764,6 +769,16 @@ func LoadConfig(path string) (*Config, error) {
 		}
 	}
```
plorenz, line 771: Can we figure this out from the controller instead of needing to set it manually?

plorenz: Or is this one of things where the controller needs to know what the router can do, so the router can't ask the controller? Or more of a temporary thing, where in the long run everything will assume the router is running with the data model?

router/internal/edgerouter/config.go
```diff
@@ -132,6 +136,33 @@ func (config *Config) LoadConfigFromMap(configMap map[interface{}]interface{}) e
...
+	}
+
+	if config.Db == "" {
+		config.Db = "./db.json.gzip"
+
+		if value, found := configMap[router.PathMapKey]; found {
+			configPath := value.(string)
+			configPath = strings.TrimSpace(configPath)
+			config.Db = configPath + ".json.gzip"
+		} else {
```
plorenz, line 150: linter err

router/posture/access.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"fmt"
+	"github.com/michaelquigley/pfxlog"
+	"github.com/openziti/ziti/common"
+	"github.com/openziti/ziti/common/eid"
+	"github.com/openziti/ziti/common/pb/edge_ctrl_pb"
+)
+
+func HasAccess(rdm *common.RouterDataModel, identityId string, serviceId string, cache *Cache) (*edge_ctrl_pb.DataState_ServicePolicy, error) {
```
plorenz, line 11: What's the point of the instance random id? Just log correlation?

tests/ats-edge.router.yml.json.gzip
plorenz, line 1: This seems like a weird file, (yml.json.gzip?), I did a quick scan, don't see it used anywhere. Can we drop it?

plorenz (approved):
Several smaller things, but nothing blocking other than linter fixes (and maybe getting rid of that .gzip file).

controller/env/broker.go
andrewpmartinez: It is meant to replace the pub key updates currently coming from the raft's peer list directly (which may not be a complete list). I have that logic separate from this because it was experimental for quite a bit and one of the last pieces.

router/config.go
andrewpmartinez: This is a temporary feature flag that is meant to be removed once the HA work is generally available. Without this flag, the routers report to controllers whether they support the new router data model, and the controller sends the appropriate messages.

With this flag, when `false`, the router will never signal to the controller that it supports the new router data model.

This was asked for in order to "limit the blast radius" of bugs that HA changes bring to existing networks. Without it, if people upgrade after this PR is merged, the new router data model code starts to run even if not being used. The end goal is that every network uses routers using the new router data model and legacy ApiSessions/Sessions and messaging on the ctrl channel is removed.

tests/ats-edge.router.yml.json.gzip
andrewpmartinez: Wasn't supposed to be added. Will remove.

It is the backup router data model that gets created and saved periodically.

router/posture/access.go
andrewpmartinez: I did it for testing purposes. I had multiple instances posture checks evaluating at the same time and it was hard from logs to always tell which logging messages were related. I added that id as a quick way to group them in scenarios where log messages were being interleaved.

It isn't strictly necessary, but it was helpful to me at one time.

router/internal/edgerouter/config.go
andrewpmartinez: Fixed

common/event_cache.go
andrewpmartinez: Fixed.

PR #1788, opened 2024-02-29
Terminator chaos

plorenz:


controller/db/api_session_store.go
```diff
@@ -149,44 +151,43 @@ func (store *apiSessionStoreImpl) onEventualDelete(db boltz.Db, name string, api
...
-
-		if err != nil {
-			pfxlog.Logger().WithError(err).WithFields(map[string]interface{}{
-				"eventName":    name,
-				"apiSessionId": string(apiSessionId),
-				"sessionId":    id,
-			}).Error("error deleting for session associated to an api session during onEventualDelete")
-		}
+	if store.stores.rateLimiter.GetQueueFillPct() > 0.5 {
+		time.Sleep(time.Second)
```
andrewpmartinez, line 155: This seems odd. Intentional?

andrewpmartinez (approved):

controller/db/api_session_store.go
plorenz: Yes, I did encounter a case where the background processing of API sessions seemed to cause an issue. 
On the other, unbeknownst to me at the time, I had exhausted my I/O burst capacity on the disk. 
So on the one hand this shouldn't hurt anything. It just slows down background processing if there's a lot of foreground activity happening. On the other, it might not be necessary. I can take it out or leave it.

PR #1800, opened 2024-03-06
add controller and router packages and container images

qrkourier:


ziti/cmd/create/config_templates/controller.yml
```diff
@@ -7,7 +7,7 @@ v: 3
 #  memory:
 #    path: ctrl.memprof
 
-db:                     "{{ .ZitiHome }}/db/ctrl.db"
+db:                     "{{ .ZitiHome }}/{{ .Controller.Database.DatabaseFile }}"
```
qrkourier, line 10: use same default, but allow overriding with `ZITI_CTRL_DATABASE_FILE`.

qrkourier:
Package builds for testing (arm64 also available):

- deb: https://netfoundry.jfrog.io/artifactory/zitipax-fork-deb-stable/pool/openziti-controller/amd64/openziti-controller_0.32.0_amd64.deb
- rpm: https://netfoundry.jfrog.io/artifactory/zitipax-fork-rpm-stable/redhat/x86_64/openziti-controller-0.32.0.x86_64.rpm

.github/workflows/fablab-db-creation.yml
```diff
@@ -20,6 +20,7 @@ jobs:
   build:
     name: Build and Run
     runs-on: ubuntu-latest
+    if: github.repository_owner == 'openziti'
```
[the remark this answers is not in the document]
dovholuknf: you really should try to do this sorta cleanup in separate pr's imo...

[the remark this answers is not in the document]
qrkourier: It's a necessary change for me to test this PR.

ziti/cmd/create/config_templates/controller.yml
dovholuknf: remove the `{{ .ZitiHome }}/` bit and add/expect it in the `.Controller.Database.DatabaseFile` field, pls

dist/dist-packages/linux/openziti-controller/env
```diff
@@ -0,0 +1,62 @@
...
+
+# disable JSON logging during bootstrapping
+PFXLOG_NO_JSON=true
+
+#
+# for "ziti pki" and "ziti create config controller" commands in bootstrap.bash
+#
+
+# the advertised address of the controller is a domain name that can be resolved by all devices (default: hostname -f)
+ZITI_CONTROLLER_ADVERTISED_ADDRESS=
```
dovholuknf, line 16: use ZITI_CTRL_* -- they're meant for the same sort of precedence purpose and work that way

dist/dist-packages/linux/openziti-controller/ziti-controller.service
```diff
@@ -0,0 +1,40 @@
...
+
+[Service]
+# "ziti controller run" is the main process managed by this service and replaces entrypoint.bash
+Type=simple
+
+# manage the user and permissions for the service automatically
+DynamicUser=yes
+
+# allow binding low ports, e.g., 443/tcp
+AmbientCapabilities=CAP_NET_BIND_SERVICE
```
scareything, line 13: should this be optional? I'd think most users will forward from their router or load balancer, and don't need to actually listen on `:443`

qrkourier: For convenience, I'm leaning toward a set of defaults that provides the necessary privileges for common running states such as binding to privileged ports or running a tproxy router with `NET_ADMIN` capability (pertains to the router branch).

I could go either way, however. A more conservative default would be to leave this ambient grant as a comment.

qrkourier: I think so, yes. I'd prioritized convenience, but on second thought, it's better to prioritize the default security posture and comment this grant out as a hint for the operator who does need to bind a privileged port.

That reminds me, too, that including a comment here about pass-through TLS for load balancing would be helpful.

ziti/cmd/create/config_templates/controller.yml
qrkourier: In case it's an unintended consequence, this will change the way the environment is computed to include the full path:

Before:

```
ZITI_HOME=/path/to/cwd
ZITI_CTRL_DATABASE_FILE=db/ctrl.db
```

After:

```
ZITI_HOME=/path/to/cwd
ZITI_CTRL_DATABASE_FILE=/path/to/cwd/db/ctrl.db
```

It will also entail parsing any value set in `ZITI_CTRL_DATABASE_FILE` to determine whether or not it's a qualified path, and if not, to assume the qualified path is relative to `ZITI_HOME`.

qrkourier:
New CI artifacts tested in a repo fork:
https://github.com/qrkourier/ziti/actions/runs/8303508508

ziti/cmd/create/create_config_test.go
```diff
@@ -33,43 +33,44 @@ const (
 
 func getZitiEnvironmentVariables() []string {
 	return []string{
-		"ZITI_HOME",
```
scareything, line 36: This kind of change is hard to verify visually. Personally, I would do re-ordering like this in a different PR.

ziti/cmd/create/create_config_controller.go
```diff
@@ -150,7 +149,6 @@ func NewCmdCreateConfigController() *CreateControllerConfigCmd {
 
 func (options *CreateConfigControllerOptions) addFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&options.CtrlPort, optionCtrlPort, constants.DefaultCtrlAdvertisedPort, "port used for the router to controller communication")
-	cmd.Flags().StringVar(&options.DatabaseFile, optionDatabaseFile, "ctrl.db", "location of the database file")
```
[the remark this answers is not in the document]
qrkourier: Yes. This removes an incomplete implementation that was never in effect. The command-line option had been added, but the template never respected the argument because it was hard-coded with the default value.

ziti/constants/constants.go
```diff
@@ -58,62 +60,64 @@ const (
...
+	PkiAltServerCertVarDescription                   = "Path to controller's root identity alternative server certificate. Requires ZITI_PKI_ALT_SERVER_KEY"
 	PkiAltServerKeyVarName                           = "ZITI_PKI_ALT_SERVER_KEY"
-	PkiAltServerKeyVarDescription                    = "Key to use with the alternative server cert. Must be specified with ZITI_PKI_ALT_SERVER_CERT"
+	PkiAltServerKeyVarDescription                    = "Path to controller's root identity alternative private key. Requires ZITI_PKI_ALT_SERVER_CERT"
 	ZitiEdgeRouterNameVarName                        = "ZITI_ROUTER_NAME"
-	ZitiEdgeRouterNameVarDescription                 = "The Edge Router Name"
+	ZitiEdgeRouterNameVarDescription                 = "A slug by which to name the router's identity-related files"
 	ZitiEdgeRouterPortVarName                        = "ZITI_ROUTER_PORT"
-	ZitiEdgeRouterPortVarDescription                 = "Port of the Ziti Edge Router"
+	ZitiEdgeRouterPortVarDescription                 = "Router's exposed TCP port"
```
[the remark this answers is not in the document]
qrkourier: Despite the apparent distinction implied by the var names `ZitiEdgeRouterPortVarName` and `ZitiEdgeRouterListenerBindPortVarName`, the current router config generation logic does not distinguish between listener/bind/exposed and advertised/published/forwarded ports.

`ZitiEdgeRouterPortVarName` (`ZITI_ROUTER_PORT`) overrides the default value 3022, where a router listens for edge connections as well as the port that is advertised to edge clients authorized by ERP.

`ZitiEdgeRouterListenerBindPortVarName` (`ZITI_ROUTER_LISTENER_BIND_PORT`) overrides the default value 10080, where a router listens for incoming transport links as well as the port that is advertised to all other routers.

For the purpose of this minimal, sane bootstrapper, we'll have a single option `ZITI_ROUTER_PORT` that overrides the default 3022 for all four meanings:

- listen for edge conns
- advertise edge conns
- listen for links
- advertise links

ziti/constants/constants.go
```diff
@@ -58,62 +60,64 @@ const (
 // Env Var Constants
 const (
 	ZitiHomeVarName        = "ZITI_HOME"
-	ZitiHomeVarDescription = "Root home directory for Ziti related files"
+	ZitiHomeVarDescription = "Root home directory for Ziti-related files"
 
 	PkiCtrlCertVarName                               = "ZITI_PKI_CTRL_CERT"
-	PkiCtrlCertVarDescription                        = "Path to Identity Cert for Ziti Controller"
+	PkiCtrlCertVarDescription                        = "Path to controller's default identity client cert"
```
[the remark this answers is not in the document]
qrkourier: > the controller's

Sounds a little better. I rectified the others.

ziti/constants/constants.go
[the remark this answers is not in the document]
qrkourier: I'm planning to express the same convention for single address and port for both controller and router by changing the precedence as described in this separate issue:

ziti/constants/constants.go
```diff
@@ -58,62 +60,64 @@ const (
...
+	CtrlBindAddressVarDescription                    = "The address on which the controller will listen on for router control plane connections"
 	CtrlAdvertisedAddressVarName                     = "ZITI_CTRL_ADVERTISED_ADDRESS"
-	CtrlAdvertisedAddressVarDescription              = "The address routers will use to connect to the Ziti Controller"
+	CtrlAdvertisedAddressVarDescription              = "The address routers will use to connect to the controller"
 	CtrlAdvertisedPortVarName                        = "ZITI_CTRL_ADVERTISED_PORT"
-	CtrlAdvertisedPortVarDescription                 = "The port routers will use to connect to the Ziti Controller"
+	CtrlAdvertisedPortVarDescription                 = "The port routers will use to connect to the controller"
 	CtrlEdgeBindAddressVarName                       = "ZITI_CTRL_EDGE_BIND_ADDRESS"
-	CtrlEdgeBindAddressVarDescription                = "The address on which the edge controller will listen on for API connections"
+	CtrlEdgeBindAddressVarDescription                = "The address on which the controller will listen on for API connections"
```
dovholuknf, line 80: the word edge is important here. put it back pls. since you're touching this file, would you uncapitalize Ziti-related too.

ziti/constants/constants.go
```diff
@@ -58,62 +60,64 @@ const (
...
+	CtrlPkiEdgeCAVarDescription                      = "Path to the controller's web identity root CA cert"
 	PkiAltServerCertVarName                          = "ZITI_PKI_ALT_SERVER_CERT"
-	PkiAltServerCertVarDescription                   = "Alternative server certificate to use. Must be specified with ZITI_PKI_ALT_SERVER_KEY"
+	PkiAltServerCertVarDescription                   = "Path to the controller's root identity alternative server certificate. Requires ZITI_PKI_ALT_SERVER_KEY"
 	PkiAltServerKeyVarName                           = "ZITI_PKI_ALT_SERVER_KEY"
-	PkiAltServerKeyVarDescription                    = "Key to use with the alternative server cert. Must be specified with ZITI_PKI_ALT_SERVER_CERT"
+	PkiAltServerKeyVarDescription                    = "Path to the controller's root identity alternative private key. Requires ZITI_PKI_ALT_SERVER_CERT"
 	ZitiEdgeRouterNameVarName                        = "ZITI_ROUTER_NAME"
-	ZitiEdgeRouterNameVarDescription                 = "The Edge Router Name"
+	ZitiEdgeRouterNameVarDescription                 = "A filename prefix for the router's key and certs"
```
dovholuknf, line 110: the old value was correct. it's the name "in" ziti. please put it back. it's just happenstance that the name is used to namespace files like this

ziti/cmd/pki/pki_create_client.go
```diff
@@ -79,6 +79,7 @@ func (o *PKICreateClientOptions) addPKICreateClientFlags(cmd *cobra.Command) {
 	cmd.Flags().IntVarP(&o.Flags.CAPrivateKeySize, "private-key-size", "", 4096, "Size of the RSA private key, ignored if -curve is set")
 	cmd.Flags().StringVarP(&o.Flags.EcCurve, "curve", "", "", "If set an EC private key is generated and -private-key-size is ignored, options: P224, P256, P384, P521")
 	cmd.Flags().StringVar(&o.Flags.SpiffeID, "spiffe-id", "", "Optionally provide the path portion of a SPIFFE id. The trust domain will be taken from the signing certificate.")
+	cmd.Flags().BoolVar(&o.Flags.AllowOverwrite, "allow-overwrite", false, "Allow overwrite existing certs")
```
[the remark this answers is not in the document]
dovholuknf: this change we need to discuss. this will block my review

ziti/cmd/create/create_config_controller.go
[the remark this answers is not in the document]
dovholuknf: we can never be sure someone didn't/doesn't use it for whatever reason. it'll still need a "we removed this and replaced it with an env var" release note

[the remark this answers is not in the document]
qrkourier: @dovholuknf The implementation that I removed here was incomplete, and so there's no possibility that it was ever used. The option didn't do anything at all.

ziti/constants/constants.go
qrkourier: With this change, I'm working toward filling a particular gap between how this var is documented in the output of `ziti create config environment` and the fact that it's ignored by `ziti create config router`, and replaced by a required flag `--routerName` ([link to dev doc example of this confusing experience](https://github.com/openziti/ziti/blob/main/doc/003-local-deploy.md#initialize-an-edge-router)).

The result of any changes here should be a description that answers the question "How is this variable used by Ziti's built-in config generator?" because only users of the built-in config generator will see this description, not quickstart users.

I recognize the quickstarts ([link to line](https://github.com/openziti/ziti/blob/linux-package-router/ziti/cmd/edge/quickstart.go#L139)) do co-opt the same variable to set the router's entity name in the Ziti mgmt API, but that use case should be documented as part of the quickstarts since this config generator has other applications that do not co-opt the variable.

ziti/constants/constants.go
qrkourier: I could not find any occurrences of strings like "Ziti related" or "Ziti-related" in the option var descriptions in this file, constants.go, that I could uncapitalize. The closest thing I found were the Go variable names, but I doubt that's what you meant.

qrkourier:
Converted ready PR to draft PR while I add the interactive controller install prompts and normalize the answer file pattern across both controller and router Linux packages.

qrkourier:
This PR was for the controller Linux PR only and now has aggregated changes for the following due to their interdependence and shared timing:

- controller
  - add linux package w/ optional bootstrapping
  - add optional bootstrapping to the container image
- router
  - linux package w/ optional bootstrapping
  - add optional bootstrapping to the container image

dist/docker-images/ziti-controller/compose.yml
```diff
@@ -0,0 +1,50 @@
...
+    volumes:
+      - ziti-controller:/mnt
+    working_dir: /mnt
+    # these declared vars pass through to container and should be assigned in an .env file or exported from parent env
+    # to ensure consistency throughout the compose project
+    environment:
+      # *** these are the important vars to set ***
+      ZITI_CTRL_ADVERTISED_ADDRESS:                                   # domain name of this controller (required)
+      ZITI_CTRL_ADVERTISED_PORT: ${ZITI_CTRL_ADVERTISED_PORT:-1280}   # exposed port of this controller
+      ZITI_PWD: ${ZITI_PWD:-}                                         # password for the default admin user
```
dovholuknf, line 27: i think `:-` should be removed here and make the user set it.

qrkourier: It's optional in this revision because it's only required for the bootstrapping (first) run. I'll split the inputs between two containers: bootstrap and run. That way, it will be clearer which variables apply to which.

qrkourier:
Demos:

- Linux: https://youtu.be/SU_piIyc4wg
- Docker: https://youtu.be/bQ0U4V1V-I0

PR #1844, opened 2024-03-19
add tags to old VPCs

qrkourier:


qrkourier:
@plorenz My goal here is to include the AWS entity tags when I send a report like "I'll delete these VPCs today at 2PM."

plorenz:
> @plorenz My goal here is to include the AWS entity tags when I send a report like "I'll delete these VPCs today at 2PM."

Thanks, @qrkourier , makes sense

PR #1864, opened 2024-03-25
Fix Controller List & Pub Key Sync

andrewpmartinez:
- fixes permissions on the controller list endpoint
- updates API information returned for client and management
- updates pub key sync to use controller list from the leader


controller/db/controller_store.go
```diff
@@ -91,19 +101,58 @@ func (store *controllerStoreImpl) NewEntity() *Controller {
...
 	entity.LoadBaseValues(bucket)
 	entity.Name = bucket.GetStringOrError(FieldName)
-	entity.Address = bucket.GetStringWithDefault(FieldControllerAddress, "")
+	entity.CtrlAddress = bucket.GetStringWithDefault(FieldControllerCtrlAddress, "")
 	entity.CertPem = bucket.GetStringOrError(FieldControllerCertPem)
 	entity.Fingerprint = bucket.GetStringOrError(FieldControllerFingerprint)
 	entity.IsOnline = bucket.GetBoolWithDefault(FieldControllerIsOnline, false)
 	entity.LastJoinedAt = bucket.GetTime(FieldControllerLastJoinedAt)
+	entity.ApiAddresses = map[string][]ApiAddress{}
+
```
plorenz, line 110: I'm not sure there's a better way to do this for now. TypedBucket supports maps and string lists. Could do it with either a map or two string-lists, but neither is ideal.

plorenz: oh, it's not a list, it's keyed by apiKey. So maybe a map is the right way. Not a big deal, though.

plorenz (approved):

PR #1894, opened 2024-04-04
fix: close cpu profile

testwill:


plorenz (changes requested):
Hi @testwill thanks for the PR. Unfortunately, it doesn't look quite right. The file needs to stay open while the profiler is running. It should be closed in the Run method, after the StopCPUProfile method is called.

plorenz (approved):
looks good to me. Thank you!

plorenz:
Hi @testwill, I noticed that your commit isn't signed. Would you be willing to set up commit signing and replace the commit with a signed commit? 
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #1899, opened 2024-04-04
Optimize network run loop. Fixes #1897

### LINKED ISSUE #1897: Optimize network run loop
When routers change (connect/disconnect), we re-check the mesh. We are passing in the routers to the channel and they can build up, which is unnecessary. We only need to signal once for all the routers that change since the last time we checked, that the mesh needs to be checked. 

We also pass link changes through the run method, which is unnecessary since we pass it to a new goroutine. This inefficiency should be fixed.

plorenz:
* When routers change (connect/disconnect), we re-check the mesh. We are passing in the routers to the channel and they can build up, which is unnecessary. We only need to signal once for all the routers that change since the last time we checked, that the mesh needs to be checked.
* We pass link changes through the run method, which is unnecessary since we pass it to a new goroutine. This inefficiency should be fixed.
* We're evaluating all faulted links for rerouting, even if they were still pending. Only reroute connected links.
* We're using time.After in each for loop, which can accumulate timers. Use a single Ticker instead.

### DIFF
--- a/controller/api_impl/link_router.go
+++ b/controller/api_impl/link_router.go
@@ -122,7 +122,7 @@ func (r *LinkRouter) Patch(n *network.Network, rc api.RequestContext, params lin
 		if fields.IsUpdated("down") {
 			l.SetDown(params.Link.Down)
 		}
-		n.LinkChanged(l)
+		n.RerouteLink(l)
 		return nil
 	})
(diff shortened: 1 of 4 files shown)


dovholuknf (approved):
lgtm

PR #1901, opened 2024-04-04
Improve link mgmt backwards compatibility. Fixes #1900

### LINKED ISSUE #1900: Improve link management backwards compatibility
* We're sending peer change messages to older routers that don't understand them.
* We should probably use the older controller link management code for routers without the link management stability fixes

plorenz:
* We're sending peer change messages to older routers that don't understand them.
* We should probably use the older controller link management code for routers without the link management stability fixes
* Includes a few minor cleanups

### DIFF
--- a/controller/network/router.go
+++ b/controller/network/router.go
@@ -31,6 +31,7 @@ import (
 	"github.com/openziti/ziti/controller/fields"
 	"github.com/openziti/ziti/controller/xt"
 	"google.golang.org/protobuf/proto"
+	"maps"
 	"reflect"
 	"strings"
 	"sync"
@@ -104,6 +105,14 @@ func (entity *Router) SetMetadata(metadata *ctrl_pb.RouterMetadata) {
 
 func (entity *Router) HasCapability(capability ctrl_pb.RouterCapability) bool {
 	return entity.Metadata != nil && genext.Contains(entity.Metadata.Capabilities, capability)
+}
+
+func (entity *Router) SupportsRouterLinkMgmt() bool {
+	if entity.VersionInfo == nil {
+		return true
+	}
+	supportsLinkMgmt, err := entity.VersionInfo.HasMinimumVersion("0.32.1")
+	return err != nil || supportsLinkMgmt
 }
 
 func NewRouter(id, name, fingerprint string, cost uint16, noTraversal bool) *Router {
@@ -593,9 +602,7 @@ func (self *RouterLinks) Add(link *Link, otherRouterId string) {
 
 	byRouter := self.GetLinksByRouter()
 	newLinksByRouter := map[string][]*Link{}
-	for k, v := range byRouter {
-		newLinksByRouter[k] = v
-	}
+	maps.Copy(newLinksByRouter, byRouter)
 	forRouterList := newLinksByRouter[otherRouterId]
 	newForRouterList := append([]*Link{link}, forRouterList...)
 	newLinksByRouter[otherRouterId] = newForRouterList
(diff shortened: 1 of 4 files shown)


controller/network/router.go
```diff
@@ -106,6 +107,14 @@ func (entity *Router) HasCapability(capability ctrl_pb.RouterCapability) bool {
 	return entity.Metadata != nil && genext.Contains(entity.Metadata.Capabilities, capability)
 }
 
+func (entity *Router) SupportsRouterLinkMgmt() bool {
+	if entity.VersionInfo == nil {
+		return true
+	}
+	supportsLinkMgmt, err := entity.VersionInfo.HasMinimumVersion("0.32.1")
```
dovholuknf, line 114: 0.32.1 seems like it should be a constant? maybe documentation/link to the readers of the future here? feels like it could use some meta to me

controller/network/router_messaging.go
```diff
@@ -188,7 +188,12 @@ func (self *RouterMessaging) syncStates() {
 		}
 
 		if v.sendInProgress.Load() {
+			continue
+		}
 
+		if !notifyRouter.SupportsRouterLinkMgmt() {
+			// if the router doesn't support router based link mgmt, don't send these messages
+			delete(self.routerUpdates, k)
```
dovholuknf, line 196: seems like a log here would be useful? not every time but at least once in a while that this is happening? are there repercussions that somehow noticeable in some other way?

controller/network/router_messaging.go
```diff
@@ -277,6 +282,12 @@ func (self *RouterMessaging) sendTerminatorValidationRequest(routerId string, up
 		}
 	}
 
+	supportsVerifyV2, err := notifyRouter.VersionInfo.HasMinimumVersion("0.31.1")
```
dovholuknf, line 285: seems like it should be a function with some meta around it like a comment or links to issues or something -- and a constant?

controller/network/router_messaging.go
```diff
@@ -295,9 +306,16 @@ func (self *RouterMessaging) sendTerminatorValidationRequest(routerId string, up
 		return
 	}
 
-	req := &ctrl_pb.ValidateTerminatorsV2Request{
-		Terminators: terminators,
-		FixInvalid:  true,
+	var req protobufs.TypedMessage
+	if !supportsVerifyV2 {
```
dovholuknf, line 310: nit: is the negation here needed? seems like it could be removed

controller/network/router_messaging.go
```diff
@@ -309,6 +327,11 @@ func (self *RouterMessaging) sendTerminatorValidationRequest(routerId string, up
 		if self.managers.Dispatcher.IsLeaderOrLeaderless() {
 			if err := protobufs.MarshalTyped(req).WithTimeout(time.Second * 1).SendAndWaitForWire(ch); err != nil {
 				pfxlog.Logger().WithError(err).WithField("routerId", notifyRouter.Id).Error("failed to send validate terminators request to router")
+			} else if !supportsVerifyV2 {
+				// V1 doesn't send responses, it will just send deletes if the terminator is invalid.
+				// we're going to mark these ok. If they're not, we should get a delete message. Older
+				// routers can still fail to delete, if the delete gets lost for some reason.
+				self.generateMockResponseForV1(notifyRouter, updates)
```
dovholuknf, line 334: another 'does this deserve a warning' log? or maybe a metric?

dovholuknf (approved):
lgtm other than nits :)

plorenz:
> lgtm other than nits :)

I think we can drop support for pre-1.0 versions in 6-12 months, so I think it's the right level of abstracted. I don't want to add any extra logs for noting version differences, especially for recurring events, since they could be quite high volume.

PR #1924, opened 2024-04-09
Add release validation workflow. Fixes #1923

### LINKED ISSUE #1923: Add release validation test suite
Add workflow to run chaos tests before release. Should be triggered manually.

plorenz:


controller/config.go
```diff
@@ -52,13 +52,16 @@ const (
...
+	DefaultTlsHandshakeRateLimiterEnabled = false
+
 	// TlsHandshakeRateLimiterMinSizeValue is the minimum size that can be configured for the tls handshake rate limiter
 	// window range
 	TlsHandshakeRateLimiterMinSizeValue = 5
 
 	// TlsHandshakeRateLimiterMaxSizeValue is the maximum size that can be configured for the tls handshake rate limiter
 	// window range
-	TlsHandshakeRateLimiterMaxSizeValue = 5000
+	TlsHandshakeRateLimiterMaxSizeValue = 10000
```
dovholuknf, line 64: can this be even larger and set down if people need/want to? i almost feel like it should default to max int/long/whatever

dovholuknf: nevermind - i guess you have to turn it on to use it anyway

dovholuknf (approved):

CHANGELOG.md
```diff
@@ -1,1346 +1,100 @@
...
 
-# Release 0.34.0
+What does marking OpenZiti as 1.0 mean?
 
-## What's New
+### Backwards Compatibility
+We've guaranteed API stability for SDK clients for years and worked hard to ensure that routers 
+and controllers would be backwards and forward compatible. However, we have had a variety of 
+management API changes and CLI changes. For post 1.0 releases we expect to make additions to the 
+APIs and CLI, but won't remove anything until it's been first marked as deprecated and then only
```
andrewpmartinez, line 11: Do we need a deprecation policy spellout and linked to?

Here is a quick one I came up with:


```
# OpenZiti Feature Deprecation Policy

## Overview
This policy outlines the approach for deprecating features within OpenZiti. Our goal is to minimize disruption for our users while ensuring that OpenZiti continues to evolve and improve.

## Policy Statement
- We provide a minimum of one version notice before a feature is deprecated.
- Notification of deprecation will be included in the release notes of a version preceding deprecation.
- Deprecations that break backwards compatability will be denoted by a major version change.

## Process
1. **Announcement**: Deprecation will be announced in:
   - Release notes
2. **Deprecation Period**: The feature will remain functional for at least the current major version cycle of the announcement.
3. **Support**: During the deprecation period, support will be provided as follows:
   - Documentation on alternatives or replacements.
   - Technical support for migration via Discourse
4. **Removal**: After the deprecation period, the feature will be removed in the next major release.

```

controller/config.go
plorenz: Yeah, I switched it to off by default, just to be safe.

andrewpmartinez (approved):

CHANGELOG.md
plorenz: That sounds reasonable. Where would we put that doc? Top level of the project? In the project readme?

andrewpmartinez: I would say so. Other than that, "Deprecation.md" is the only other place I would suggest.

PR #1981, opened 2024-04-25
Update deps and changelog

plorenz:


andrewpmartinez (approved):
Codespell but otherwise great.

PR #1988, opened 2024-04-26
correct the var name in test workflow

qrkourier:


.github/workflows/test-quickstart.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     name: Test all-in-one Quickstart
     runs-on: ubuntu-latest
     env:
-      ZITI_QUICK_TAG: local  # override default :latest since we're not pulling from registry
+      ZITI_CLI_TAG: local  # override default :latest since we're not pulling from registry
```
dovholuknf, line 46: I assume you searched the entire repo for this reference along with maybe seraching the ziti-doc repo? :) assuming yes - lgtm

dovholuknf (approved):

.github/workflows/test-quickstart.yml
qrkourier: I had failed to do that. I've broadened my search and finished separating the references to the two different quickstart images. It was a mistake to co-opt the `ZITI_QUICK_IMAGE` var for the all-in-one because it refers to a distinct container image: `openziti/quickstart`. Now, the all-in-one uses the `ZITI_CLI_IMAGE` var because it's the same image as the CLI, `openziti/ziti-cli`.

PR #2029, opened 2024-05-08
set default container log format from json to console

qrkourier:


dist/docker-images/ziti-cli/Dockerfile
```diff
@@ -59,5 +59,6 @@ RUN chmod 0755 /usr/local/bin/ziti
 RUN /usr/local/bin/ziti completion bash > /etc/bash_completion.d/ziti_cli
 
 USER ziggy
+ENV PFXLOG_NO_JSON=true
```
dovholuknf, line 62: we really want to recommend to people to log text and not ship json??? that doesn't feel right at all to me.

qrkourier: Your guess is as good as mine. Personally, when I'm knocking about in ziti I usually read my logs, not ingest them in an aggregator. Anyone that's exporting and aggregating structured log data can easily set this to false. My aim is to reduce the visual clutter in the documentation and examples by setting a default that works for, in my estimation, the majority of users.

dovholuknf: I would expect a production deployment to ship logs to an aggregator in some json format wehre it's parsed and retained, just like what nfconsole does. we have an existing best-practice for this established, so i'd keep it as json. 

now -- letting people choose to use a format that us humans can read easily through some kind of flag/switch would be friendly and appreciated, but the default should stay json.

qrkourier: The human-friendly examples frequently set `PFXLOG_NO_JSON=true` to override the default in favor of a more human-friendly format. You suggest replacing those with a new flag that does the same thing?

dovholuknf: i suggest leaving the default to log json in a production deployment. it's the default that ships with ziti for a reason. if we want the default not json, it should change in ziti not in the container.

PR #2141, opened 2024-06-13
Move release process to be based off of tagging instead of pushes to main

plorenz:


.github/workflows/release.yml
```diff
@@ -0,0 +1,257 @@
+name: release
+
+on:
+  push:
```
ekoby, line 4: not release trigger?

plorenz: Tag creates the release, marked as pre-release. Could do it the other way, but it's even more manual

RELEASING.md
```diff
@@ -59,12 +60,24 @@ Perform these steps in PR branches based on release-next (trunk).
...
+1. Push a tag like v*, typically on default branch HEAD to trigger the pre-release workflow named `release.yml`.
+
+## Stable and Latest Release
+
+Pre-releases are releases, but they're not promoted as "latest" in GitHub or automatically shipped downstream. Marking a
+release as not a prerelease makes it a stable release. There can be one stable release that's also marked "latest"
+(`isLatest: true`).
+
+1. After an arbitrary burn-in period, unmark "prerelease" in GitHub Releases (`isPrerelease: false`) to stop
+   discouraging users from consuming the now-stable release. This will automatically promote and advertise the downstreams.
```
ekoby, line 78: is this supposed to be `non-stable`?

`stop discouraging` reads like double negative, I had to think what it means

ekoby (approved):

.github/workflows/release.yml
ekoby: don't we want release to be a deliberate manual process? 
either way is fine

plorenz: It is a deliberate manual process, it's just one done via tag, push tag, rather than through the GH UI.

qrkourier (approved):

PR #2203, opened 2024-07-08
use the controller image so the console is available

qrkourier:


.github/workflows/test-quickstart.yml
```diff
@@ -72,7 +72,9 @@ jobs:
         working-directory: ./quickstart/docker/all-in-one
         env:
           ZITI_GO_VERSION: ${{ steps.setup-go.outputs.go-version }}
-        run: docker compose --profile test run --rm quickstart-test
+        run: |
+          sleep 9  # wait a moment for the :local image to build
```
dovholuknf, line 76: A while loop would be better

dovholuknf (approved):

PR #2206, opened 2024-07-09
Merge fabric and controller model code. Fixes #2205

### LINKED ISSUE #2205: Merge fabric and edge model code
Since combing the fabric and edge code bases, we have been harmonizing them. The db layer is now coherent. The next step up is the model layer. If we can merge that, it will remove duplicate and divergent code and cleanup some technical debt.

plorenz:
* Fabric model code has been merged into the model package, which has the edge model managers. 
    * Note that model.Service had to be renamed to EdgeService, to avoid conflicting with the fabric level Service
* Network and some related code is still in network. There's an argument that it should be in model, but it doesn't own any entities specifically, so it also seems like it could be a layer above
* AppEnv is now initialized by the root controller
* The edge controller type still exists. A future pass will merge it into the root controller
* Config code has been consolidated in the config package. The network, raft and edge config types now live there as well
* In general Network is now less central and is more focused on routing and mesh management. AppEnv should be used in more places where we only need access to managers. 

Note: The controller will likely still be able to run without the edge enable. Some of the ATs run in this mode, and they still work. I have not tested it outside of that environment.


controller/env/appenv.go
```diff
@@ -228,7 +227,7 @@ func (ae *AppEnv) ValidateServiceAccessToken(token string, apiSessionId *string)
 		return nil, err
 	}
```
plorenz, line 229: This is a bug fix. `tokenRevocation` is checked above and can only be nil here. I assume that this was supposed to be just `revocation`

controller/model/api_session_certificate_manager.go
```diff
@@ -89,8 +89,6 @@ func (self *ApiSessionCertificateManager) CreateFromCSR(apiSessionId string, lif
 		PEM:          string(certPem),
 	}
```
plorenz, line 91: I don't know why this was here, but it has no effect, so removing it should be ok

controller/model/command.go
```diff
@@ -78,29 +83,14 @@ func (self *CommandManager) decodeDeleteEntityCommand(_ int32, data []byte) (com
...
 	}
 
-	decoder, err := ioc.Get[deleteDecoderF](self.Registry, msg.EntityType+DeleteDecoder)
+	decoder, err := ioc.Get[deleteDecoderF](self.registry, msg.EntityType+DeleteDecoder)
 	if err != nil {
 		return nil, err
 	}
 
 	return decoder(msg)
 }
```
plorenz, line 93: decodeSyncSnapshotCommand moved to network

controller/model/link_manager.go
```diff
@@ -298,106 +305,8 @@ func (linkController *linkController) hasDirectedLink(a, b *Router, linkProtocol
 	return false
 }
 
-func (linkController *linkController) linksInMode(mode LinkMode) []*Link {
-	return linkController.linkTable.allInMode(mode)
-}
-
```
plorenz, line 304: Moved to Network

controller/model/link_manager.go
```diff
@@ -298,106 +305,8 @@ func (linkController *linkController) hasDirectedLink(a, b *Router, linkProtocol
...
-				DestRouterId:  dest,
-				Dialed:        link.Src.Id == router.Id,
-			}
-			result.LinkDetails = append(result.LinkDetails, detail)
-		}
-	}
-
-	cb(result)
-}
-
```
plorenz, line 392: Moved to Network

controller/model/link_manager_test.go
```diff
@@ -72,25 +72,15 @@ func TestLifecycle(t *testing.T) {
...
 	l0.SetState(Connected)
-	linkController.add(l0)
+	linkController.Add(l0)
 
-	neighbors := linkController.connectedNeighborsOfRouter(r0)
+	neighbors := linkController.ConnectedNeighborsOfRouter(r0)
 	assert.Equal(t, 1, len(neighbors))
 	assert.Equal(t, r1, neighbors[0])
 }
-
```
plorenz, line 87: moved to testing.go

controller/model/router_manager.go
```diff
@@ -53,68 +49,6 @@ const (
 	RouterDequiesceFlag uint32 = 2
 )
```
plorenz, line 51: split to router_model, to match other types

controller/model/router_manager.go
```diff
@@ -384,11 +305,6 @@ func (self *RouterManager) HandleRouterDelete(id string) {
 	} else {
 		log.Debug("deleted router not connected, no further action required")
 	}
-
```
plorenz, line 387: moved to Network

controller/model/service_manager.go
```diff
@@ -26,61 +26,40 @@ import (
 	"github.com/openziti/ziti/controller/fields"
 	"github.com/openziti/ziti/controller/models"
 	"github.com/orcaman/concurrent-map/v2"
-	"github.com/pkg/errors"
 	"go.etcd.io/bbolt"
 	"google.golang.org/protobuf/proto"
-	"reflect"
 	"time"
 )
```
plorenz, line 33: Moved to service_model.go

controller/model/terminator_manager.go
```diff
@@ -30,131 +30,52 @@ import (
 	"github.com/openziti/ziti/controller/fields"
 	"github.com/openziti/ziti/controller/models"
 	"github.com/openziti/ziti/controller/xt"
-	"github.com/pkg/errors"
 	"go.etcd.io/bbolt"
 	"google.golang.org/protobuf/proto"
 	"reflect"
 	"strings"
 	"time"
 )
```
plorenz, line 39: moved to terminator_model.go

controller/model/terminator_manager.go
```diff
@@ -223,34 +144,18 @@ func (self *TerminatorManager) Update(entity *Terminator, updatedFields fields.U
...
+	return self.GetDb().Update(ctx, func(ctx boltz.MutateContext) error {
 		if cmd.Entity.IsSystemEntity() {
 			ctx = ctx.GetSystemContext()
 		}
 
 		self.checkBinding(terminator)
-		return self.GetStore().Update(ctx, terminator.toBolt(), cmd.UpdatedFields)
-	})
-}
-
```
plorenz, line 235: Read and readInTx are replaced by generic versions in base_manager

michaelquigley (approved):
🚢

PR #2220, opened 2024-07-12
Fix responsive check. Fixes #2171. Add router-connections inspection. Fixes #2219. Improve can't load data file log message. Fixes #2195

### LINKED ISSUE #2171: Routers should consider control channels unresponsive if they are not connected
right now unresponsive is measured purely off of latency measurements. That logic can consider an unconnected control channel responsive if the channel was disconnected after a latency response is received and before the next one is sent.

### LINKED ISSUE #2195: cached data model file set to
Hello, I am showing the following error when starting the public route, I have not seen this error in the previous version.
How can I solve this problem so that the public route starts smoothly
image

router/internal/edgerouter/config.go:154
func:github.com/openziti/ziti/router/internal/edgerouter.(*Config).LoadConfigFromMap
level:info
msg:cached data model file set to: /home/work/projec/related/www.router.com/www.router.com.yaml.json.gzip
time:2024-07-03T09:13:24.195Z

error:open  /home/work/projec/related/www.router.com/www.router.com.yaml.json.gzip: no such file or directory
/router/state/manager.go:176
func:github.com/openziti/ziti/router/state.(*ManagerImpl).LoadRouterModel
level:error
msg:could not load router model from file [ /home/work/projec/related/www.router.com/www.router.com.yaml.json.gzip]
time:2024-07-03T09:13:24.196Z

### LINKED ISSUE #2219: Add inspection for router connections
We want to be able to inspect the state of a router's connections. In a distributed controller environment this can help us figure out why a router is sending requests to a given controller vs a different one.

Example output:

```
$ ziti fabric inspect router-controllers 
Results: (1)
yPnQ1sjc2p.router-controllers
Controllers:
  ctrl1:
    address: tls:localhost:6262
    connected: true
    controllerId: ctrl1
    latency: 2.486287ms
    responsive: true
    timeSinceLastContact: 5.197s
    version: v0.0.0
  ctrl2:
    address: tls:localhost:6363
    connected: true
    controllerId: ctrl2
    latency: 509.058µs
    responsive: true
    timeSinceLastContact: 5.214s
    version: v0.0.0
  ctrl3:
    address: tls:localhost:6464
    connected: true
    controllerId: ctrl3
    latency: 2.937728ms
    responsive: true
    timeSinceLastContact: 5.198s
    version: v0.0.0
```

plorenz:


router/state/manager.go
```diff
@@ -210,7 +210,7 @@ func (sm *ManagerImpl) LoadRouterModel(filePath string) {
 	model, err := common.NewReceiverRouterDataModelFromFile(filePath, RouterDataModelListerBufferSize)
 
 	if err != nil {
-		pfxlog.Logger().WithError(err).Errorf("could not load router model from file [%s]", filePath)
+		pfxlog.Logger().WithError(err).Infof("could not load router model from file [%s]", filePath)
```
dovholuknf, line 213: WithError but Info? this feels strange to be an info?

dovholuknf (approved):

router/state/manager.go
plorenz: It's usually not an error because on initial startup the file won't exist. If the file doesn't exist, we'll just get the full state from the controller again, so it doesn't prevent anything from working. Given that, it feels like it doesn't reach the level of error.

PR #2227, opened 2024-07-12
Sort connected routers first for session ers. Increase er limit to 25. Fixes #2046

### LINKED ISSUE #2046: Edge Routers list for edge sessions should be sorted by online status first
When creating an Edge Session, the system currently sorts the Edge Routers by id, and limits the overall list to 10.  This can create a situation where online routers are excluded and offline routers are in the session "slots".  The list should be sorted first by online status, to give the identity the most robust operational state.

plorenz:


controller/model/edge_router_manager.go
```diff
@@ -232,6 +234,16 @@ func (self *EdgeRouterManager) IsAccessToEdgeRouterAllowed(identityId, serviceId
 	return result, nil
 }
 
+func (self *EdgeRouterManager) ListForIdentityAndServiceWithTx(tx *bbolt.Tx, identityId, serviceId string) (*EdgeRouterListResult, error) {
+	query := fmt.Sprintf(`anyOf(identities) = "%v" and anyOf(services) = "%v" SORT BY connected DESC LIMIT 25`, identityId, serviceId)
```
dovholuknf, line 238: still feels like a moving target. feels like 25 shouldn't be hardcoded in this way. also feels like maybe this should be randomized somehow?

dovholuknf (approved):
don't agree with the hardcoded 25 but other than that, lgtm

controller/model/edge_router_manager.go
plorenz: I agree that this should be configurable in some way. Most flexible would probably be per service or per identity. This fix will solve the immediate issue, and hopefully we can make it more flexible with some follow-up work.

PR #2242, opened 2024-07-16
fixed #2240 secondary ext-jwt processing would fail...

### LINKED ISSUE #2240: Secondary ext-jwt Auth Policy check incorrectly requires primary ext-jwt auth to be enabled
Configuring an authentication policy with primary cert + required secondary ext-jwt incorrectly fails with a controller error stating that the primary ext-jwt capability is not allowed on the authentication policy.

andrewpmartinez:
...if ext-jwt primary wasn't enabled.

- fixes extra JWT processing on authentication endpoints (that would never pass)
- updates error message for missing NamedIndexStore implementations
- fixes ext-jwt store error stating that a named index is not present


plorenz (approved):

andrewpmartinez:
Added bulk processing for primary plus secondary JWT processing (if configured)

dovholuknf (approved):
string literals should be constants imo

PR #2257, opened 2024-07-24
Migrated edge router tunneler code to use the router data model. Fixes #2121

### LINKED ISSUE #2121: Use router data model for edge router tunnel
This should allow the ER/T to react to model changes, including add/remove/update of services as well as changes to the edge router identity. This should also fix the issue where terminator creates go into a loop sometimes.

The implementation should also remove the need for api-sessions and sessions.

plorenz:


controller/command/command.go
```diff
@@ -65,6 +66,10 @@ func (self *LocalDispatcher) IsLeaderOrLeaderless() bool {
 	return true
 }
 
+func (self *LocalDispatcher) IsLeaderless() bool {
+	return false
```
andrewpmartinez, line 70: Why is this always fase?

controller/handler_edge_ctrl/common.go
```diff
@@ -377,6 +377,31 @@ func (self *baseSessionRequestContext) verifyIdentityEdgeRouterAccess() {
 	}
 }
 
+func (self *baseSessionRequestContext) verifyEdgeRouterServiceBindAccess() {
+	if self.err == nil {
```
andrewpmartinez, line 381: I assume this is to skip unnecessary work and that `.err` being set is resolved somewhere else.

andrewpmartinez (approved):

controller/command/command.go
plorenz: LocalDispatcher is only used when running in non-HA mode, without raft, so there's only one node which is the equivalent of being the leader.

PR #2280, opened 2024-08-01
Ensure controller is initialized before raft. Fixes #2279

### LINKED ISSUE #2279: Ensure xweb initialized before RAFT
There's race condition where raft can try to use xweb before xweb is initialized if the order is wrong.

plorenz:


ekoby (approved):
looks good, I don't hit that issue with this change

andrewpmartinez (approved):

PR #2292, opened 2024-08-05
also send partial chain GetChain() when computing the spiffe id

qrkourier:


qrkourier:
This addresses a problem where the client cert and CA pool alone do not construct a full chain because the intermediate is ignored when loaded from the client cert file.

andrewpmartinez (approved):

PR #2303, opened 2024-08-06
conditionally add trust domains to controller config

qrkourier:
Stop hard-coding controller config YAML `trustDomain: quickstart` in production deployments for Linux and Docker, and instead allow optionally configuring `trustDomain`


qrkourier:
I have preserved the settable `trustDomain` implementation in [this tag](https://github.com/qrkourier/ziti/commits/create-config-controller-settable-trustdomain/) (in Ken's fork) in case we decide to resurrect that idea.

PR #2304, opened 2024-08-07
set spiffe id when generating a controller pki

qrkourier:
Add SPIFFE ID to the Linux and Docker controller's certs

### DIFF
--- a/controller/config/config.go
+++ b/controller/config/config.go
@@ -23,6 +23,12 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
+	"math"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+
 	"github.com/hashicorp/go-hclog"
 	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/channel/v3"
@@ -38,11 +44,6 @@ import (
 	"github.com/openziti/ziti/router/xgress"
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
-	"math"
-	"net/url"
-	"os"
-	"strings"
-	"time"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


qrkourier:
This change is not upgade-safe and I have changed the PR status to DRAFT. 

This change would be safe if:

1. we re-issue the full chain instead of only the leaf, or
2. we stop re-issuing the leaf, or
3. we conditionally set spiffe id in the leaf

controller/config/config.go
```diff
@@ -680,50 +681,62 @@ func isSelfSigned(cert *x509.Certificate) (bool, error) {
...
+		fingerprint := fmt.Sprintf("%x", rawHash)
+		idStr := "spiffe://" + fingerprint
 
-	spiffeId, err := url.Parse(idStr)
+		spiffeId, err := url.Parse(idStr)
 
-	if err != nil {
-		return nil, fmt.Errorf("could not parse generated SPIFFE id [%s] as a URI: %w", idStr, err)
-	}
+		if err != nil {
```
andrewpmartinez, line 733: What is w/ all the whitespace change?

andrewpmartinez: I wouldn't expect formatting changes due to Go's opinionated formatting and that Goland auto formats most things.

qrkourier: For some reason, the `Identity.ServerCert()` type is a slice of cert byte slices, whereas `Identity.Cert()` is a slice of bytes, so I needed to indent one additional iteration or limit to only the first chain. 

I went with iterating, though it's admittedly unclear to me when there could be more than one. This prompted me to change some returns to continue to allow for a subsequent iteration to find a complete chain and generate the default spiffe id.

qrkourier: More precisely,

```go
type Identity interface {
	Cert() *tls.Certificate
	ServerCert() []*tls.Certificate
        ...
}
```

I was referring to the fact that Cert() is a point to a cert, whereas ServerCert() is a slice of pointers, so I need one more indent in order to iterate over the slice.

dist/dist-packages/linux/openziti-controller/bootstrap.bash
```diff
@@ -71,27 +72,43 @@ issueLeafCerts() {
...
       --pki-root "${ZITI_PKI_ROOT}" \
       --ca-name "${ZITI_INTERMEDIATE_FILE}" \
       --key-file "${ZITI_SERVER_FILE}" \
       --client-file "${ZITI_CLIENT_FILE}" \
-      --allow-overwrite >&3  # write to debug fd because this runs every startup
+      --allow-overwrite
+    )
+
+    # fall back to renewing without SPIFFE ID if issuer cert lacks trust domain in URI SAN
+    "${_create_client[@]}" --spiffe-id "/controller/${ZITI_CLIENT_FILE}" >&3 || {
```
qrkourier, line 108: The spiffe id for client and server certs should be the same, not different

controller/config/config.go
qrkourier: If the first chain doesn't resolve to self-signed, should we return with an error instead of iterating then? We should use an alt chain for the default spiffy ID, right?

andrewpmartinez: I am not sure that this code would defend against a full root chain being provided in altServerCerts that also has SPIFFE IDs.

qrkourier: In case of conflation, this is the portion of the code that generates a default id as a last resort, not the part that looks for SPIFFE IDs in URI SANs.

andrewpmartinez: That is correct. However, I still see an issue if an alt server cert + root is provided.

It will be to the wrong root. We would want the root of the network's PKI not some other random PKI. Unfortunately the identity interface doesn't distinguish between the piles of certs.

qrkourier: If the first chain doesn't resolve to self-signed, should we return with an error instead of iterating then? We should NOT use an alt chain for the default spiffy ID, right?

qrkourier:
Reverting to DRAFT status pending availability of an Identity library tag that presents only the default server cert chain.

qrkourier:
Not solving this for existing quickstarts/deployments right now: the workaround is to set `trustDomain` in the controller config.

PR #2310, opened 2024-08-08
Linux docker configure console

qrkourier:
I combined two changes in this PR, which can be split if necessary.

1. enable generating a controller config with the `zac` binding
2. enable the Linux/Docker deployments to use this option when appropriate


ziti/cmd/create/create_config_controller.go
```diff
@@ -253,27 +267,44 @@ func SetWebConfig(data *ControllerTemplateValues) {
 	SetWebIdentityCA(data)
 	SetCtrlAltServerCerts(data)
 }
+
+func SetConsoleConfig(v *ConsoleValues, o *ConsoleOptions) {
+	if o.Location == "DISABLED" {
```
dovholuknf, line 272: This should test for "" instead. the default location should be {{ .ZitiHome }} and not a linux-path

ziti/cmd/create/create_config_controller.go
```diff
@@ -152,6 +159,7 @@ func (options *CreateConfigControllerOptions) addFlags(cmd *cobra.Command) {
 	cmd.Flags().DurationVar(&options.EdgeIdentityEnrollmentDuration, optionEdgeIdentityEnrollmentDuration, edge.DefaultEdgeEnrollmentDuration, "the edge identity enrollment duration, use 0h0m0s format")
 	cmd.Flags().DurationVar(&options.EdgeRouterEnrollmentDuration, optionEdgeRouterEnrollmentDuration, edge.DefaultEdgeEnrollmentDuration, "the edge router enrollment duration, use 0h0m0s format")
 	cmd.Flags().IntVar(&options.MinCluster, optionMinCluster, 0, "minimum cluster size. Enables HA mode if > 0")
+	cmd.Flags().StringVar(&options.Console.Location, "console-location", "DISABLED", "enable and set path to console static files")
```
dovholuknf, line 162: the default should probably be "" here and don't overload a boolean into a path. if the location is set to anything non-empty it's 'set'

dovholuknf (approved):

PR #2320, opened 2024-08-12
add docker deployment test workflow

qrkourier:


dist/docker-images/ziti-controller/compose.yml
```diff
@@ -3,6 +3,10 @@ volumes:
   ziti-controller:
     driver: local
 
+networks:
+  ziti:
+    driver: bridge
```
dovholuknf, line 8: why was/is this necessary?

dist/dist-packages/linux/openziti-router/bootstrap.bash
```diff
@@ -69,13 +69,15 @@ enroll() {
...
+    _ziti_home="$(dirname "${_config_file}")"
     shift
   else
-    echo "ERROR: no config file path provided" >&2
+    echo "ERROR: no config file path provided in first param" >&2
     return 1
   fi
 
-  if [[ ! -s "${ZITI_HOME}/${ZITI_ROUTER_IDENTITY_CERT}" || "${1:-}" == --force ]]; then
+  if [[ ! -s "${_ziti_home}/${ZITI_ROUTER_IDENTITY_CERT}" || "${1:-}" == --force ]]; then
```
dovholuknf, line 80: in general, this line looks wrong to me now that it's singled out. i would not expect this to be prepended with any path. it should just be `${ZITI_ROUTER_IDENTITY_CERT}`.

dist/docker-images/ziti-cli/Dockerfile
```diff
@@ -21,6 +21,8 @@ ARG TARGETARCH
 ARG ZUID=2171
 ARG ZGID=2171
 
+ARG HOME=/home/ziggy
```
dovholuknf, line 24: small changes like this might cause issues downstream for anything using this container. i'd rather not make this particular change (which means the ones below need to change back too)

.github/workflows/test-docker-deployments.yml
```diff
@@ -0,0 +1,111 @@
...
+      ZITI_PWD: ziggylicious
+      ZITI_CTRL_ADVERTISED_ADDRESS: ctrl1.127.21.71.0.sslip.io
+      ZITI_ROUTER_ADVERTISED_ADDRESS: router1.127.21.71.0.sslip.io
+      ZITI_ROUTER_NAME: router1
+      # define the compose project files: compose.yml and compose.test.yml files for ziti-controller and ziti-router
+      COMPOSE_FILE: ${{ github.workspace }}/dist/docker-images/ziti-controller/compose.yml:${{ github.workspace }}/dist/docker-images/ziti-controller/compose.test.yml:${{ github.workspace }}/dist/docker-images/ziti-router/compose.yml:${{ github.workspace }}/dist/docker-images/ziti-router/compose.test.yml
+      ZITI_CONTROLLER_IMAGE: ziti-controller:local
+      ZITI_ROUTER_IMAGE: ziti-router:local
+    steps:
+      - name: Shallow checkout
```
dovholuknf, line 31: I strongly dislike tests that can't be run easily and succinctly locally. though it looks nice within the GitHub UI, these sorts of things should be easily performed locally.

I'd much rather see most of these steps condensed into a local script that someone can run locally.

dist/docker-images/ziti-controller/compose.test.yml
```diff
@@ -0,0 +1,91 @@
+volumes:
+  ziti-login:
```
dovholuknf, line 2: why is this volume even needed? it just adds complexity to this compose file and I don't see the value

dist/docker-images/ziti-controller/compose.test.yml
```diff
@@ -0,0 +1,91 @@
...
+
+  chown-login:
+    profiles:
+      - test
+    image: busybox
+    command: chown -R ${ZIGGY_UID:-2171} /ziti-login
+    volumes:
+      - ziti-login:/ziti-login
+
+  ziti-login:
```
dovholuknf, line 57: why was this even needed? the tests wait for the controller and login as well? i would not expect any of this?

dovholuknf (changes requested):
needs some changes

dist/docker-images/ziti-controller/compose.yml
qrkourier: Defining a custom network, "ziti" in this case, is a prerequisite to defining any compose service aliases, which is itself a prerequisite for ziti to have a single FQDN that is valid both inside and outside of the Docker network.

In other words, we're defining a DNS name for the controller that can be used by other containers inside the Docker network as well as all other clients outside the network.

This is the best way to enable containers inside the Docker network to operate in a ziti-normal way because it is non-trivial to enable hairpinning, and not possible to reliably enable hair-pinning at all.

dist/dist-packages/linux/openziti-router/bootstrap.bash
qrkourier: I found another way that doesn't require this to be a relative path.

dist/docker-images/ziti-cli/Dockerfile
qrkourier: It should have been this way since we enabled CLI auto-complete, which involved defining a HOME dir. In other words, this expresses one of the assumptions that's built in to the container image. This is a default value, not an override, so anyone who happens to be using this container with some other value for HOME will not be affected. This can only cause things to start working correctly, not break.

dist/docker-images/ziti-controller/compose.test.yml
qrkourier: The test relies on a successful login as a proxy for a successful controller startup. After the mgmt API is available, it will attempt to use it.

dist/dist-packages/linux/openziti-router/bootstrap.bash
qrkourier: I found another way that doesn't require this to be a relative path.

.github/workflows/test-docker-deployments.yml
qrkourier: I'm also hungry for a way to run Actions locally. I'd guess it would frequently save me a significant amount of time. It's probably possible for most of our jobs. I'd have to go back to the drawing board on this one because so much of it is scaffolding from other jobs that, like this one, are intended to run in GitHub. Adapting all of our jobs to shell scripts doesn't seem like the right solution unless the job doesn't benefit from any GitHub Actions.

.github/workflows/test-docker-deployments.yml
```diff
@@ -0,0 +1,36 @@
+name: Test Deployments
+on:
+  workflow_dispatch:
```
dovholuknf, line 3: i'm not convinced it's actually worthwhile to add this to every PR. I'm more like 60/40 that it feels like this is better as a test that's run prior to publishing artifacts and not every PR. I don't merge to ziti enough to know how annoying this test will be ... probably best to ask the group to see what they think.

dist/docker-images/ziti-controller/compose.test.bash
```diff
@@ -0,0 +1,95 @@
...
+ZITI_CTRL_ADVERTISED_ADDRESS="ctrl1.127.21.71.0.sslip.io" \
+ZITI_CTRL_ADVERTISED_PORT="1280" \
+ZITI_CONTROLLER_IMAGE="ziti-controller:local" \
+ZITI_ROUTER_IMAGE="ziti-router:local" \
+ZITI_ROUTER_NAME="router1"
+
+export ZITI_ROUTER_ADVERTISED_ADDRESS="${ZITI_ROUTER_NAME}.127.21.71.0.sslip.io" \
+ZITI_ENROLL_TOKEN="/home/ziggy/.config/ziti/${ZITI_ROUTER_NAME}.jwt"
+
+mkdir -p ./release/amd64/linux
```
dovholuknf, line 43: why not use a /tmp dir? alternatively, i don't think there's a need for "amd64" in this path, is there?

dist/docker-images/ziti-controller/compose.test.bash
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env bash
```
dovholuknf, line 1: this worked great once i stopped my local controller. might be worthwhile to test for port 1280 being open first before continuing, but overall this worked really well. thanks for changing it.

dist/docker-images/ziti-controller/compose.test.bash
```diff
@@ -0,0 +1,95 @@
...
+
+cleanup(){
+	docker compose --profile test down --volumes --remove-orphans
+}
+
+BASEDIR="$(cd "$(dirname "${0}")" && pwd)"
+REPOROOT="$(cd "${BASEDIR}/../../.." && pwd)"
+cd "${REPOROOT}"
+
+[[ -s ./go.work ]] && {
```
dovholuknf, line 18: i'm not sold on this one way or the other. seems like it might be useful to be able to build/test a docker container that uses a specific set of dependencies. that or allow someone to pass in the pre-built binary location so they can work around this sort of decision that might not make sense for them.

dist/docker-images/ziti-controller/compose.test.bash
qrkourier: It's an artifact of the symbols used by `gox` for OS and architecture, and the patterns used by our build that wraps `gox`. 

The release workflow builds container images by stuffing the release ziti binary into the image, avoiding the need to re-build it. That Dockerfiles used in that workflow adopt the same `gox` variables and patterns to locate the binary.

This test workflow avoids changing the Dockerfiles by further extending the pattern, e.g., the directory structure.

TL;DR Some things resemble the release workflow because the test is designed to exercise the release, not to express a minimal pattern for its own sake.

dist/docker-images/ziti-controller/compose.test.bash
qrkourier: I agree. I'd encountered a glitch when I first attempted to build with a workspace, but this becomes a more useful kind of local smoketest if that can be solved. I'll peck at it.

dist/docker-images/ziti-controller/compose.test.bash
qrkourier: Good point. There's no reason for this test to require all interfaces since it's only using 127.21.71.0, and it can use a non-standard port to further reduce the chance of a collision. Checking for an open port is also doable, but makes the script less portable and slightly more complicated. I'll take a stab at a minimal port check.

.github/workflows/test-docker-deployments.yml
qrkourier: I aim to surface issues as early as practical in the life cycle. Advertised this change to the team in case it's perceived as a problem.

dovholuknf (approved):

PR #2322, opened 2024-08-14
test linux deployments

qrkourier:


dist/docker-images/compose.test.bash
```diff
@@ -13,17 +13,18 @@ cleanup(){
 
 portcheck(){
     PORT="${1}"
-    if [ "$(nc -zv -G 5 localhost "$PORT" 2>&1 | grep -c 'succeeded')" -lt 1 ]; then
```
qrkourier, line 16: Always returns error on Linux which is interpreted here as the port is available even if it is not available. This is because `-G` isn't valid for  the OpenBSD netcat or Nmap's ncat which are both commonly used in Linux.

Refactored to trust the exit code when doing a simple SYN check.

dovholuknf: nice

.github/workflows/test-deployments.yml
```diff
@@ -0,0 +1,236 @@
...
+        with:
+          nfpm_version: ${{ env.NFPM_VERSION }}
+          packager: ${{ matrix.packager }}
+          config: dist/dist-packages/linux/nfpm-${{ matrix.package_name }}.yaml
+          target: release/
+        env:
+          ZITI_VERSION: 0.0.0
+          ZITI_MAINTAINER: "OpenZiti Maintainers <developers@openziti.org>"
+          ZITI_HOMEPAGE: "https://openziti.io"
+          ZITI_VENDOR: "NetFoundry, Inc."
```
dovholuknf, line 65: It never has a comma. `NetFoundry Inc.`

dist/dist-packages/linux/linux.test.bash
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env bash
```
dovholuknf, line 1: there's a fair bit of duplicated code in the docker test as here. can you extract this into two steps?
a.) install the things
b.) test the things

Then the "test the things" section would be identitcal and reusable across any / all deployment mechanism

qrkourier: That sounds great but I only see one redundant line:

```bash
go test -v -count=1 -tags="quickstart manual" ./ziti/cmd/edge/...
```

PR #2348, opened 2024-08-23
Add verify traffic verify network

dovholuknf:
These two commands are nested under 'ops' and are meant to help people with verifying their configurations.

`verify-network` will allow you to specify config files which will then be probed and/or inspected looking for common issues such as unroutable addresses etc.

`verify-traffic` will test the deployed network making sure it can send traffic over the overlay. example output
image

### DIFF
--- a/internal/rest/mgmt/helpers.go
+++ b/internal/rest/mgmt/helpers.go
@@ -79,5 +79,5 @@ func ServicePolicyFromFilter(client *rest_management_api_client.ZitiEdgeManageme
 }
 
 func NameFilter(name string) string {
-	return "name=\"" + name + "\""
+	return `name="` + name + `"`
 }
--- a/internal/rest/mgmt/helpers.go
+++ b/internal/rest/mgmt/helpers.go
@@ -1,0 +1,83 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+package mgmt
+
+import (
+	"context"
+	"github.com/openziti/edge-api/rest_management_api_client"
+	"github.com/openziti/edge-api/rest_management_api_client/identity"
+	"github.com/openziti/edge-api/rest_management_api_client/service"
+	"github.com/openziti/edge-api/rest_management_api_client/service_policy"
+	"github.com/openziti/edge-api/rest_model"
+	log "github.com/sirupsen/logrus"
+	"time"
+)
+
+func IdentityFromFilter(client *rest_management_api_client.ZitiEdgeManagement, filter string) *rest_model.IdentityDetail {
+	params := &identity.ListIdentitiesParams{
+		Filter:  &filter,
+		Context: context.Background(),
+	}
+	params.SetTimeout(5 * time.Second)
+	resp, err := client.Identity.ListIdentities(params, nil)
+	if err != nil {
+		log.Debugf("Could not obtain an ID for the identity with filter %s: %v", filter, err)
+		return nil
+	}
+
+	if resp == nil || resp.Payload == nil || resp.Payload.Data == nil || len(resp.Payload.Data) == 0 {
+		return nil
+	}
+	return resp.Payload.Data[0]
+}
+
+func ServiceFromFilter(client *rest_management_api_client.ZitiEdgeManagement, filter string) *rest_model.ServiceDetail {
+	params := &service.ListServicesParams{
+		Filter:  &filter,
+		Context: context.Background(),
+	}
+	params.SetTimeout(5 * time.Second)
+	resp, err := client.Service.ListServices(params, nil)
+	if err != nil {
+		log.Debugf("Could not obtain an ID for the service with filter %s: %v", filter, err)
+		return nil
+	}
+	if resp == nil || resp.Payload == nil || resp.Payload.Data == nil || len(resp.Payload.Data) == 0 {
+		return nil
+	}
+	return resp.Payload.Data[0]
+}
+
+func ServicePolicyFromFilter(client *rest_management_api_client.ZitiEdgeManagement, filter string) *rest_model.ServicePolicyDetail {
+	params := &service_policy.ListServicePoliciesParams{
+		Filter:  &filter,
+		Context: context.Background(),
+	}
+	params.SetTimeout(5 * time.Second)
+	resp, err := client.ServicePolicy.ListServicePolicies(params, nil)
+	if err != nil {
+		log.Errorf("Could not obtain an ID for the service with filter %s: %v", filter, err)
+		return nil
+	}
+	if resp == nil || resp.Payload == nil || resp.Payload.Data == nil || len(resp.Payload.Data) == 0 {
+		return nil
+	}
+	return resp.Payload.Data[0]
+}
+
+func NameFilter(name string) string {
+	return "name=\"" + name + "\""
+}
--- a/ziti/cmd/verify/common.go
+++ b/ziti/cmd/verify/common.go
@@ -20,6 +20,13 @@ import (
 	"runtime"
 )
 
+type controller struct {
+	user string
+	pass string
+	host string
+	port string
+}
+
 func configureLogFormat() {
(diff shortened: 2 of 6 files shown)


ziti/cmd/verify/common.go
```diff
@@ -0,0 +1,48 @@
...
+
+type controller struct {
+	user string
+	pass string
+	host string
+	port string
+}
+
+func configureLogFormat() {
+	log.SetFormatter(&log.TextFormatter{
```
andrewpmartinez, line 31: I haven't seeen something call `log.SetFormatter` in a long time. Most of our code calls `pfxlog.Logger()` and gets the globally configured logger.

I don't understand why we would need to do it here over use the established patterns, unless there is something I'm missing.

internal/rest/mgmt/helpers.go
```diff
@@ -0,0 +1,83 @@
+/*
```
andrewpmartinez, line 1: These are the kinds of functions I'd like to see moved into the Go SDK and a proper wrapper developed.

However, they would need to be changed:

-  to take in an external `context` as that is how external developers control the lifecycle of long-running asynchronous tasks. Using context.Background() is an orange flag for a public externally focused function
- would need to work with the API abstractions there instead of directly using the `edge-api` repo. This code does not support HA failover (like all of our CLI tooling at the moment).

internal/rest/mgmt/helpers.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil
+	}
+	if resp == nil || resp.Payload == nil || resp.Payload.Data == nil || len(resp.Payload.Data) == 0 {
+		return nil
+	}
+	return resp.Payload.Data[0]
+}
+
+func NameFilter(name string) string {
+	return "name=\"" + name + "\""
```
andrewpmartinez, line 82: You can avoid escaping by using ticked strings:

> println(fmt.Sprintf(`name="%s"`, name))
or
> println(`name="` + name + `"`)

ziti/cmd/verify/ops_verify_network.go
```diff
@@ -0,0 +1,263 @@
...
+	controllerConfig string
+	routerConfig     string
+	verbose bool
+}
+
+func NewVerifyNetwork() *cobra.Command {
+	n := &network{}
+
+	cmd := &cobra.Command{
+		Use:   "verify-network",
```
andrewpmartinez, line 43: Should probably be `verify-network -c <controller-config-file> -r <router-config-file> [--verbose]`

ziti/cmd/verify/ops_verify_network.go
```diff
@@ -0,0 +1,263 @@
...
+	return cmd
+}
+
+type StringMap map[string]interface{}
+
+func (m StringMap) mapFromKey(key string) StringMap {
+	if v, ok := m[key]; ok {
+		return v.(StringMap)
+	}
+	log.Fatalf("map didn't contain key %s", key)
```
andrewpmartinez, line 83: It is very odd for a data structure like this to have a fatal error out instead of returning an error and letting the caller decide.

ziti/cmd/verify/ops_verify_network.go
```diff
@@ -0,0 +1,263 @@
...
+
+func (m StringMap) listFromKey(key string) StringMapList {
+	if v, ok := m[key]; ok {
+		return v.([]interface{})
+	}
+	log.Fatalf("map didn't contain key %s", key)
+	return nil
+}
+
+func (p ProtoHostPort) testPort(msg string) bool {
```
andrewpmartinez, line 97: Style choice, but these usually appear after the type definition.

ziti/cmd/verify/ops_verify_network.go
```diff
@@ -0,0 +1,263 @@
...
+	proto string
+	host  string
+	port  string
+}
+
+func (p ProtoHostPort) address() string {
+	return p.host + ":" + p.port
+}
+
+func fromString(input string) *ProtoHostPort {
```
andrewpmartinez, line 120: In most Go code, if this was scoped to a package, you would see a `portHostPort.New()` or `portHostPort.new()` function. Mixed into a larger package, I would typically expect to see `NewPortHostPortFromString(...)` or the like.

ziti/cmd/verify/ops_verify_network.go
```diff
@@ -0,0 +1,263 @@
...
+			anyFailure = true
+		} else {
+			log.Infof("listener binding[%d] is valid", pos)
+		}
+		pos++
+	}
+	return anyFailure
+}
+
+func stringOrNil(input interface{}) string {
```
andrewpmartinez, line 229: This is really `stringOrEmpty`

ziti/cmd/verify/ops_verify_network.go
```diff
@@ -0,0 +1,263 @@
...
+		panic(err)
+	}
+	advertiseAddress := stringOrNil(ctrlCfg.mapFromKey("ctrl").mapFromKey("options")["advertiseAddress"])
+	host := fromString(advertiseAddress)
+	anyFailure := host.testPort("controller advertise address")
+
+	web := ctrlCfg.listFromKey("web")
+
+	log.Infof("verifying %d web entries", len(web))
+	for _, item := range web {
```
andrewpmartinez, line 159: This is fine for iteration one, but I'd rather see us rework the config parsing so it can be reused.

This code will silently break if the configuration files are reworked.

ziti/cmd/verify/ops_verify_traffic.go
```diff
@@ -0,0 +1,575 @@
...
+	serverIdName string
+	clientIdName string
+	bindSPName   string
+	dialSPName   string
+}
+
+func NewVerifyTraffic() *cobra.Command {
+	t := &traffic{}
+	cmd := &cobra.Command{
+		Use:   "verify-traffic",
```
andrewpmartinez, line 65: It should probably include the usage flags like the above.

ziti/cmd/verify/ops_verify_traffic.go
```diff
@@ -0,0 +1,575 @@
...
+	read, err := zitiReader.ReadString('\n')
+	if err != nil {
+		log.Errorf("error reading from reader: %v", err)
+	} else {
+		log.Debugf("Received: %s", strings.TrimSpace(read))
+	}
+	return nil
+}
+
+func terminatorExists(client *rest_management_api_client.ZitiEdgeManagement, serviceName string) bool {
```
andrewpmartinez, line 235: Is there a reason this isn't next to the other API helpers in helpers.go?

andrewpmartinez (approved):
I think this PR exposes issues w/ our config parsing doing too much in one step instead of allowing a `parse > verify > use` pattern.

However doing that work will take quite a bit of effort, but should be done if this set of commands is useful as they will break sometime in the future when we forget this code exists.

Read through my comments, there isn't anything major, mostly style. Address the linter, who hates me too.

ziti/cmd/verify/common.go
dovholuknf: largely it was just an aesthetics thing...  removing the timestamp, colorizing the level and removing the package/function. 

example before/after setting the format. if there's a way to remove the timestamp/package/function though, it'd work. I just didn't find it in options. i admit, i might not have known what to look for

image

It's also somewhat confusing to have a clone of it in the openziti repo but not be using it so i just kept it to "one fewer import"

internal/rest/mgmt/helpers.go
dovholuknf: that's why i put it into the internal folder... i am always cautious with my new, external functions since you can't ever take them back without deprecating etc.

anything "HA" would be spectacular to add here. I'd like to get "rev 1" out first without HA support but if there are other checks that make sense, I would expect this code to expand to help people with that too

internal/rest/mgmt/helpers.go
dovholuknf: indeed! i copied and pasted it outta something and didn't even think about it - i'll change it.

ziti/cmd/verify/ops_verify_network.go
dovholuknf: I'm torn. it's helpful, but it also feels easy to screw up when changing/adding flags. Also seems like with the flags enumerated on help it's less important. happy to discuss

ziti/cmd/verify/ops_verify_network.go
dovholuknf: changed StringMapList --> stringMapList. unintentionally public

ziti/cmd/verify/ops_verify_network.go
dovholuknf: moved types to the top

ziti/cmd/verify/ops_verify_network.go
dovholuknf: sure. it's not exported and i was trying to keep the total line length short and avoided any extra, helpful context. i can make it longer but preferred to not do that at this time

PR #2358, opened 2024-08-26
Make ziti agent snapshot-db match ziti fabric db snapshot. Fixes #2333

### LINKED ISSUE #2333: Give zit agent controller snapshot-db same capabilities as ziti fabric db snapshot
see https://github.com/openziti/fabric/issues/794 for features added

plorenz:


qrkourier (approved):
I confirmed the old CLI signature without the path still works and helpfully returns the full path to the auto-generated filename. When specifying the optional output file path, only the specified path is returned (not a problem, just an inconsistency).

PR #2362, opened 2024-08-28
Update 004-controller-pki.md

astroicers:
The path description for the Windows hosts file was updated from:

- Windows: `windows: %windir%\system32\drivers\etc\hosts`

to

- Windows: `%windir%\system32\drivers\etc\hosts`

This change removes the redundant "windows:" prefix, making the path clearer and more accurate for users.


dovholuknf (approved):
thank you looks good

PR #2395, opened 2024-09-10
stop hard-coding trustDomain

qrkourier:


dovholuknf (approved):

qrkourier:
This breaks new quickstarts < v1.1.9

dovholuknf:
> This breaks new quickstarts < v1.1.9

I don't think that statement is entirely accurate fwiw. but this pr seems fine to me :)

PR #2414, opened 2024-09-18
Add anonymize db utility. Fixes #2413

### LINKED ISSUE #2413: Add db anonymization utility
Add `ziti ops db anonymize` utility, which anonymizes a database.

plorenz:


plorenz:
> can this get a readme or anything that might tell the person what it will do (ideally enumerating the actions it'll take)? anything like that? also i feel like it should come with a warning like "we don't guarantee that this is 100% bug free" or somesuch...
> 
> Feels like it needs some 'legalize' even though IANAL.... stuff... just me?

Add this text to the help:

```
This utility attempts to remove any personal information from the db. It does the following
1. Renames all identities, services, edge-routers, policies and configs with generic names 
2. Makes all role attributes generic 
3. Replaces host.v1, host.v2 and intercept.v1 configs with generic versions
4. Deletes authenticators and enrollments.

WARNINGS:
* There may be personal information in other database fields, as this doesn't cover every field in every type.
* This works in place on the provided database. Make sure to run this on a copy of your database
```

Let me know if there's anything I should add to that

qrkourier (approved):

andrewpmartinez (approved):

PR #2421, opened 2024-09-19
Remove common header type from xgress messages since the overlap isn't complete

plorenz:


router/xgress/acker.go
```diff
@@ -7,7 +7,11 @@ import (
 	"sync/atomic"
 )
```
plorenz, line 9: Allow mocking the acker in tests

router/xgress/retransmitter.go
plorenz, line 1: Only change should be renaming retransmitter to self

router/xgress/acker.go
plorenz, line 1: Allow replacing the acker in tests

michaelquigley (approved):

PR #2434, opened 2024-09-23
fix quickstart release

qrkourier:


dovholuknf (approved):

qrkourier:
@dovholuknf I fixed these issues after your review.

- getZiti() always fails if ZITI_BIN_DIR and ZITI_VERSION_OVERRIDE are both set because _check_env_variable() always fails unless _check_env_variable() if run first to set the required values for the override version - solution is to swap the order so _check_env_variable() runs first in commit https://github.com/openziti/ziti/pull/2434/commits/7d86540ce6312103e3516134d75498dd79e08519
- some release tarballs have a nested directory "ziti" and some don't leading to inconsistent ziti binary path inside the quickstart container image or wherever getZiti() is run - solution is to handle both is a dir and not a dir in commit https://github.com/openziti/ziti/pull/2434/commits/f26e00235b90b561dc23b72f99a6baedc759cb9a

dovholuknf:
> some release tarballs have a nested directory "ziti" and some don't

sounds like the ziti archives need to be fixed to me

qrkourier:
> sounds like the ziti archives need to be fixed to me

I should clarify that old release tarballs had the nested directory, and current release tarballs have only the binary, not nested in a directory. The way I said it before may have sounded like it was inconsistently nested in current releases, which isn't the case.

PR #2437, opened 2024-09-24
linux, docker, k8s router - fix auto-renewal

qrkourier:
- fix for missing IP SAN issue reported in Discourse: https://openziti.discourse.group/t/can-zrok-be-used-to-expose-local-mqtt-broker-to-internet/3126/39?u=qrkourier
- fix for Linux router failing to auto-renew identity certs

### DIFF
--- a/dist/dist-packages/linux/openziti-router/bootstrap.bash
+++ b/dist/dist-packages/linux/openziti-router/bootstrap.bash
@@ -31,12 +31,6 @@ makeConfig() {
           ZITI_ROUTER_PORT \
           ZITI_ROUTER_LISTENER_BIND_PORT="${ZITI_ROUTER_PORT}"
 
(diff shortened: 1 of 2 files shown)


dist/dist-packages/linux/openziti-router/bootstrap.bash
```diff
@@ -31,6 +31,12 @@ makeConfig() {
           ZITI_ROUTER_PORT \
           ZITI_ROUTER_LISTENER_BIND_PORT="${ZITI_ROUTER_PORT}"
 
+  if [[ "${ZITI_ROUTER_ADVERTISED_ADDRESS}" =~ ^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
dovholuknf, line 34: same comment here as the other PR. it's better to not try to detect this sort of thing and instead give people a way to specifically tell the system that it's using an IP address.

dovholuknf (approved):

Anzal-calixto:
need fixing for issues causing the router to have incorrect subject alternative names

PR #2463, opened 2024-10-02
Update deps and changelog

plorenz:


controller/raft/mesh/mesh.go
```diff
@@ -247,7 +247,7 @@ type impl struct {
 	closeNotify          chan struct{}
 	closed               atomic.Bool
 	raftAccepts          chan net.Conn
-	bindHandler          channel.BindHandler
+	bindHandler          concurrenz.AtomicValue[channel.BindHandler]
```
dovholuknf, line 250: are these changes expected? they are more than "devs and changelog"

plorenz: It's a 'just in case' thread safety fix. Didn't turn out to be the issue, but it's still the safe thing to do, so I left it in

dovholuknf (approved):

PR #2479, opened 2024-10-11
fixes #2478 legacy auth queries missing values

### LINKED ISSUE #2474: Enable Ext JWT Enrollment/Generic Trust Bootstrapping
When using external JWT signers to authenticate in SDK apps, it is currently not accessible to bootstrap trust with the controller or locate it. Much like CA-based enrollment, it would be nice to have a generic, controller-wide JWT that can be used to communicate the controller(s) hostname/IP and a method to verify it.

### LINKED ISSUE #2478: JWT signer secondary auth: not enough information to continue
if I set up an auth policy with JWT signer as secondary auth and authenticate with key/cert I get an api session (irrelevant field removed):
```
  "authQueries": [
    {
      "provider": "url",
      "typeId": "EXT-JWT"
    }
  ],
  "identityId": "JTR.7wXlT",

  "isMfaComplete": false,
  "isMfaRequired": false
}
```

1. `authQueries` does not give enough information to start OIDC authentication (url, clientId, claim, audience are needed) 
     maybe external signer id/name would be enough to do a lookup
1. should `isMfaRequired` be set? I think that's the case with TOTP secondary

andrewpmartinez:


controller/internal/routes/current_api_session_api_model.go
```diff
@@ -90,6 +90,9 @@ func MapApiSessionAuthQueriesToRestEntity(ae *env.AppEnv, rc *response.RequestCo
 			MinLength:  authQuery.MinLength,
 			Provider:   authQuery.Provider,
 			TypeID:     authQuery.TypeID,
```
ekoby, line 92: I don't see these fields populated in API session returned from `/authenticate?method=cert` call

qrkourier:
seems to at least partially resolve https://github.com/openziti/ziti/issues/2478

ekoby:
> seems to at least partially resolve #2478

#2474 seems like a wrong reference

controller/internal/routes/current_api_session_api_model.go
andrewpmartinez: That would be odd for legacy, specifically in this PR, because there are tests in this PR that ensure that it is the exact thing.

andrewpmartinez: I also added a set of OIDC value checks, and the values I put in seem to come out.

andrewpmartinez:
> > seems to at least partially resolve #2478
> 
> #2474 seems like a wrong reference

Fixed.

PR #2483, opened 2024-10-14
start nameserver only if tproxy mode; resolves #2482

### LINKED ISSUE #2482: router run command - improperly binds 127.0.0.1:53/udp when tunnel mode is not tproxy
Re: https://openziti.discourse.group/t/ziti-router-from-quickstart-conflicting-with-existing-local-dns-after-update/3233/8?u=qrkourier

`ziti router run` always attempts to bind 127.0.0.1:53, even when the tunnel mode is not tproxy. For deployments where the mode is not tproxy and the run-as user is not root, this results in an error log and has no obvious effect on the router operation.

```json
{
  "error": "dns server failed to start: listen udp 127.0.0.1:53: bind: permission denied",
  "file": "github.com/openziti/ziti/router/xgress_edge_tunnel/tunneler.go:75",
  "func": "github.com/openziti/ziti/router/xgress_edge_tunnel.(*tunneler).Start",
  "level": "error",
  "msg": "failed to start DNS resolver. using dummy resolver",
  "time": "2024-10-14T09:05:16.665Z"
}
```

However, when the router is run-as root it unnecessarily and successfully binds the listening UDP socket and prevents other PIDs on the host from also binding 0.0.0.0:53.

qrkourier:


qrkourier:
the same change is also needed for rdm mode: https://github.com/openziti/ziti/pull/2715

PR #2489, opened 2024-10-18
makes OIDC issuer dynamic based on xweb bind point address

andrewpmartinez:
This makes each OIDC API exposure use its xweb bind point address as the hostname of the issuer rather than defaulting to the `edge.api.address` configuration value. The result is that the bind point address will be used in the well-known OIDC configuration.

### DIFF
--- a/controller/oidc_auth/provider.go
+++ b/controller/oidc_auth/provider.go
@@ -33,35 +33,58 @@ func NewNativeOnlyOP(ctx context.Context, env model.Env, config Config) (http.Ha
 	cert, kid, method := env.GetServerCert()
 	config.Storage = NewStorage(kid, cert.Leaf.PublicKey, cert.PrivateKey, method, &config, env)
 
-	oidcHandler, err := newHttpRouter(ctx, config)
+	handlers := map[string]http.Handler{}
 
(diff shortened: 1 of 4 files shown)


controller/oidc_auth/provider.go
```diff
@@ -33,35 +33,58 @@ func NewNativeOnlyOP(ctx context.Context, env model.Env, config Config) (http.Ha
...
 
-	oidcHandler, err := newHttpRouter(ctx, config)
+	handlers := map[string]http.Handler{}
 
-	openzitiClient := NativeClient(common.ClaimClientIdOpenZiti, config.RedirectURIs, config.PostLogoutURIs)
-	openzitiClient.idTokenDuration = config.IdTokenDuration
-	openzitiClient.loginURL = newLoginResolver(config.Storage)
-	config.Storage.AddClient(openzitiClient)
+	for _, issuer := range config.Issuers {
+		issuerUrl := "https://" + issuer + "/oidc"
```
dovholuknf, line 39: should this test for HasPrefix https:// to be friendly? or nah?

dovholuknf (approved):

PR #2513, opened 2024-10-31
refinements to ops verify-traffic to integrate and work with ziti edge login

dovholuknf:
the existing ziti ops verify-traffic were not integrated with the ziti edge login functionality. Jens needs/wants an example of how to integrate better with ziti login but apparently we haven't done that yet when using the edge-apis library.

This should provide a mechanism to obtain a new management client from the edge-apis library while integrating with ziti edge login as well. This will allow not only for updb style login but also enables cert/key/file-based login:
```
ziti ops verify-traffic --cleanup --file testadmin.json
```

Using an identity file greatly simplifies the command and is more secure than shipping a user/pwd. example:
image

### DIFF
--- a/controller/oidc_auth/storage.go
+++ b/controller/oidc_auth/storage.go
@@ -91,7 +91,7 @@ type HybridStorage struct {
 	serviceUsers cmap.ConcurrentMap[string, *Client]
 
 	startOnce sync.Once
-	issuer    string
+//linter issue	issuer    string
 	config    *Config
 
 	keys cmap.ConcurrentMap[string, *pubKey]
--- a/ziti/cmd/verify/ops_verify_network.go
+++ b/ziti/cmd/verify/ops_verify_network.go
@@ -17,6 +17,7 @@ package verify
 
 import (
 	"fmt"
+	"io"
 	"net"
 	"os"
 	"strings"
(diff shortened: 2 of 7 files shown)


controller/oidc_auth/storage.go
```diff
@@ -91,7 +91,7 @@ type HybridStorage struct {
 	serviceUsers cmap.ConcurrentMap[string, *Client]
 
 	startOnce sync.Once
```
plorenz, line 93: If this is unused, should we just remove it?

ziti/cmd/verify/ops_verify_network.go
```diff
@@ -48,7 +49,7 @@ type stringMapList []interface{}
 
 type StringMap map[string]interface{}
```
plorenz, line 51: Why add this, just to ignore it?

controller/oidc_auth/storage.go
dovholuknf: dunno. it's an @andrewpmartinez question. I just made the linter happy :) i was pointed it out to andrew...

ziti/cmd/verify/ops_verify_network.go
dovholuknf: meh - it was just a consistency thing... all the other commands pass these params so it was consistent. i had it with no params, but then verify-traffic needed in/out streams...  so -- yeah that's 'why'

plorenz (approved):

PR #2538, opened 2024-11-14
add quickstart support for ha

dovholuknf:
- add oidc to controller apis
- add ha flag support to router config gen
- add warn when creating a CA with a trust-domain with path information specified
- if a spiffe-id is supplied to server/client, use whatever is supplied
- if a trust-domain is provided to ca with a path, remove the path information and warn the user
- add hidden ha commands/flags
- add -- non-voting for ha
- add -- no-router flag to start just a controller if that's desirable


plorenz (approved):
Could use a github workflow to call the ha quickstart test, but otherwise, looks good to me :+1:

PR #2539, opened 2024-11-14
fixes #2505 adds central configuration with events

### LINKED ISSUE #2505: Centralized Configuration

andrewpmartinez:
Requires: https://github.com/openziti/edge-api/pull/137


plorenz:
My initial reaction is that having config settings by defined in a typed/static way is good for discoverability, but makes adding new values very complicated. I'm wondering if we want to either go with a loosely typed JSON-ey map type structure or at least restrict the structured config type to the presentation layer. 

If we go with the map structure all the way through, we lose discoverability and will rely more on doc. We'd probably also want code which checked that incoming config values names were in a hard-coded list. But adding new values would not required updating the REST/Model/Command/DB layers.

If we went with structured at the presentation layer, we wouldn't lose any discoverability and would only need to update that layer and would avoid needing to make updates to the model/command/db layers.

Thoughts?

PR #2545, opened 2024-11-15
Check peer certs

plorenz:


michaelquigley (approved):

andrewpmartinez:
There are two sides to certificate verification: server and client. Ensuring both sides properly verify and reject connections needs to be done. Statically analyzing the code (which is error prone) it appears that the tls.Config's used during dials are going to verify the server properly via go's built-in TLS logic. The server side verifies after the TSL connection during bind handling.

Since this area of code does not require flexible verification of client certificates, if possible, we should rely on Go's internal TLS implementation. This can be done by ensuring the tls.Config returned by calling `id.ServerTLSConfig()` has the field `ClientAuth` changed to `tls.RequireAndVerifyClientCert`. As long as the tls.Config is appropriately setup with the root CA and the client is providing a client certificate chain of intermediates, this will allow go's TLS package to handle verifying certs for us.

plorenz:
Discussed with @andrewpmartinez and given the complexity that ALPN adds, adding a post check seems reasonable.

PR #2592, opened 2024-12-13
fix.2591.split.apis.cause.login.fail

andrewpmartinez:


plorenz (approved):

controller/webapis/zac.go
```diff
@@ -0,0 +1,69 @@
+package webapis
```
dovholuknf, line 1: can you add the copyright i apparently forgot to add? :)

dovholuknf (approved):

PR #2618, opened 2025-01-06
set Linux package architecture for 32-bit ARM

qrkourier:
Set Linux package metadata to reflect the target architecture of the 32-bit ARM executable: ARM v7 w/ hardware floating point, aka "hard float."

- DEB targets `armhf`
- RPM targets `armv7hl`

### DIFF
--- a/.github/workflows/publish-linux-packages.yml
+++ b/.github/workflows/publish-linux-packages.yml
@@ -29,8 +29,8 @@ jobs:
             rpm: aarch64
           - goreleaser: armv7
             gox: arm
-            deb: armv7
-            rpm: armv7
+            deb: armhf
+            rpm: armv7hl
         packager:
           - rpm
           - deb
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-linux-packages.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       ZITI_MAINTAINER: "OpenZiti Maintainers <developers@openziti.org>"
       ZITI_HOMEPAGE: "https://openziti.io"
       ZITI_VENDOR: "NetFoundry Inc."
```
plorenz, line 43: TARGETARCH is a bool here, but a string below (I think). Should this be a an arch string instead of a bool?

qrkourier: It's always a string for the intents and purposes of this workflow. L44 tests if the packager for the current matrix iteration is `deb` and sets the string to what's expected by `dpkg` if so, and for `rpm` if not.

plorenz: I wasn't familiar with GH action expression language ternery operations. Makes more sense now :+1:

plorenz (approved):

PR #2645, opened 2025-01-13
fixes #2644 fixes log output for secondary jwt on mismatched identity

### LINKED ISSUE #2644: enhance mismapped external identity logging
I misconfigured my overlay today. I had an identity enrolled via OTT with secondary auth needed via a different auth policy than the default.

After successfully completing the pkce flow and submitting the jwt, the controller informed me i did not authenticate properly because the wrong identity was mapped to the external id.

Enhance the logging to indicate the identity discovered by external id so that I would have known I mismapped it

andrewpmartinez:
enhances logging and checks for scenarios where the wrong identity was configued with the matching external id


plorenz (approved):

controller/model/authenticator_mod_ext_jwt.go
```diff
@@ -703,6 +709,11 @@ func (a *AuthModuleExtJwt) verifyCandidate(context AuthContext, isPrimary bool,
 		return result
 	}
 
+	if targetIdentity != nil && targetIdentity.Id != identity.Id {
+		result.Error = fmt.Errorf("jwt mapped to identity [%s - %s], which does not match the current sessions identity [%s - %s]", identity.Id, identity.Name, targetIdentity.Id, targetIdentity.Name)
```
ekoby, line 713: is this the error that will be send back to SDK?

PR #2646, opened 2025-01-14
Add event source id to all events. Fixes #2619

### LINKED ISSUE #2619: Add source id to all events
All events should have a field with the id of the controller that is emitting the event.

plorenz:


controller/events/dispatcher_metrics.go
```diff
@@ -134,8 +134,9 @@ func (self *Dispatcher) unregisterMetricsEventHandler(val interface{}) {
 
 func (self *Dispatcher) newMetricEvent(msg *metrics_pb.MetricsMessage, metricType string, name string, id string) *event.MetricsEvent {
 	result := &event.MetricsEvent{
-		MetricType:    metricType,
 		Namespace:     event.MetricsEventsNs,
+		EventSrcId:    self.network.GetAppId(),
```
andrewpmartinez, line 138: And now "app id" == event source id == index id?

controller/event/terminators.go
```diff
@@ -36,6 +36,7 @@ const (
 type TerminatorEvent struct {
 	Namespace                 string              `json:"namespace"`
 	EventType                 TerminatorEventType `json:"event_type"`
+	EventSrcId                string              `json:"event_src_id"`
```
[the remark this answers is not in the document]
plorenz: event source id is the ID of the controller generating the event

controller/events/dispatcher_metrics.go
plorenz: app id == event source id == the id of the controller. They are distinct from the raft/datastore index.

andrewpmartinez: OH. I see. I was drawing on my event sourcing background which has an "event source id" which is the "index" in those systems. But this is literally the "id of the source" not the "id of the event sourced event"

andrewpmartinez (approved):

PR #2655, opened 2025-01-16
adds support for ${TMP} ${TEMP} and ${TMPDIR} in configs

andrewpmartinez:
- TMP/TEMP/TMPDIR are standards from windows, macOs, bsd, and linux
- TMP/TEMP/TMPDIR are set for all oses to allow cross platform configs
- Priotity order for selection is TMP/TEMP/TMPDIR
- For windows if TMP/TEMP are not set C:\temp is the default

### DIFF
--- a/common/config/env.go
+++ b/common/config/env.go
@@ -19,30 +19,13 @@ package config
 import (
 	"github.com/michaelquigley/pfxlog"
 	"os"
-	"runtime"
 	"sync"
 )
 
 var ensureTmpDirEnvOnce sync.Once
 
 func EnsureTempDirEnv() {
-	defaultTemp := "/tmp"
-
-	if runtime.GOOS == "windows" {
-		defaultTemp = os.Getenv("TEMP")
-
-		if defaultTemp == "" {
-			defaultTemp = "C:\\temp"
-		}
-	}
-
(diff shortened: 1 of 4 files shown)


common/config/env.go
```diff
@@ -16,9 +16,56 @@
...
+	"github.com/michaelquigley/pfxlog"
+	"os"
+	"runtime"
+	"sync"
+)
+
+var ensureTmpDirEnvOnce sync.Once
+
+func EnsureTempDirEnv() {
+	defaultTemp := "/tmp"
```
dovholuknf, line 29: what's wrong with just `os.TempDir()`?

andrewpmartinez: I didn't even know that existed. Updating.

dovholuknf (approved):
i don't "get it" but it looks good to me :)

qrkourier:
This causes a surprising result: a specified TMPDIR is silently ignored. Each temp dir var should be respected if already defined.

This assumes the OS default according to stdlib is the desired temporary dir. Still, it's a common practice to assign a var like `TMPDIR` in the parent environment by way of specifying a temporary directory for a particular purpose, which will be silently ignored by `ziti` and the OS default will be used instead of the specified directory.

PR #2657, opened 2025-01-16
Add doc generator for events. Add event doc. Fixes #2593

### LINKED ISSUE #2593: Update event documentation with missing event types
Documentation should be generated from source, so it's easier to keep up to date.

The following event types are missing doc: 

* cluster
* connect
* entityChange
* sdk
* terminator

plorenz:


controller/events/dispatcher_api_session.go
```diff
@@ -150,3 +157,24 @@ func (self *apiSessionEventAdapter) IsWrapping(value event.ApiSessionEventHandle
...
+
+func (self *apiSessionEventOldNsAdapter) IsWrapping(value event.ApiSessionEventHandler) bool {
+	if self.wrapped == value {
+		return true
+	}
+	if w, ok := self.wrapped.(event.ApiSessionEventHandlerWrapper); ok {
+		return w.IsWrapping(value)
+	}
+	return false
+}
```
andrewpmartinez, line 180: Is this associated with some other issue? The only issue I see mentioned is to add doc, but this also seems to do more?

Did writing the doc reveal the gap between legacy API Sessions vs HA API Sessions?

plorenz: No, this is here to allow the namespace to match the configuration field. So if the config file has
` - edge.apiSessions`, then that'll be the NS in the event. If the config file has `- apiSession`, then that will the NS in the event.

PR #2670, opened 2025-01-22
Fix api session memory leak in router. Fixes #2669

### LINKED ISSUE #2669: Router api session tracker leaks memory.
Router api session sync tracker isn't cleared after sync is complete and flag is being checked incorrectly, so new api sessions continue to pile up in tracker.

plorenz:


dovholuknf (approved):
lgtm

PR #2673, opened 2025-01-23
Obtain mgmt client from login creds

dovholuknf:
allows a cli command to login inline with the command itself if necessary.

### DIFF
--- a/ziti/cmd/verify/ops_verify_traffic.go
+++ b/ziti/cmd/verify/ops_verify_traffic.go
@@ -96,18 +96,13 @@ func NewVerifyTraffic(out io.Writer, errOut io.Writer) *cobra.Command {
 			t.clientIdName = t.prefix + ".client"
 			t.bindSPName = t.prefix + ".bind"
 			t.dialSPName = t.prefix + ".dial"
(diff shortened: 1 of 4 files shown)


ziti/cmd/verify/ops_verify_traffic.go
```diff
@@ -346,7 +342,7 @@ func createService(client *rest_management_api_client.ZitiEdgeManagement, name s
 	serviceParams.SetTimeout(5 * time.Second)
 	resp, err := client.Service.CreateService(serviceParams, nil)
 	if resp == nil || err != nil {
```
plorenz, line 344: Nitpick: Usual form would be log.WithError(err).Fataf(...)

dovholuknf: thx. i'll look at that for the next rev

plorenz (approved):

qrkourier:
You don't need the mgmt API URL if you log in with an identity JSON file.

dovholuknf:
are you trying to imply there's a problem or just trying to educate me? i'm just not following what you're trying to say?

PR #2675, opened 2025-01-23
fixes #2674 default port and local address routing for oidc auth

### LINKED ISSUE #2674: 404 not found on well-known OIDC configuration with default ports/localhost
*Experienced*

The `/.well-known/openid-conifguration` url 404s with `edge-oidc` bound when `:443` is used as the address and callers do not explicitly set the port (i.e. use default port for https). This also occurs on localhost vs 127.0.0.1 vs ::1 addresses.

*Expected*

The default port works if not explicitly set by callers. Local addresses work regardless of the one specified.

*Reproduction*

1) Bind edge-oidc to port :443
2) Call w/o the port

andrewpmartinez:
- fixes :443 default port calls from clients (e.g. https and :443 left off)
- fixes localhost vs 127.0.0.1 vs ::1 bindings

### DIFF
--- a/controller/webapis/oidc-api.go
+++ b/controller/webapis/oidc-api.go
@@ -86,7 +86,7 @@ func (h OidcApiHandler) RootPath() string {
 }
 
 func (h OidcApiHandler) IsHandler(r *http.Request) bool {
-	return strings.HasPrefix(r.URL.Path, h.RootPath()) || strings.HasPrefix(r.URL.Path, oidc_auth.WellKnownOidcConfiguration)
+	return strings.HasPrefix(r.URL.Path, h.RootPath()) || r.URL.Path == oidc_auth.WellKnownOidcConfiguration
 }
 
 func (h OidcApiHandler) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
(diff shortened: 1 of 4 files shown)


plorenz (approved):

controller/webapis/oidc-api.go
```diff
@@ -86,7 +86,7 @@ func (h OidcApiHandler) RootPath() string {
 }
 
 func (h OidcApiHandler) IsHandler(r *http.Request) bool {
-	return strings.HasPrefix(r.URL.Path, h.RootPath()) || strings.HasPrefix(r.URL.Path, oidc_auth.WellKnownOidcConfiguration)
+	return strings.HasPrefix(r.URL.Path, h.RootPath()) || r.URL.Path == oidc_auth.WellKnownOidcConfiguration
```
dovholuknf, line 89: should this be and? what if the edge/v1/client|management is prefixed? should that be allowed?

andrewpmartinez: No.

This is checking if the path is `/oidc` or `/.well-known/opendid-configuration`

PR #2680, opened 2025-01-23
Changed log.fatal to panic to allow catching the error instead

jensalm:


ziti/cmd/ascode/exporter/exporter.go
```diff
@@ -119,7 +119,7 @@ func (exporter *Exporter) Init(out io.Writer) error {
 	var err error
 	exporter.client, err = exporter.loginOpts.NewMgmtClient()
 	if err != nil {
-		log.Fatal(err)
+		panic(err)
```
dovholuknf, line 122: the functions already indicate they return errors, probably even better to just return instead?

ziti/cmd/ascode/importer/importer.go
```diff
@@ -111,7 +90,7 @@ func NewImportCmd(out io.Writer, errOut io.Writer) *cobra.Command {
 	return cmd
 }
 
-func (importer *Importer) Init() {
+func (importer *Importer) Execute(input []string) (map[string]any, error) {
```
dovholuknf, line 93: i'd probably change this to take named args since it's a function. passing a slice is a bit strange imo

dovholuknf (approved):

PR #2686, opened 2025-01-23
fixes #2681 adds targetToken support for ext jwt signers

### LINKED ISSUE #2681: Support specifying which token to use on external jwt signers
For some IdPs (Google Public OIDC), their access tokens are not externally verifiable nor intended to be inspected. Therefore, the only token that can relied on is the ID token. In general, access tokens should be used, but when we cannot, sharing the ID token between Ziti SDK/Clients and Ziti controllers should be supported.

- Support selecting which token to use via the Management API on create/update/patch
- Support reporting which token to use via the Management API on list/detail
- Support reporting which token to use via the Client API on list/detail
- Token fields to be supported: ACCESS, ID
- Default existing External JWT signers to ACCESS

Additional Information:

Another option is making the OpenZiti controller a relying party, allowing the configuration of a client_id/client_secret. However, it involves maintaining and defending against replay and novel attacks from separate user scopes, which increases the controller's attack surface with little benefit over using code flow + PKCE on the user device.

andrewpmartinez:
- adds targetToken of values ACCESS, ID for management API CRUD
- adds targetToken to client API reads
- adds --target-token to external jwt signers CLI
- updates/adds tests


controller/db/external_jwt_signer_store.go
```diff
@@ -43,8 +43,11 @@ const (
 	FieldExternalJwtSignerAudience        = "audience"
 	FieldExternalJwtSignerClientId        = "clientId"
 	FieldExternalJwtSignerScopes          = "scopes"
+	FieldExternalJwtSignerTargetToken     = "targetToken"
 
 	DefaultClaimsProperty = "sub"
+
+	TargetTokenAccess = "ACCESS"
```
dovholuknf, line 50: No TargetTokenID ?

tests/external_jwt_signer_test.go
```diff
@@ -518,12 +605,13 @@ func Test_ExternalJWTSigner(t *testing.T) {
...
-				Kid:      S(uuid.New().String()),
-				Issuer:   S(uuid.NewString()),
-				Audience: S(uuid.NewString()),
+				CertPem:     &jwtSignerCertPem,
+				Enabled:     &jwtSignerEnabled,
+				Name:        &jwtSignerName,
+				Kid:         S(uuid.New().String()),
+				Issuer:      S(uuid.NewString()),
+				Audience:    S(uuid.NewString()),
+				TargetToken: ToPtr(rest_model.TargetTokenID),
```
dovholuknf, line 614: should there be a test when it's left empty or when it's an invalid value? i didn't see such a test but it's kinda hard to scan for a missing value

tests/external_jwt_signer_test.go
```diff
@@ -459,12 +466,91 @@ func Test_ExternalJWTSigner(t *testing.T) {
...
+		ctx.Req.Equal(http.StatusCreated, resp.StatusCode())
+
+		jwtSignerUpdate := &rest_model.ExternalJWTSignerUpdate{
+			CertPem:     &jwtSignerCertPemUpdated,
+			Enabled:     &jwtSignerEnabledUpdated,
+			Name:        &jwtSignerNameUpdated,
+			Kid:         S(uuid.NewString()),
+			Issuer:      S(uuid.NewString()),
+			Audience:    S(uuid.NewString()),
+			TargetToken: nil,
```
dovholuknf, line 553: oh i think i found the nil test

dovholuknf (approved):
lgtm

controller/db/external_jwt_signer_store.go
andrewpmartinez: Not needed at this level. This is the storage level and can't access values up the dep chain (i.e. in the rest definitions). This is the only default value necessary here.

PR #2699, opened 2025-01-27
reconcile ziti-tunnel entrypoint script with ziti-edge-tunnel's

qrkourier:
The main reason for this is to enable the ziti-tunnel image to enroll and run with only an enrollment token env var.


plorenz (approved):

qrkourier:
closing in favor of new PR because I can't push to this branch because it contains unsigned commits and appears to have unrelated history from default branch: https://github.com/openziti/ziti/pull/2808

PR #2702, opened 2025-01-27
Add verify ext jwt OIDC

### LINKED ISSUE #2701: help users debug ext-jwt-signers
right now, ext-jwt-signers are primarily used for OIDC. OIDC discovery endpoints are good, but looking at the token that comes back is often incredibly difficult making it very hard to get just right without a fair bit of effort.

Add a command to `ziti ops` to help test/verify an external-jwt-signer works with oidc.

Example usage might be something like:
```
ziti ops verify-ext-auth oidc keycloak-signer --controller-url localhost:1280 --username abc --password xyz
```

This would lookup the signer `keycloak-signer` and complete a PKCE flow from the configured IdP

```
Usage:
  ziti ops verify-ext-auth oidc [flags]

Available Commands:


Flags:
      --access-token          Display the full Access Token to the screen. Use caution.
  -s, --additional-scopes     List of additional scopes to add
      --controller-url        The url of the controller
      --id-token              Display the full ID Token to the screen. Use caution.
      --refresh-token         Display the full Refresh Token to the screen. Use caution.
```

dovholuknf:
closes #2701

Adds  `ziti ops verify-ext-auth oidc` to the ziti CLI to hopefully assist users when configuring/testing their IdP of choice


internal/rest/client/helper.go
```diff
@@ -0,0 +1,97 @@
...
+
+func NewClientApiClient() (*rest_client_api_client.ZitiEdgeClient, error) {
+	cachedCreds, _, loadErr := util.LoadRestClientConfig()
+	if loadErr != nil {
+		return nil, loadErr
+	}
+
+	cachedId := cachedCreds.EdgeIdentities[cachedCreds.Default] //only support default for now
+	if cachedId == nil {
+		return nil, errors.New("no identity found")
```
andrewpmartinez, line 44: "no identity found, please login to the controller" ?

internal/rest/client/helper.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/openziti/edge-api/rest_client_api_client/external_jwt_signer"
+	"github.com/openziti/edge-api/rest_model"
+	"github.com/openziti/edge-api/rest_util"
+	internalconsts "github.com/openziti/ziti/internal/rest/consts"
+	"github.com/openziti/ziti/ziti/util"
+	"net/http"
+	"os"
+)
+
+func NewClientApiClient() (*rest_client_api_client.ZitiEdgeClient, error) {
```
andrewpmartinez, line 36: I would be surprised if something like this already didn't exist in the CLI. I don't know for sure, but I sense this block keeps repeating.

dovholuknf: Usually all the ziti cli stuff is always using the management API, this is the first time i've personally seen/used the client api with the ziti CLI on purpose.

internal/rest/client/helper.go
dovholuknf: turns out this whole function could just go away... thx

andrewpmartinez (approved):

PR #2706, opened 2025-01-28
set appropriate filemodes on unwrapped identity files

qrkourier:
On `!windows` systems, it's currently necessary to correct the filemode on unwrapped files. This respects the filemode mask on POSIX systems with appropriate defaults and always sets the default mode on Windows: 

- owner: read+write on key, cert, ca
- group, others: read-only on cert, ca

### DIFF
--- a/ziti/cmd/file_mode_unix.go
+++ b/ziti/cmd/file_mode_unix.go
@@ -1,5 +1,22 @@
 //go:build !windows
 
+/*
+	Copyright NetFoundry Inc.
+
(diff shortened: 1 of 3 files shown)


ziti/cmd/file_mode_unix.go
```diff
@@ -0,0 +1,27 @@
+//go:build !windows
```
andrewpmartinez, line 1: missing standard hearder

andrewpmartinez (approved):
Just ensure new files have our NF header.

PR #2708, opened 2025-01-29
Ensure restores and migrations work properly with RDM. Add restore from db for HA Clusters. Fixes #2549. Fixes #2649. Fixes #2707

### LINKED ISSUE #2549: Handle Index Non HA to HA Transitions During Upgrades
When migrating to HA, the index source will change from an internal index to a raft managed index at a low value (probably 1). This should be handled so Edge Routers do not confuse the index types and discard events as "old" because their old non-raft index is higher.

### LINKED ISSUE #2649: Make restoring an HA cluster from a DB backup easier

### LINKED ISSUE #2707: Ensure database restores work with RDM enabled routers
If controller comes online with restored DB, routers using RDM may not notice and may remain out of sync.

plorenz:


controller/network/network.go
```diff
@@ -1291,91 +1278,89 @@ func (network *Network) SnapshotDatabaseToFile(path string) (string, error) {
...
 		return nil
 	}
 
+	//exitAfterRestore := true
+	//err = network.GetDb().View(func(tx *bbolt.Tx) error {
+	//	if currentIndex := db.LoadCurrentRaftIndex(tx); currentIndex == 0 {
+	//		exitAfterRestore = false
+	//	}
+	//	return nil
+	//})
```
andrewpmartinez, line 1313: Dead

controller/network/network.go
```diff
@@ -1291,91 +1278,89 @@ func (network *Network) SnapshotDatabaseToFile(path string) (string, error) {
...
+//	}
+//
+//	cmd := &command.SyncSnapshotCommand{
+//		TimelineId:   snapshotId,
+//		Snapshot:     buf.Bytes(),
+//		SnapshotSink: network.RestoreSnapshot,
+//	}
+//
+//	return network.Managers.Dispatcher.Dispatch(cmd)
+//}
```
andrewpmartinez, line 1362: Dead

andrewpmartinez (approved):

PR #2709, opened 2025-01-29
tunneler intercept mode on linux only

dovholuknf:
An effort to stop seeing the tunneling error on anything not linux since it won't work on windows/macos and this error isn't relevant on either OS

```
[   8.513]   ERROR ziti/tunnel/dns.NewDnsServer: system resolver test failed: failed to resolve ziti-tunnel.resolver.test: lookup ziti-tunnel.resolver.test: no such host

ziti-tunnel runs an internal DNS server which must be first in the host's
resolver configuration. On systems that use NetManager/dhclient, this can
be achieved by adding the following to /etc/dhcp/dhclient.conf:

    prepend domain-name-servers 127.0.0.1:53;
```


scareything (approved):
IMO it would be more direct to have `dns_linux.go` for the linux implementation and `dns.go` for the stub. Otherwise lgtm.

dovholuknf:
> IMO it would be more direct to have `dns_linux.go` for the linux implementation and `dns.go` for the stub. Otherwise lgtm.

I am to please! I'll rename 'em

PR #2710, opened 2025-01-29
refinements to original add-verify-ext-jwt-oidc

### LINKED ISSUE #2711: move verify-* commands to subcommands
ziti ops verify-* should be moved to `ziti ops verify {subcommand}`. it just looks cleaner and logically groups these "verify-related" command together

dovholuknf:
closes #2711 

slight refactorings to make the verify-ext-jwt command actually be `ext-jwt-signer`. This pattern is emerging too, I plan to change the verify commands all to subcommands


CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+# Release 1.3.4
+
+# What's New
+
+* Moved `ziti ops verify-network` to `ziti ops verify network`
+* Moved `ziti ops verify traffic` to `ziti ops verify traffic`
+* Added `ziti ops verify ext-jwt-signer oidc` to help users with configuring OIDC external auth
```
andrewpmartinez, line 7: This breaks backward compatibility. We usually care about that for other commands. These though? I don't feel strongly as they are debugging tools.

andrewpmartinez, line 7: This breaks backward compatibility. We usually care about that for other commands. These though? I don't feel strongly as they are debugging tools.

plorenz: We can move it.
We can leave stub commands that direct users to the new commands.
We can leave the old commands present, but hidden, with a deprecation warning.
I'm fine with any of those.

andrewpmartinez (approved):
Consider Paul's comment.

dovholuknf:
based on the still relative newness of verify-network/verify-traffic, i'm fine with just moving them and not providing any hints back. people use these commands highly infrequently (if at all). thx

PR #2723, opened 2025-01-31
attempts to probe advertise address on startup to ensure the SANS is correct

### LINKED ISSUE #2722: router: check advertised address on startup
it's insanely easy for users to start a router that won't work by having a mismatch between the config file advertise address and the pki being presented.

it'd be great if the router when it starts could attempt to probe it's own advertised addresses and **ensure** the advertised address returns a certificate that will be valid.

For example, if i start a router and with a pki using "my.router.local" in the SANS but set the advertise address of any of the bindings to "this.is.invalid", the router will start and advertise "this.is.invalid". If a client connects to this router, the certificate chain won't be able to be verified and a connection won't complete

dovholuknf:
fixes #2722

too many users continually change the advertise address without understanding they need to change the PKI


andrewpmartinez:
I would expect this to stop the router. Also we don't need to probe TLS, we have all the certificate information internally. We can just look at those on startup.

router/router.go
```diff
@@ -181,6 +185,78 @@ func (self *Router) GetConnectEventsConfig() *env.ConnectEventsConfig {
...
+			continue
+		}
+
+		found := false
+		for _, s := range sans {
+			if s == host {
+				found = true
+				break
+			}
+		}
```
andrewpmartinez, line 248: This doesn't correctly match all variants of SAN values.

DNS
- case insensitivity
- trailing dog ignore
- IDN Punycode conversions
IP
- ipv4 normalization
- ipv6 normalization

andrewpmartinez (dismissed):
See comment.

router/router.go
andrewpmartinez: If you have an `x509.Ceritificate`, it has a function called VerifyHostname(hostname string) that should handle all of this for you. I took a cursory look, it seems like a good fit. It uses the ip.Equal which handles IP normalization. DNS matching looks "fine" but has some caveats.

ziti/router/run.go
```diff
@@ -114,7 +114,7 @@ func run(cmd *cobra.Command, args []string) {
 			pfxlog.Logger().WithError(err).Error("unable to start CLI agent")
 		}
 	}
-
+
```
andrewpmartinez, line 117: Impactful change.

dovholuknf: prolly too subtle... i changed it back

internal/tls/tls_util.go
```diff
@@ -0,0 +1,72 @@
+package tls
```
andrewpmartinez, line 1: Every function here takes in an identity as the first argument. It makes me feel like this should just be in identity, as these are validation functions useful for all kinds of servers.

tests/internal/tls/validFor_test.go
```diff
@@ -0,0 +1,162 @@
...
+func (m *mockIdentity) SetServerCert(pem string) error { return nil }
+func (m *mockIdentity) GetConfig() *identity.Config    { return nil }
+func (m *mockIdentity) ValidFor(address string) bool   { return true }
+
+const (
+	validDNS   = "example.com"
+	invalidDNS = "invalid.com"
+	validIP4   = "192.168.1.1"
+	invalidIP4 = "10.0.0.1"
+	validIP6   = "::1"
```
andrewpmartinez, line 39: I would expect to see an ipv6 compress vs expanded test. These two ipv6 addresses are equivalent:

- 2001:0db8:0000:0000:0000:ff00:0042:8329
- 2001:db8::ff00:42:8329

tests/internal/tls/validFor_test.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func TestValidFor_InvalidIPv4(t *testing.T) {
+	id := createmockIdentity([]string{}, []string{validIP4})
+
+	err := ziti_tls.ValidFor(id, invalidIP4+":"+validPort)
+	if err == nil {
+		t.Errorf("Expected error for invalid IP, got nil")
+	}
+	assert.Equal(t, "identity is not valid for provided host: ["+invalidIP4+"]. is valid for: ["+validIP4+"]", err.Error())
```
andrewpmartinez, line 102: We have the `require` library in the project to make test asserts more compact and expressive.

```
req := require.New(t)
...
req.NoError(err)
req.NoError(err, "expected valid hostname got error, %v", err))
req.Equal(1,2)
req.Equal(1,value, "value was expected to be 1, got %v", value)
```

Making you aware of it because I don't know if you have written tests in this project before.

internal/tls/tls_util.go
andrewpmartinez: Additionally, this package name collides with/ the base `tls` package. This means it will most likely need to be aliased if used (like in your test file imports); another good reason to move this to identity as identity interface methods.

tests/internal/tls/validFor_test.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func TestValidFor_NoCerts(t *testing.T) {
+	id := createmockIdentity([]string{}, []string{})
+
+	err := ziti_tls.ValidFor(id, validDNS+":"+validPort)
+	if err == nil {
+		t.Errorf("Expected error for no valid certs, got nil")
+	}
+	assert.Equal(t, "identity is not valid for provided host: ["+validDNS+"]. is valid for: []", err.Error())
```
andrewpmartinez, line 161: I am generally not a big fan of testing error strings. 

The alternative is to create a custom error type, implement the Error interface, write a test for the Error interface, and then inspect your returned error for the expected error type with specific fields inside of it.

Providing a thought, not asking for a change.

internal/tls/tls_util.go
dovholuknf: > Every function here takes in an identity as the first argument. It makes me feel like this should just be in the identity repo as methods on the identity interface, as these are validation functions useful for all kinds of servers.

moved to https://github.com/openziti/identity/pull/60

tests/internal/tls/validFor_test.go
dovholuknf: added these as tests. see https://github.com/openziti/identity/pull/60/files#diff-1c7a84c01de0f235cdda7cd9434265796e328b9b335eaf138a96a9df29c78769R163-R197

tests/internal/tls/validFor_test.go
dovholuknf: i'll adapt the tests over there

dovholuknf:
> I would expect this to stop the router from starting. Also we don't need to probe TLS, we have all the certificate information internally. We can just look at those on startup.

it'll fail now if any advertise address isn't valid per `VerifyHostname`

router/router.go
dovholuknf: used it. thx

dovholuknf:
force pushed a new clean changeset to this branch. old comments are addressed but this should be re-reviewed. i didn't know a better place to check for bindPoints

plorenz (approved):

controller/config/config.go
```diff
@@ -728,6 +744,46 @@ func LoadConfig(path string) (*Config, error) {
 
 	return controllerConfig, nil
 }
+func validateBindPoints(m map[interface{}]interface{}) []error {
```
andrewpmartinez, line 747: Is there a reason you didn't just ....use the parsed xweb during controller startup and do your validation there?

controller/controller.go line 232?

You have the parsed xweb instance and the controller identity.

andrewpmartinez:
```

	c.xweb = xweb.NewDefaultInstance(c.xwebFactoryRegistry, c.config.Id)
	
	xcfg := c.xweb.GetConfig()
	for _, svcConfig := range xcfg.ServerConfigs {
		for _, bindPoint := range svcConfig.BindPoints {
			svcCertChains := c.config.Id.VerifyFor(bindPoint.Address)
		}
	}
``` 

?

andrewpmartinez: Mmm actually this is self-contained within xweb's config section. It can just go in xweb.

dovholuknf: just didn't know any better way.  i've moved it to xweb anyway and will push shortly

controller/config/config.go
```diff
@@ -585,6 +585,14 @@ func LoadConfig(path string) (*Config, error) {
 						return nil, fmt.Errorf("error loading channel options for [ctrl/options] (%v)", err)
 					}
 				}
+				if value != nil {
+					m := value.(map[interface{}]interface{})
+					a := strings.TrimPrefix(m["advertiseAddress"].(string), "tls:")
+					v := controllerConfig.Id.ValidFor(strings.Split(a, ":")[0])
```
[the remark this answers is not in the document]
dovholuknf: what about routers connecting? those aren't xweb, right?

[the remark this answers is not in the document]
dovholuknf: but - routers/controllers are always specified outside of this config file? I just didn't know (still don't) if this address is used outside of xweb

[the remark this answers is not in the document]
andrewpmartinez: This section is only for controllers xweb powered APIs. So it won't affect anything else.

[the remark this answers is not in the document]
andrewpmartinez: I see your point about the ctrl listener using the root identity though. Which is a reasonable check.

andrewpmartinez (approved):

router/config.go
```diff
@@ -880,6 +880,42 @@ func LoadConfigWithOptions(path string, loadIdentity bool) (*Config, error) {
 		return nil, err
 	}
 
+	if loadIdentity {
+		var errs []error
+		for _, c := range cfg.Link.Listeners {
+			a := c["advertise"]
+			if a != nil {
+				// should start with tls:
```
qrkourier, line 888: "Should" or "can?" My naive observation has been that the `tls:` prefix is implied if omitted, and that we don't currently support any alternatives, so there's not currently any ambiguity, but there will be ambiguity in the future if we support another on-the-wire protocol, e.g., dtls.

PR #2727, opened 2025-02-03
Changing export/import structs to allow running it without using a Co…

jensalm:
…bra command


ziti/cmd/ascode/exporter/exporter.go
```diff
@@ -38,24 +38,29 @@ import (
...
 
 type Exporter struct {
-	loginOpts        edge.LoginOptions
-	client           *rest_management_api_client.ZitiEdgeManagement
-	ofJson           bool
-	ofYaml           bool
+	Out              io.Writer
+	Err              io.Writer
+	LoginOpts        edge.LoginOptions
+	OfJson           bool
```
dovholuknf, line 44: OfJson and OfYaml don't make sense to me. How about OutputType and take json or yaml with yaml being the default and eliminate one of the flags?

ziti/cmd/ascode/exporter/exporter.go
```diff
@@ -86,46 +91,43 @@ func NewExportCmd(out io.Writer, errOut io.Writer) *cobra.Command {
...
 	var err error
-	exporter.client, err = exporter.loginOpts.NewMgmtClient()
+	exporter.client, err = exporter.LoginOpts.NewMgmtClient()
 	if err != nil {
 		return err
 	}
 
-	if exporter.filename != "" {
-		o, err := NewOutputToFile(exporter.loginOpts.Verbose, exporter.ofJson, exporter.ofYaml, exporter.filename, exporter.loginOpts.Err)
+	if exporter.Filename != "" {
```
dovholuknf, line 121: with the change to allow supplying an output stream, it feels like the output file/input file stuff could be refactored to be relevant to the command exclusively, not to the exporter/importers

ziti/cmd/ascode/importer/importer_auth_policies.go
```diff
@@ -117,7 +117,7 @@ func (importer *Importer) ProcessAuthPolicies(input map[string][]interface{}) (m
 			}
 		}
 
-		if importer.loginOpts.Verbose {
+		if importer.LoginOpts.Verbose {
```
dovholuknf, line 120: i think this should just be a call to Debug not Info and doesn't need to bre wrapped in the if statement.

ziti/cmd/ascode/exporter/exporter.go
```diff
@@ -86,52 +163,27 @@ func NewExportCmd(out io.Writer, errOut io.Writer) *cobra.Command {
...
 	return cmd
 }
 
-func (exporter *Exporter) Execute(input []string) error {
+func (exporter *Exporter) Execute(client *rest_management_api_client.ZitiEdgeManagement, input []string) (map[string]interface{}, error) {
 
 	logLvl := logrus.InfoLevel
-	if exporter.loginOpts.Verbose {
+	if exporter.verbose {
 		logLvl = logrus.DebugLevel
```
dovholuknf, line 181: it doesn't make sense to change the log level here. it should be established already at this point

ziti/cmd/ascode/exporter/exporter.go
```diff
@@ -86,52 +163,27 @@ func NewExportCmd(out io.Writer, errOut io.Writer) *cobra.Command {
...
 	}
 
 	pfxlog.GlobalInit(logLvl, pfxlog.DefaultOptions().Color())
```
dovholuknf, line 184: same for the log setup. this is all stuff that should happen in the cmd or in some 'main' or... really anywhere but here

dovholuknf (approved):
lgtm

PR #2732, opened 2025-02-03
Endpoint management updates

### LINKED ISSUE #2108: Add `ctrls` property to non-ha router enrollment

### LINKED ISSUE #2724: Allow configuring endpoints file full path instead of directory

### LINKED ISSUE #2728: Write initial router endpoints file based on ctrls in JWT
Also need to figure out why all controllers aren't currently listed in JWT ctrls section

### LINKED ISSUE #2729: Enrollment doesn't contain controller which created the enrollment
It's only putting in 'peers' which don't contain self (or even all other controllers necessarily). Should use controller manager instead.

plorenz:
* Allow configuring endpoints file full path instead of directory. Fixes #2724
* Write initial router endpoints file based on ctrls in JWT. Fixes #2728
* Add ctrls property to non-ha router enrollment. Fixes #2108
* Enrollment doesn't contain controller which created the enrollment. Fixes #2729


router/configedge.go
ekoby, line 1: filename looks weird, maybe `edge_config.go` to match the type in it?

ekoby (approved):

router/configedge.go
plorenz: I had config_edge, but I think that matched some processor type, because it became invisible. Apparently the go recommendation is flat case, not snake case, so I went with configedge, so that it was together with config.

PR #2747, opened 2025-02-05
Fix: uint16 ports to cover the whole range

mratsim:
Previous int16 ports allowed setting ports up to 32768 only but some situations require higher range.

### DIFF
--- a/ziti/cmd/edge/quickstart.go
+++ b/ziti/cmd/edge/quickstart.go
@@ -58,9 +58,9 @@ type QuickstartOpts struct {
 	AlreadyInitialized bool
 	Home               string
 	ControllerAddress  string
-	ControllerPort     int16
+	ControllerPort     uint16
 	RouterAddress      string
-	RouterPort         int16
+	RouterPort         uint16
 	out                io.Writer
 	errOut             io.Writer
 	cleanOnExit        bool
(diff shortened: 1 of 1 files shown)


dovholuknf:
Thanks for the PR. OpenZiti requires signed commits in order to accept a PR. Would you be able to setup commit signing and amend your commit?

dovholuknf:
Also, would you mind submitting the CLA if you have not? https://openziti.io/policies/CONTRIBUTING.html#submitting-the-cla

dovholuknf:
Thanks for the cla and for signing your commit

dovholuknf (approved):
Looks good. Thanks again. I certainly intended to support uint16 😃

mratsim:
You're welcome.

Regarding the failing tests:
- Fablab smoketest and HA smoketest are failing due to missing test-report.xml. Unsure of the root cause.
image
- mattermost-ziti-webhook is failing due to an invalid config in ziti-sdk-c (called through ziti-sdk-nodejs, see https://github.com/search?q=repo%3Aopenziti%2Fziti-sdk-c%20INVALID_CONFIG&type=code) 
image
  However I can't find zid.json in those repos so I assume it's a Github secret injected in the workflow through `ziti-id: ${{ secrets.ZITI_MATTERMOST_IDENTITY }}`. And the ziti-sdk-c line number doesn't match with latest `main` and unsure where the version is pinned. So I'm unsure if my PR broke the C SDK. And the `parse_ziti_config` code seems to not come from `ziti-sdk-c`

dovholuknf:
no worries. the test has been finicky of late. i merged the pr. thanks again

PR #2758, opened 2025-02-07
Allow filtering on posture check type. Fixes #2668

### LINKED ISSUE #2668: [Feature Request] Filterable field for posture check type
Hello there!

Been working heavily with an API of Ziti Edge and found out that its impossible to filter posture checks based on their types.
Here is a piece of JSON response from `/posture-checks` endpoint.

```json
"meta": {
        "filterableFields": [
            "promptOnUnlock",
            "promptOnWake",
            "id",
            "createdAt",
            "updatedAt",
            "tags",
            "isSystem",
            "name"
        ],
}
````

It would be nice to have a way to filter out the posture checks of a certain type upon a request. Eg "type" as filterable would suffice as `/configs` endpoint has one.

plorenz:


dovholuknf:
this is more like the "refactor the pki command" and oh also fix posture check filtering. it was jarring at first to see so many unrelated commits mushed in here

dovholuknf (approved):

PR #2763, opened 2025-02-07
Update ubuntu version as 20.04 is being deprecated. Fixes #2761

### LINKED ISSUE #2761: Update GH runnings to use ubuntu-latest or ubuntu-22.04
ubuntu-20.04 is being deprecated

plorenz:


dovholuknf (approved):

qrkourier:
Focal has glibc 2.31. Here's a few with older glibc that won't be able to execute ziti.

### Distributions with glibc < 2.31

- **RedHat Linux 8**: glibc 2.28
- **Oracle Linux 8**: glibc 2.28
- **CentOS 7**: glibc 2.17
- **Oracle Linux 7**: glibc 2.17
- **Ubuntu Bionic**: glibc 2.27-3ubuntu1.6
- **Debian Buster**: glibc 2.28-10+deb10u2

.github/workflows/release.yml
```diff
@@ -82,7 +82,7 @@ jobs:
 
   linux-build:
     name: Build Linux binaries
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
qrkourier, line 85: Should we run tests on the oldest or newest available runner? I have a preference for pinning the runner, whichever version we decide to use.

.github/workflows/validation-links.yml
```diff
@@ -37,6 +37,11 @@ jobs:
       - name: Install Ziti CI
         uses: openziti/ziti-ci@v1
 
+      - name: Install Terraform
+        uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_version: "1.5.5"
```
qrkourier, line 43: Strong preference for pinning patch version or ~1.5 (latest 1.5.x) is good?

qrkourier (commented):
adding a builder layer in https://github.com/openziti/ziti/pull/2765

.github/workflows/main.yml
```diff
@@ -99,7 +99,7 @@ jobs:
 
   linux-build:
     name: Build Linux binaries
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
```
scareything, line 102: Why use 24.04 here, but 22.04 in the release workflow? imo it's best to build on the oldest kernel that's feasible. the resulting binaries will depend on the kernel's userspace abi, and care is taken to ensure that newer kernels support the abis from older kernels.

If the kernels used for the different workflows are intentionally different then it's probably worth commenting (in each workflow) why this is the case.

qrkourier:
Replaced by https://github.com/openziti/ziti/pull/2765

PR #2765, opened 2025-02-07
build ziti on the ziti-builder container image

qrkourier:


.github/workflows/main.yml
```diff
@@ -99,7 +99,8 @@ jobs:
 
   linux-build:
     name: Build Linux binaries
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
+    container: openziti/ziti-builder:v2
```
scareything, line 103: fyi using `steps`.`container` to run the build will cause the build actions to use the node environment from the container. this was problematic in the ziti-tunnel-sdk-c repo (due to node version deprecations in GH actions), which is why we ended up launching the containers from a build action via [`runs`.`using`.`docker`](https://github.com/openziti/ziti-tunnel-sdk-c/blob/main/.github/actions/openziti-tunnel-build-action/action.yml#L11) in ziti-tunnel-sdk-c.

.github/workflows/codeql.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
   analyze:
     name: Analyze
```
plorenz, line 28: Any reason not to keep the ubuntu-latest?

qrkourier: Not married to it, just a slight preference for pinning the runner for the sake of transparency about the relationship between the runner version and check outcomes. 

That is, when the latest runner inevitably shifts to the next LTS, we may not notice if, by chance, there are no breaking changes, which isn't guaranteed.

If we pin the runner, we'll be forced to opt-in to the change when our pinned runner is eventually sunset, and it will be more obvious that bumping the pinner runner version is a factor in any check outcomes at that time.

There's a chance we'll never need to do anything if we use latest, but if we do have to fix something it'll be harder to figure out due to the ambiguity about which underlying running was working and whether it changed.

.github/workflows/main.yml
qrkourier: Thank you for that memory jogger. Yes, the deprecated Node.js version was necessary for building on Ubuntu Bionic 18.04 because it provided a version of Node.js with a compatible GLIBC for that older distribution if I recall correctly.

I did make an error in this PR. I said in CHANGELOG.md that building on the ziti-builder:v2 image will produce a binary that requires a minimum GLIBC of 2.27 (Ubuntu Bionic 18.04), but in fact, the `:v2` image is based on Ubuntu Focal 20.04 and so the required minimum GLIBC version will not change with this PR since the `ziti` binary was previously built directly on the `ubuntu-focal` runner.

I've removed the erroneous changelog announcement about backward compatibility. I don't expect we'll encounter the same problems with Node.js in the ziti project because we're building on Focal with `:v2`, and the tunneler SDK is building on Bionic (`ziti-builder:v1`), which required the deprecated Node.js until we switched the compile step to a Docker Action to isolate it from the job runner environment.

plorenz (approved):

scareything (approved):

PR #2769, opened 2025-02-10
only publish release candidate if fablab succeeded or was skipped, 

qrkourier:
…but not if fablab failed

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -400,6 +400,7 @@ jobs:
       && (needs.mac-os-build.result == 'success' || needs.mac-os-build.result == 'skipped')
       && (needs.windows-build.result == 'success' || needs.windows-build.result == 'skipped')
       && (needs.fablab-smoketest.result == 'success' || needs.fablab-smoketest.result == 'skipped')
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -400,6 +400,7 @@ jobs:
       && (needs.mac-os-build.result == 'success' || needs.mac-os-build.result == 'skipped')
       && (needs.windows-build.result == 'success' || needs.windows-build.result == 'skipped')
       && (needs.fablab-smoketest.result == 'success' || needs.fablab-smoketest.result == 'skipped')
+      && (needs.fablab-ha-smoketest.result == 'success' || needs.fablab-ha-smoketest.result == 'skipped')
```
qrkourier, line 403: @plorenz This prevents the publish check from running, since it can only succeed once, until the fablabs succeed or are skipped, excluding failures. 

That's how the conditional is already written for the non-ha smoketest, but the ha-smoketest didn't get added until now, so whenever the ha-smoketest alone failed, the publish check would run prematurely.

plorenz (approved):

PR #2782, opened 2025-02-11
fixes #2737 open-id config issuer mismatch

### LINKED ISSUE #2737: OIDC issue mismatch with alt server certs
When using alt_server_certs on the OIDC API is it possible to have more issuers than the `bindPoint.address` indicates. The following situations can cause this:

- use of default TLS ports (explicit vs implicit)
- additional SANs in server_cert or alt_server_certs
- additional bind points with different ports
- use the root .well-known vs /oidc/.well-known

andrewpmartinez:
- scans all SANs for possible supported hostname/ips
- add copyright, add better accept handling
- re-arrange where issuers are determined
- removes support for root openid-config
- fixes issuer mapping
- fixes endpoint tests


controller/webapis/oidc-api_test.go
```diff
@@ -0,0 +1,248 @@
...
+		bindPoints := []*xweb.BindPointConfig{
+			{
+				Address: bindPoint1Address,
+			},
+			{
+				Address: bindPoint2Address,
+			},
+		}
+
+		issuers := getPossibleIssuers(id, bindPoints)
```
dovholuknf, line 143: probably should test a wildcard cert?

plorenz (approved):

controller/oidc_auth/negotiate.go
```diff
@@ -12,26 +30,47 @@ import (
...
 	}
 
 	return "", newNotAcceptableError(acceptHeader)
 }
 
+// parseContentType extracts the media type from a Content-Type header value
+func parseContentTypeMediaType(contentType string) string {
+	contentType = strings.TrimSpace(contentType)
+
+	if contentType == "" {
```
dovholuknf, line 59: this if doesn't seem necessary unless you're trying to optimize the function?

andrewpmartinez: Stylistic, it is a guard statement. Simply makes the condition explicit.

controller/webapis/oidc-api_test.go
andrewpmartinez: mmmm now that complicates things.

controller/oidc_auth/negotiate.go
```diff
@@ -49,17 +88,71 @@ func negotiateBodyContentType(r *http.Request) (string, *errorz.ApiError) {
...
+		parts := strings.Split(entry, ";")
+
+		mediaType := strings.TrimSpace(parts[0])
+		qValue := 1.0 // Default q-value
 
-	for i, part := range parts {
-		typeAndFactor := strings.Split(strings.TrimSpace(part), ";")
-		contentTypes[i] = typeAndFactor[0]
+		// Check if there's a q= parameter
+		if len(parts) > 1 {
```
dovholuknf, line 134: does this assume that if there's only a single `q param` that the value will always be 1 even if only "q=10" is supplied? why not just range over the parts looking for q=?

dovholuknf (approved):
unsure if wildcards need to be handled but what i see lgtm!

controller/webapis/oidc-api_test.go
andrewpmartinez: Pushed code. Tests still running.

controller/oidc_auth/negotiate.go
andrewpmartinez: Per the spec, there is one `q` per content type, and if it isn't specified,d it is defaulted to 1.

PR #2786, opened 2025-02-11
Mark peers as connected when accepting connections. Fixes #2785

### LINKED ISSUE #2785: Peer data model state not always updated
When a peer joins the mesh, it's often not marked as online. 

When the current leader accepts a peer connection instead of dialing, the certs are not set, short circuiting the update

plorenz:


controller/raft/mesh/mesh.go
```diff
@@ -817,6 +817,7 @@ func (self *impl) AcceptUnderlay(underlay channel.Underlay) error {
 		}
 
 		peer.Version = versionInfo
+		peer.SigningCerts = []*x509.Certificate{underlay.Certificates()[0]}
```
ekoby, line 820: why not take the full chain? is there a reason to take only leaf?

ekoby (approved):

controller/raft/mesh/mesh.go
plorenz: Question for @andrewpmartinez maybe. I'm just copying the behavior of the dialing side to the listening side, and didn't dig too deep into what we use those certs for. If no cert is present, we don't update the db, so that's the specific issue I was fixing.

PR #2792, opened 2025-02-12
fixes #2505 adds central configuration with events

### LINKED ISSUE #2505: Centralized Configuration

andrewpmartinez:
- uses dynamic JSON protobuf struct marshalling to avoid having to declare every setting in protobuf


controller/db/controller_setting_store.go
```diff
@@ -0,0 +1,375 @@
...
+		}
+
+		if store.settingControllerListeners[setting][id] == nil {
+			store.settingControllerListeners[setting][id] = make([]ControllerSettingListener, 0)
+		}
+
+		store.settingControllerListeners[setting][id] = append(store.settingControllerListeners[setting][id], listener)
+	}
+}
+
```
plorenz, line 204: Not clear why you're deleting the listener if the controller specific settings are removed. Isn't the listener still going to want to react to changes to the global config?

controller/db/controller_setting_store.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+	store.baseStore = newBaseStore[*ControllerSetting](stores, store)
+	store.InitImpl(store)
+	return store
+}
+
+type controllerSettingStoreImpl struct {
+	*baseStore[*ControllerSetting]
+	indexName                  boltz.ReadIndex
+	symbolEnrollments          boltz.EntitySetSymbol
```
plorenz, line 96: The map is likely being accessed/mutated across goroutines, should probably be protected

controller/env/appenv.go
```diff
@@ -407,6 +407,10 @@ func (ae *AppEnv) GetPeerSigners() []*x509.Certificate {
 	return ae.HostController.GetPeerSigners()
 }
```
plorenz, line 409: We already have GetId(). That's return the token id from the cert, AFAIK, so this should hopefully not be needed.

controller/model/controller_setting_manager.go
```diff
@@ -0,0 +1,204 @@
...
+		listener(settingPath, controllerSettingModel)
+	}, s.env.GetControllerId())
+}
+
+func (s *ControllerSettingManager) Marshall(entity *ControllerSetting) ([]byte, error) {
+	jsonData, err := json.Marshal(entity)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal entity: %w", err)
+	}
+
```
plorenz, line 135: I'm doing some json in protobuf as well for inspect stuff. I'll look at this and see if it makes sense to use there.

controller/model/controller_setting_model.go
```diff
@@ -0,0 +1,55 @@
+package model
+
+import (
+	"github.com/openziti/storage/boltz"
+	"github.com/openziti/ziti/controller/db"
+	"github.com/openziti/ziti/controller/models"
+	"go.etcd.io/bbolt"
+)
+
```
plorenz, line 9: Is TrustDomainSettings used?

plorenz (changes requested):
Overall looks good, couple of smaller things

PR #2815, opened 2025-02-19
Cache ER/T terminator ids for faster restarts. Fixes #2807

### LINKED ISSUE #2807: Cache ER/T terminator ids in the router for faster restarts
If we cache the terminator ids we can avoid deleting and recreating all the terminators.

plorenz:
Cache ER/T terminator ids for faster restarts. Fixes #2807


router/xgress_edge_tunnel_v2/tunneler.go
```diff
@@ -106,40 +112,18 @@ func (self *tunneler) Start(notifyClose <-chan struct{}) error {
...
-//				}
-//			}
-//		case <-notifyClose:
-//			return
-//		}
-//	}
-//}
+func (self *tunneler) WaitForInitialized() {
+	for !self.initialized.Load() {
+		time.Sleep(100 * time.Millisecond)
```
dovholuknf, line 124: i feel like this should have a timeout associated -- shouldn't it?

router/xgress_edge_tunnel_v2/hosted.go
```diff
@@ -115,6 +120,21 @@ func (self *hostedServiceRegistry) run() {
...
+	cache := self.env.GetRouterDataModel().GetTerminatorIdCache()
+
+	var toRemove []string
+	cache.IterCb(func(key string, v string) {
+		if !self.terminators.Has(v) {
+			toRemove = append(toRemove, key)
+		}
+	})
+
+	for _, key := range toRemove {
```
dovholuknf, line 133: dunno if it would be useful to have a log before/after this operation to have an indication of how long it takes?

router/xgress_edge_tunnel/factory_wrapper.go
```diff
@@ -172,11 +172,14 @@ func (self *FactoryWrapper) CreateListener(optionsData xgress.OptionsData) (xgre
 }
 
 func (self *FactoryWrapper) CreateDialer(optionsData xgress.OptionsData) (xgress.Dialer, error) {
-	if delegate := self.delegate.Load(); delegate != nil {
-		return delegate.CreateDialer(optionsData)
+	// wait till delegate is created. Once delegate is created, we should also be calling CreateDialer on
+	// the delegate directly, as the factory will get replaced in the registry
+	for {
```
dovholuknf, line 177: do we really want to potentially do this forever or is this another place where a timeout might be in order?

dovholuknf (approved):
coupla questions but lgtm

router/xgress_edge_tunnel_v2/hosted.go
plorenz: I would expect this to take milliseconds at most as the number of services should not be large and it's simple data structure operations

router/xgress_edge_tunnel/factory_wrapper.go
plorenz: I can't think of a case where this wouldn't exit within a reasonable time frame, but added a timeout just to be safe.

router/xgress_edge_tunnel_v2/tunneler.go
plorenz: I can't think of a case where this wouldn't exit within a reasonable time frame, but added a timeout just to be safe.

PR #2830, opened 2025-02-25
closes #2825 - use the client chain going forward

### LINKED ISSUE #2825: create config should reference client chain
In the clustered controller deployment domain, the client certificate for the controller is used to connect to other controllers. This connection should present a chain when connecting, currently it ends up configuring just the leaf.

dovholuknf:
closes #2825 

- use client chain for identity blocks, not the leaf. necessary for clustered controllers when connecting to other controllers
- get the quickstart test to successfully run again
- return errors instead of log.fatal for better test failures

### DIFF
--- a/ziti/cmd/create/create_config_controller.go
+++ b/ziti/cmd/create/create_config_controller.go
@@ -203,7 +203,7 @@ func SetControllerIdentity(data *ControllerTemplateValues) {
 func SetControllerIdentityCert(c *ControllerTemplateValues) {
 	val := os.Getenv(constants.PkiCtrlCertVarName)
 	if val == "" {
-		val = helpers.GetZitiHome() + "/" + helpers.HostnameOrNetworkName() + ".cert" // default
+		val = helpers.GetZitiHome() + "/" + helpers.HostnameOrNetworkName() + "client.chain.cert" // default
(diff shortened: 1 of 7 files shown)


ziti/cmd/create/create_config_controller.go
```diff
@@ -203,7 +203,7 @@ func SetControllerIdentity(data *ControllerTemplateValues) {
 func SetControllerIdentityCert(c *ControllerTemplateValues) {
 	val := os.Getenv(constants.PkiCtrlCertVarName)
 	if val == "" {
-		val = helpers.GetZitiHome() + "/" + helpers.HostnameOrNetworkName() + ".cert" // default
+		val = helpers.GetZitiHome() + "/" + helpers.HostnameOrNetworkName() + "client.chain.cert" // default
```
qrkourier, line 206: We should consistently name it like *.chain.cert or *.chain.pem. The `pki create client` command uses *.chain.pem to match the pre-existing `pki create server` which also uses *.chain.pem.

ref: https://github.com/openziti/ziti/blob/v1.3.3/ziti/cmd/pki/pki_create_client.go#L181

plorenz (approved):

PR #2848, opened 2025-02-26
don't fail startup on misconfigured healthcheck

### LINKED ISSUE #2845: 1.4.x Router - don't fail on invalid cert for heathchecks
With 1.4, [validation was added](https://github.com/openziti/xweb/issues/18) that causes controllers and routers to fail if the listening port is not configured with a cert valid for the advertised address. This is vital for controllers and routers normal functionality but it's not vital for healthchecks. (though it's a good idea of course)

Example:
```
web:
- name: health-check
  bindPoints:
    - interface: 0.0.0.0:8081
      address: 0.0.0.0:8081
  apis:
    - binding: health-checks
```

For now, only warn if the healthchecks are misconfigured, don't fail the router on startup.

At some point in the future healthchecks will also require a matching SANS

dovholuknf:
closes #2845 - only warn if the healthcheck address is not found in the configured certs

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -68,7 +68,6 @@ import (
 	"github.com/openziti/ziti/router/xlink"
 	"github.com/openziti/ziti/router/xlink_transport"
 	"github.com/pkg/errors"
-
 	"github.com/sirupsen/logrus"
 	"google.golang.org/protobuf/proto"
 	"gopkg.in/yaml.v3"
--- a/router/router.go
+++ b/router/router.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"github.com/openziti/foundation/v2/rate"
 	"github.com/openziti/ziti/common"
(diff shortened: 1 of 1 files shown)


router/router.go
```diff
@@ -67,7 +68,8 @@ import (
 	"github.com/openziti/ziti/router/xgress_transport_udp"
 	"github.com/openziti/ziti/router/xlink"
 	"github.com/openziti/ziti/router/xlink_transport"
-	"github.com/pkg/errors"
+	pkgerrors "github.com/pkg/errors"
```
andrewpmartinez, line 71: This dep should be removed, and we should rely on the "errors" package. Also, why was an alliase introduced?

`.Is` to `.As`  in the "errors' package.

andrewpmartinez (approved):

router/router.go
```diff
@@ -734,7 +736,19 @@ func (self *Router) initializeHealthChecks() (gosundheit.Health, error) {
 
 func (self *Router) RegisterXweb(x xweb.Instance) error {
 	if err := self.config.Configure(x); err != nil {
-		return err
+		actualErr := false
+
+		for je := err; je != nil; je = errors.Unwrap(err) {
```
andrewpmartinez, line 741: If you use the `errors` package. I'm pretty sure it handles this unwrapping for you.

router/router.go
andrewpmartinez: I see why, you have the "errors" package as well. I'd just remove this one.

router/router.go
dovholuknf: it did. i must have had the syntax wrong the first time i tried. i was "forced" to use the std erros but clearly it was just a  me problem. i treid it again and it was fine. happy to undo that! :) thx

PR #2883, opened 2025-03-11
fixes #2796 generated clients for enrollment do not work

### LINKED ISSUE #2796: Generated API client enrollment operations fail
as far as I can tell there are two operations that are defined for identity OTT enrollment:
- `enroll` (`/edge/client/v1/enroll?method=ott&token=<token>`)
   generated API client operation lacks parameter to pass CSR into the request

- `enrollOtt` (`/edge/client/v1/enrollOtt`) has `token`, and `clientCsr` parameters, however, when I try to use them:
    - golang client produces this error: `response status code does not match any response statuses defined for this endpoint in the swagger spec (status 400): {}`
    - generated java client get: `enrollOtt call failed with: 400 - {"error":{"code":"INVALID_ENROLLMENT_TOKEN","message":"The supplied token is not valid","requestId":"SgSmhtaOJ"},"meta":{"apiEnrollmentVersion":"0.0.1","apiVersion":"0.0.1"}}`

andrewpmartinez:
- default behavior complies with OpenAPI spec, but allows for legacy PEM handling through middleware intervention
- adds test for generic enrollment endpoint and for specific enrollment endpoints

### DIFF
--- a/controller/env/appenv.go
+++ b/controller/env/appenv.go
@@ -845,8 +845,7 @@ func NewAppEnv(host HostController) (*AppEnv, error) {
 
 	//enrollment consumer, leave content unread, allow modules to read
 	clientApi.ApplicationXPemFileConsumer = noOpConsumer
-	clientApi.ApplicationPkcs10Consumer = noOpConsumer
-	clientApi.ApplicationXPemFileProducer = &PemProducer{}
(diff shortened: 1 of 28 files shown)


plorenz (approved):
There are a few linter errors that need to be fix, but otherwise looks good.

PR #2900, opened 2025-03-18
Allow configuring size of the router metrics event queue. Fixes #2899

### LINKED ISSUE #2899: Allow configuring size of router metrics event queue size

plorenz:


CHANGELOG.md
```diff
@@ -47,11 +48,34 @@ is added to that terminator. This will bias future dials towards other terminato
 
 The failure cost can be reduced by successful dials. Failure cost is also reduced over time. In previous
 releases this was a fixed credit of 5, every minute. This is now changing to an exponential 
-amount, based on time since the last failure. The credit is now: `2 ^ (minutes since last failure/5)`.
+amount, based on time since the last failure. 
+
+If X is minutes sincew last failure, the credit is now: `min(10, 2 ^ (X/5))`.
```
dovholuknf, line 53: sincew typo

CHANGELOG.md
```diff
@@ -47,11 +48,34 @@ is added to that terminator. This will bias future dials towards other terminato
...
+
+## Router Metrics Change
+
+Router metrics has a new configuration setting:
+
+```
+metrics:
+  # Number of usage events to be able to queue. Defaults to 256. If this queue backs up, it can
```
dovholuknf, line 61: defaults to 256 or 1024? are these mismatched?

router/config.go
```diff
@@ -647,6 +648,8 @@ func LoadConfigWithOptions(path string, loadIdentity bool) (*Config, error) {
 
 	cfg.Metrics.ReportInterval = time.Minute
 	cfg.Metrics.MessageQueueSize = 10
+	cfg.Metrics.EventQueueSize = 256
```
dovholuknf, line 651: i'd really like to see these as actual defaults/consts somewhere. i know it's somewhat silly but it's nice to see them all in one place. also it'll be useful if we ever decide to add these to the configuration generator, which I'd really like to see us move to some day. maybe it's a nit but it'd be nice to have a list of these constants in one place imo

router/config.go
```diff
@@ -668,6 +671,13 @@ func LoadConfigWithOptions(path string, loadIdentity bool) (*Config, error) {
 					return nil, errors.New("invalid value for metrics.messageQueueSize")
 				}
 			}
+			if value, found := submap["eventQueueSize"]; found {
+				if intVal, ok := value.(int); ok {
+					cfg.Metrics.EventQueueSize = intVal
```
dovholuknf, line 676: makes me think the minimum size (16) actually belongs here instead of in the metrics library

dovholuknf (approved):

CHANGELOG.md
plorenz: Fixed

CHANGELOG.md
plorenz: Updated. I sometimes think that it doesn't make sense to show the setting as the default value, since you don't need it in that case, but I concede that possibility of confusion by having different numbers outweighs that minor consideration.

router/config.go
plorenz: I'll worry about it when we figure out how to make the config generator. It's very likely that we won't use constants for that anyway. It depends on what's easiest for a code generator to read.

dovholuknf: ? we have a config generator. i'd like to see us use it more

plorenz: Sorry, I was thinking config doc generator. AFAIK the config generator doesn't use any introspection. I'd like to have some sort of metadata that could be used for both, but I suspect that consts would be tricky to use in that way.

router/config.go
plorenz: Not going to worry about it for now. Setting it to a low value should have minimal effect on a system, and if someone goes to the effort to set it, they have a reason for it.

PR #2998, opened 2025-04-14
fixes #2997 adds authenticator ids for OIDC jwts

### LINKED ISSUE #2997: Add Authenticator Id to OIDC JWTs/return for current-api-session
OIDC JWTs do not contain the authenticator used (when applicable). SDKs use this contextual information to extend authenticators that are about to expire. Currently, the field is populated with the static value "oidc," which is not useful.

andrewpmartinez:
- updates current-api session logic to use new value
- sets z_authid in access/refresh tokens
- updates tests to ensure authenticator id is set on access/refres

### DIFF
--- a/tests/auth_oidc_test.go
+++ b/tests/auth_oidc_test.go
@@ -199,6 +199,24 @@ func Test_Authenticate_OIDC_Auth(t *testing.T) {
 		ctx.Req.NotEmpty(outTokens.IDTokenClaims)
 		ctx.Req.NotEmpty(outTokens.AccessToken)
 		ctx.Req.NotEmpty(outTokens.RefreshToken)
+
+		t.Run("access token has expected values", func(t *testing.T) {
+			ctx.testContextChanged(t)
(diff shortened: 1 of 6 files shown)


dovholuknf (approved):

tests/auth_oidc_test.go
```diff
@@ -199,6 +199,24 @@ func Test_Authenticate_OIDC_Auth(t *testing.T) {
 		ctx.Req.NotEmpty(outTokens.IDTokenClaims)
 		ctx.Req.NotEmpty(outTokens.AccessToken)
 		ctx.Req.NotEmpty(outTokens.RefreshToken)
+
+		t.Run("access token has expected values", func(t *testing.T) {
```
dovholuknf, line 203: all these tests have the same description. it **might** be nice to differentiate em?

andrewpmartinez: They are sub-runs so they show in the context of their parent test, both on the CLI and in IDE test runs.

PR #3012, opened 2025-04-27
feat: expose import / export features in the API

### LINKED ISSUE #3013: expose the import / export CLI commands in the Edge API
The Edge API should expose the import / export features previously implemented in CLI for use in ZAC, as an example.

Pehesi97:
This PR exposes the import / export CLI features in the edge management API for use by ZAC.

This PR depends on https://github.com/openziti/edge-api/pull/150.

Closes #3013.

### DIFF
--- a/controller/internal/routes/ascode_router.go
+++ b/controller/internal/routes/ascode_router.go
@@ -1,0 +1,181 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


andrewpmartinez:
Thank you very much for this PR.  We are not ready to add this functionality to the controller at this time, which is why the CLI does the heavy lifting for now. 

Additionally, it creates a dependency on the Edge API from the Fabric API. The Edge API is currently not required for some bare-bones configurations of the controller. The Edge and Fabric have been slowly merging, but at this point, this type of dependency isn't something we can accept.

PR #3044, opened 2025-05-09
add fips-ready ziti release

qrkourier:


.github/workflows/release.yml
```diff
@@ -127,9 +127,15 @@ jobs:
 
           $(go env GOPATH)/bin/gox -ldflags "$($(go env GOPATH)/bin/ziti-ci -q -t go-build-flags)" -cgo -os=linux -arch=amd64 -output=$GOX_OUTPUT ./ziti/
 
+          GOEXPERIMENT=boringcrypto \
+          $(go env GOPATH)/bin/gox -ldflags "$($(go env GOPATH)/bin/ziti-ci -q go-build-flags -n)" -cgo -os=linux -arch=amd64 -output="${GOX_OUTPUT}-boring" ./ziti/
+
           CC=arm-linux-gnueabihf-gcc \
           $(go env GOPATH)/bin/gox -ldflags "$($(go env GOPATH)/bin/ziti-ci -q -t go-build-flags)" -cgo -os=linux -arch=arm -output=$GOX_OUTPUT ./ziti/
 
+          GOEXPERIMENT=boringcrypto \
```
qrkourier, line 136: The arm64 build is sourced for arm64 container images. This enables the boring container images to remain deployable to the same target platform architectures as the non-boring images for Docker and Kubernetes, e.g., K3S on a RasPi would work!

.github/workflows/release.yml
```diff
@@ -164,6 +170,7 @@ jobs:
           go test -v -tags "quickstart automated" ./ziti/cmd/edge/...;
 
       - name: Run Unit and Integration Tests
+        if: ${{ vars.ZITI_SKIP_INTEGRATION_TESTS != 'true' }}
```
qrkourier, line 173: Let forks skip the tests, e.g., to test release CI, because the tests will never succeed outside the upstream repo. They require central authorization for something related to the import/export feature.

.github/workflows/publish-docker-images.yml
```diff
@@ -79,15 +78,47 @@ jobs:
           subject-digest: ${{ steps.push_cli.outputs.digest }}
           push-to-registry: true
 
+      - name: Set Up Container Image Tags for Boring CLI Container
+        env:
+          IMAGE_REPO: ${{ env.REGISTRY }}/${{ env.ZITI_CLI_IMAGE }}
+          IMAGE_TAG: ${{ env.ZITI_CLI_TAG }}-boring
```
dovholuknf, line 84: why all the `-boring` instead of `-fips`? i thought you were using `-fips`?

.github/workflows/release.yml
```diff
@@ -280,10 +287,16 @@ jobs:
 
           $(go env GOPATH)/bin/ziti-ci configure-git
 
+          # workaround: prune the boring builds from release tarballs so it will not force the regular ziti binary into a nested path /ziti/ziti
```
qrkourier, line 290: @plorenz I hesitate to change how ziti-ci archives artifacts: nest in `cmd.name` directory if multiple files, else top-level. I'm unsure what side effects changing that might have, and I remember it was problematic when we triggered that behavior by unintentionally including a helper binary in the Linux release archive.

.github/workflows/publish-docker-images.yml
qrkourier: It seems more accurate and descriptive of how this artifact differs. Either would work just fine, and now's the time to decide on a consistent representation. I vote for `openziti/ziti-controller:1.6.2-boring`, but I'll make the change to `openziti/ziti-controller:1.6.2-fips` if you feel strongly about it.

dovholuknf: it means people will have to change the image they pull if WE change from boring. which is why i don't like -boring.

PR #3050, opened 2025-05-20
Fix.3048.auth.events.impropert.chain.flag

andrewpmartinez:
- sets `improperClientCertChain` on current API Session for oidc and legacy auth
- emits events for auth success/failure (including the above flag)


common/pb/edge_cmd_pb/edge_cmd.proto
```diff
@@ -68,6 +68,9 @@ message Authenticator {
     bool isExtendRequested = 6;
     bool isKeyRollRequested = 7;
     google.protobuf.Timestamp extendRequestedAt = 8;
+    string publicKeyPrint = 9;
+    bool lastExtendPublicKeyChanged = 10;
```
dovholuknf, line 72: lastExtendPublicKeyChanged is different than extendRequestedAt, I presume? It'd be nice to keep those names consistent imo and keep the "At"... lastExtendPublicKeyChangedAt

controller/db/authenticator_store.go
```diff
@@ -29,12 +29,15 @@ const (
...
-	FieldAuthenticatorCertIsKeyRollRequested = "isKeyRollRequested"
-	FieldAuthenticatorCertExtendRequestedAt  = "extendRequestedAt"
+	FieldAuthenticatorCertFingerprint            = "certFingerprint"
+	FieldAuthenticatorCertPem                    = "certPem"
+	FieldAuthenticatorCertIsIssuedByNetwork      = "isIssuedByNetwork"
+	FieldAuthenticatorCertIsExtendRequested      = "isExtendRequested"
+	FieldAuthenticatorCertIsKeyRollRequested     = "isKeyRollRequested"
+	FieldAuthenticatorCertExtendRequestedAt      = "extendRequestedAt"
+	FieldAuthenticatorCertPublicKeyPrint         = "publicKeyPrint"
+	FieldAuthenticatorCertLastAuthResolvedToRoot = "lastAuthResolvedToRoot"
```
dovholuknf, line 39: lastExtendPublicKeyChanged vs lastExtendRolledKeys are the same right, i expect? why not keep these in alignment with the Authenticator for consistency? Either change there or here?

controller/event/authentication.go
```diff
@@ -0,0 +1,111 @@
...
+	AuthenticationEventNS = "authentication"
+
+	AuthenticationEventTypeFail    = "fail"
+	AuthenticationEventTypeSuccess = "success"
+)
+
+// An AuthenticationEvent is emitted when an authentication attempt is made
+//
+// Types of authentication events
+//   - failed - authentication failed
```
dovholuknf, line 34: failed or fail? is this supposed to be the same as what's on line 27?

controller/events/dispatcher_authentication.go
```diff
@@ -0,0 +1,112 @@
...
+		} else if includeIntfList, ok := includeVar.([]interface{}); ok {
+			for _, val := range includeIntfList {
+				includeList = append(includeList, fmt.Sprintf("%v", val))
+			}
+		} else {
+			return errors.Errorf("invalid type %v for %v include configuration", reflect.TypeOf(includeVar), event.AuthenticationEventNS)
+		}
+	}
+
+	if len(includeList) == 0 || (len(includeList) == 2 && stringz.ContainsAll(includeList, event.AuthenticationEventTypeSuccess, event.AuthenticationEventTypeFail)) {
```
dovholuknf, line 69: > len(includeList) == 2

seems very specific and it's not clear why two. Maybe this is a known pattern but it seems... odd to me? this sort of logic needs a comment explaining it imo. Is this somehow obvious with context?

controller/model/authenticator_manager.go
```diff
@@ -599,6 +601,19 @@ func (self *AuthenticatorManager) VerifyExtendCertForIdentity(isJwtBacked bool,
...
+
+	if authenticatorCert.PublicKeyPrint == "" {
+		//if we don't have a saved value, parse and calculate
+		certs := nfpem.PemStringToCertificates(authenticatorCert.Pem)
+		if len(certs) > 0 {
+			authenticatorCert.PublicKeyPrint = PublicKeySha256(certs[0])
+		}
+	}
+
+	authenticatorCert.LastExtendRolledKeys = newPublicKeyPrint != authenticatorCert.PublicKeyPrint
```
dovholuknf, line 614: PublicKeySha256 can return "" when it fails. In the unlikely case that line 604 or 610 fail this could be true for the wrong reason, right?

dovholuknf (approved):

controller/db/authenticator_store.go
andrewpmartinez: No. You can extend with the same public/private key pair.

common/pb/edge_cmd_pb/edge_cmd.proto
andrewpmartinez: One is a bool, the other is a timestamp. They aren't the same?

controller/event/authentication.go
andrewpmartinez: Ah yea, I chagned the value w/o changing the doc.

controller/events/dispatcher_authentication.go
andrewpmartinez: It was done in other event handlers. I just followed the pattern. It appears to be an optimization.

controller/model/authenticator_manager.go
andrewpmartinez: The error result is improbable. It is because PublicKey is of type  `any`  and can fail type asserting when marshalling. At this point, the cert has been parsed and used multiple times. So it is improbable that it will error.

PR #3069, opened 2025-05-29
fixes #3068 adds event dispatching for auto ca enrollment

### LINKED ISSUE #3068: Auto CA enrollments do not dispatch events in HA
*Experienced*

Identities created via OTTCA enrollment appear only on the controller they are created on.

*Expected*

They appear on every controller.

*Reproduction*

1) Add a 3rd party CA, validate, and enable auto enrollment
2) Issue a new client cert and enroll
3) Query each controller's identity list, note all but the one recieving the enrollment have the identity

andrewpmartinez:


controller/model/identity_manager.go
```diff
@@ -145,6 +146,74 @@ func (self *IdentityManager) ApplyCreateWithEnrollments(cmd *CreateIdentityWithE
 	})
 }
 
+func (self *IdentityManager) CreateWithAuthenticators(identity *Identity, authenticators []*Authenticator, ctx *change.Context) (string, []string, error) {
+	if identity.Id == "" {
```
dovholuknf, line 150: identity is a pointer. should a nil check come first or let it panic?

controller/model/enrollment_mod_ca.go
```diff
@@ -189,14 +189,14 @@ func (module *EnrollModuleCa) completeCertAuthenticatorEnrollment(log *logrus.En
 		},
 	}
 
-	_, authenticatorId, err := module.env.GetManagers().Identity.CreateWithAuthenticator(identity, newAuthenticator, context.GetChangeContext())
+	_, authenticatorIds, err := module.env.GetManagers().Identity.CreateWithAuthenticators(identity, []*Authenticator{newAuthenticator}, context.GetChangeContext())
 
 	if err != nil {
 		log.WithError(err).Error("failed to create identity with authenticator")
 		return nil, err
 	}
```
plorenz, line 198: Any chance of a panic here (can the authenticator list ever be empty)?

andrewpmartinez: Technically, that is fine. It is just a create at that point.

controller/model/identity_manager.go
```diff
@@ -145,6 +146,74 @@ func (self *IdentityManager) ApplyCreateWithEnrollments(cmd *CreateIdentityWithE
...
+		identity.Id = eid.New()
+	}
+
+	var authenticatorIds []string
+	for _, authenticator := range authenticators {
```
dovholuknf, line 155: does this mean the callee can send in authenticators that have id's already? is that a common pattern? seems odd to me to allow the callee to specify the id?

controller/model/identity_manager.go
andrewpmartinez: Panic, this is internal.

PR #3077, opened 2025-06-02
MUSL getaddrinfo() compatibility

### LINKED ISSUE #3075: TPROXY mode in MUSL - error responses to IPv6 AAAA requests invalidate the IPv4 intercept answer
MUSL's `getaddrinfo()` requests both `A` (IPv4) and `AAAA` (IPv6) in parallel, and ignores the IPv4 intercept response if the nameserver responds to the IPv6 request that it won't answer on behalf of the requested domain (`REFUSED`) or indicates that domain does not exist (`NXDOMAIN`).

This differs from GLIBC's `getaddrinfo()`, which will honor the response to the `A` query if the `AAAA` query returns an error.

This causes all Alpine-based client application containers running in Docker and Kubernetes to fail to obtain the Ziti service intercept IPv4 address.

Possible solutions include changing the Ziti nameserver to respond to `AAAA` queries with `NOERROR` and an empty set of IPv6 addresses, which would cause MUSL to respect the `A` response containing the intercept address.

qrkourier:
resolves #3075

### DIFF
--- a/tunnel/dns/server.go
+++ b/tunnel/dns/server.go
@@ -154,7 +154,7 @@ func (r *resolver) ServeDNS(w dns.ResponseWriter, query *dns.Msg) {
 	case dns.TypeAAAA:
 		// Always respond with NOERROR for AAAA queries (success, but empty answer)
 		msg.Authoritative = true
-		msg.Rcode = dns.StringToRcode["NOERROR"]
+		msg.Rcode = dns.RcodeSuccess
 	}
 	log.Tracef("response:\n%s\n", msg.String())
 	err := w.WriteMsg(&msg)
--- a/tunnel/dns/server.go
+++ b/tunnel/dns/server.go
@@ -152,8 +152,7 @@ func (r *resolver) ServeDNS(w dns.ResponseWriter, query *dns.Msg) {
 			msg.Rcode = dns.RcodeRefused // fail fast, and inspire resolver to query next name server in its list.
 		}
 	case dns.TypeAAAA:
-		// Always respond with NOERROR for AAAA queries (success, but empty answer); enables MUSL getaddrinfo() to resolve the intercept domain to IPv4 address
-		log.Debugf("NOERROR for AAAA query: %s", q.Name)
+		// Always respond with NOERROR for AAAA queries (success, but empty answer)
 		msg.Authoritative = true
 		msg.Rcode = dns.StringToRcode["NOERROR"]
 	}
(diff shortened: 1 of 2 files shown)


tunnel/dns/server.go
```diff
@@ -151,6 +151,10 @@ func (r *resolver) ServeDNS(w dns.ResponseWriter, query *dns.Msg) {
 		} else {
 			msg.Rcode = dns.RcodeRefused // fail fast, and inspire resolver to query next name server in its list.
 		}
+	case dns.TypeAAAA:
+		// Always respond with NOERROR for AAAA queries (success, but empty answer)
+		msg.Authoritative = true
+		msg.Rcode = dns.StringToRcode["NOERROR"]
```
dovholuknf: this should be `msg.Rcode = dns.RcodeSuccess`

dovholuknf: seems like it would be better to follow what `A` does and use `RcodeRefused`

scareything: Have you verified that this will inspire the desired behavior on alpine, and that it doesn't break ziti dns on other resolvers? I'm concerned that it will cause our ipv4 answers to be ignored by resolvers that send dual queries and favor the ipv6 answer.

scareything: Would following the zet approach of not responding to `AAAA` queries  at all meet the need?

qrkourier: @dovholuknf wrote:
> this should be `msg.Rcode = dns.RcodeSuccess`

That is logically equivalent, since `NOERROR` also maps back to response code int 0. It would be slightly more efficient not looking up the symbolic form, at the possible expense of clear intent. Either is probably fine.

@dovholuknf wrote:
> seems like it would be better to follow what `A` does and use `RcodeRefused`

`RcodeRefused` is the current response this aims to change because responding non-zero is treated as an error and causes MUSL's `getaddrinfo()` to fail both IPv4 and IPv6 queries.

@scareything wrote:
> Have you verified that this will inspire the desired behavior on alpine, and that it doesn't break ziti dns on other resolvers?

It inspires the desired behavior on Alpine, i.e., the IPv4 intercept is no longer ignored when calling `getaddrinfo()`.  For non-Alpine, the empty answer for `AAAA` didn't break the caller's ability to obtain the IPv4 intercept or retrying the non-Ziti query on the next nameserver.

@scareything wrote:
> Would following the zet approach of not responding to `AAAA` queries at all meet the need?

I tried that at your suggestion. The `A` answer with the intercept was ignored on the Alpine-based container. I assume it means MUSL's `getaddrinfo()` treated the nameserver as unavailable because the `AAAA` query timed out. This is the same disposition as the `AAAA` query getting a `REFUSED` or `NXDOMAIN` response.

In contrast, the non-Alpine container allows the `AAAA` timeout, accepting the `A` answer.

scareything (approved):

PR #3097, opened 2025-06-13
Add support for tracking network interfaces on routers and identities. Fixes #3082. Add network discover to router. Fixes #3083

### LINKED ISSUE #3082: Add interfaces to controller data model
We want to be able to optionally collect information about host network interfaces from SDKs and routers. We need a place to store this information in the data model as well as APIs to retrieve it. 

 We need it on 

* identities
* routers
* routers with identities 
    * how do we want to handle this? Should the identity return the data from the router? Or do we not worry about it and let API clients pull from the right place based on the identity type

### LINKED ISSUE #3083: Add optional network interface discovery to routers
Routers should be able to discover the network interfaces on their host and inform the controller. Changes to network interfaces should also be reported to the controller as they happen (with some reasonable delay).

Notes:
1. See if we can use netlink to implement this
2. Feature should be optional, disabled by default

plorenz:


controller/db/interface_db_model.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+	}
+	return result
+}
+
+func storeInterfaces(interfaces []*Interface, ctx *boltz.PersistContext) {
+	if ctx.ProceedWithSet(FieldInterfaces) {
+		m := map[string]any{}
+		for idx, v := range interfaces {
+			m[fmt.Sprintf("%d.%s", idx, v.Name)] = v.toBoltMap()
```
andrewpmartinez, line 133: So this uses the report order + interface name as the unique key? If the OS reports the order differently, it would have to be an entirely new map with every update. 

It isn't an issue as long as we never directly reference the key values as foreign keys and only use this as information for a user to make other decisions. Otherwise, MAC address would probably have been better.

andrewpmartinez (approved):

controller/db/interface_db_model.go
plorenz: The list is always sorted by name, so the order should be consistent. interface names are supposed to be unique, so the index shouldn't even be required, but I added it just in case there was some oddity. The value is never exposed to the end user.

PR #3105, opened 2025-06-16
Add HA data flow test. Fixes #2862

### LINKED ISSUE #2862: Comprehensive HA test
Add a test which includes lots of links, terminators and circuits. Should have an existing test that we can base it off of, but need to add more automation.

plorenz:


zititest/models/circuit-perf/main.go
```diff
@@ -105,6 +105,21 @@ var Model = &model.Model{
...
+		model.FactoryFunc(func(m *model.Model) error {
+			simServices := zitilibOps.NewSimServices(func(s string) string {
+				return "component#" + s
+			})
+
+			m.AddActivationStageF(simServices.SetupSimControllerIdentity)
+			m.AddOperatingStage(simServices.CollectSimMetricStage("metrics"))
+
+			m.AddOperatingStageF(func(run model.Run) error {
+				time.Sleep(time.Hour * 24)
```
andrewpmartinez, line 117: Does this make the test run as a soak test for 24hr?

plorenz: I usually run the test as follows: `fablab exec-loop 24h testIteration` which does try to run it for 24 hours.

You can  do `fablab run`, which will spin it up in metrics relay mode, collecting metrics and pushing them to an influx instance. That's useful for running the test interactively and watching the metrics as the test runs. There's not really a reason to pin that for 24h, except you're unlikely to run it interactively for that long. It should probably just be a channel wait, so it's indefinite.

plorenz: I'll update it in one of the bugfix PRs

PR #3121, opened 2025-06-17
Using the same heartbeatmsg instance across channels causes data race. Fixes #3119

### LINKED ISSUE #3119: Using the same heartbeatmsg instance across channels causes data race

plorenz:


router/state/heartbeat.go
```diff
@@ -94,7 +94,10 @@ func (operation *heartbeatOperation) beat(tokens []string) {
 
 	operation.env.GetNetworkControllers().ForEach(func(ctrlId string, ch channel.Channel) {
 		for _, msg := range msgs {
-			if err := ch.Send(msg); err != nil {
+			// can't send the same messages across channels, will run into problems with channle heartbeats and
```
dovholuknf: typo: `channle`

router/state/heartbeat.go
```diff
@@ -94,7 +94,10 @@ func (operation *heartbeatOperation) beat(tokens []string) {
...
+			// concurrent writes
+			msgCopy := channel.NewMessage(msg.ContentType, msg.Body)
```
dovholuknf, line 99: there's no more efficient way of doing this? or just not worth the 'premature optimization'?

plorenz: The messages need to be per-channel, as sequence numbers are assigned by the channel and headers can also be affected. 
However, it's just the wrapper. The body can be shared since that is involatile.

router/state/heartbeat.go
plorenz: fixed

PR #3125, opened 2025-06-18
fix ha circuit id gen

### LINKED ISSUE #3124: ids used by circuits and ingress/egress can conflict in an HA setup
circuit/ingess/egress ids are only unique per controller. Switch to uuids to prevent conflicts.

plorenz:
- **ids used by circuits and ingress/egress can conflict in an HA setup. Fixes #3124**
- **Update deps and changelog**


controller/idgen/id.go
```diff
@@ -51,10 +51,20 @@ func (self *shortIdGenerator) NextId() (string, error) {
 	}
 }
 
-func NewUUIDString() string {
+func MustNewUUIDString() string {
```
andrewpmartinez, line 54: "MustNew" ? Does the panic ever happen? It seems weird to panic at all rather than pass back an error.

andrewpmartinez (approved):

controller/idgen/id.go
plorenz: I would guess that the error never happens, but before we were ignoring the error. I figured it was better to find out that something had gone wrong than not.

plorenz: I'll make a note to go back and switch all the uses over to the error returning one.

andrewpmartinez: If that were the case, I suggest making the original method panic and using that everywhere.

PR #3137, opened 2025-06-25
fix #3236 SDK/ENV info is not saved on OIDC auth

andrewpmartinez:


controller/oidc_auth/login.go
```diff
@@ -348,6 +349,12 @@ func (l *login) authenticate(w http.ResponseWriter, r *http.Request) {
 		return
 	}
```
plorenz, line 353: Does this need to go through the manager so it gets turned into a raft journal entry?

andrewpmartinez: The store here is the OIDC OP `store` which is defined in `storage.go` and does `return s.env.GetManagers().Identity.UpdateSdkEnvInfo(identity, envInfo, sdkInfo, changeCtx)`

The "store" naming here came from the interface defined by zitadel/oidc, and unfortunately, it conflicts with our internal "store" naming.

plorenz: ah, ok. Makes sense

plorenz (approved):

PR #3141, opened 2025-06-26
send repo dispatch event on release

qrkourier:


.github/workflows/release.yml
```diff
@@ -329,3 +328,30 @@ jobs:
...
+
+  repository-dispatch:
+    if: ${{ !cancelled() && needs.publish.result == 'success' }}
+    needs:
+      - publish
+      - call-publish-docker-images
+    name: Trigger Repository Dispatch
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Git Checkout
```
dovholuknf, line 340: is the checkout required?

qrkourier: Yes, it errored without the checkout.

qrkourier: I remembered incorrectly. The checkout isn't required for `gh api` because it's a repo-less operation, no `--repo` or Git checkout introspection.

PR #3154, opened 2025-07-07
Add convenience method for getting configs for identity

plorenz:


common/subscriber.go
```diff
@@ -55,6 +56,16 @@ func (self *IdentityService) IsEncryptionRequired() bool {
...
 }
 
+func (self *IdentityService) GetConfig(configTypeName string, v any) (bool, error) {
+	if config, ok := self.Configs[configTypeName]; ok {
+		if err := json.Unmarshal([]byte(config.Config.DataJson), &v); err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+	return false, nil
```
ekoby, line 66: feels like should return an error here

plorenz: The bool indicates whether or not the config is present. The error indicates if there as a problem loading the config

PR #3180, opened 2025-07-21
fix #3178 apis missing from controllers

### LINKED ISSUE #3178: Controller List Edge APIs Missing
*Experienced*

On networks where the leader is changed (removal, election, etc.), it is possible to miss controller API endpoint reports during connection. This causes the controller list at edge/[client|management]/v1/controllers to be incorrect. Causing some controllers to fall out of use by SDKs.

*Expected*

No lost controller list updates.

*Reproduction*

Hard, trigger hellos during an election or right before an election where no leader is present.

andrewpmartinez:
- fixed controller store to no longer blank apis on all CUD actions
- controllers now report their current config state on leader change to avoid stale information
- fixes peer disconnect to only set online state


ekoby (approved):
no tests?

andrewpmartinez:
> no tests?

We don't have a HA test harness for integration tests, nor do I know of a good way to write tests in fablab to startup/take down controllers over and over.

PR #3200, opened 2025-08-05
Ensure routers react to link listener address changes for other routers. Fixes #3199

### LINKED ISSUE #3199: Other routers don't react to link listener address changes
Other routers don't react to link listener address changes

plorenz:


router/link/link_events.go
```diff
@@ -121,6 +122,16 @@ func (self *linkDestUpdate) ApplyListenerChanges(registry *linkRegistryImpl, des
 					registry.evaluateLinkState(newLinkState)
 				} else {
 					log.Info("link already known")
+					if existingLinkState.listener.Address != listener.Address {
+						log.Info("link address changed, updating")
```
dovholuknf: imo, it would be best to see what it went from an to in this log message.

dovholuknf: also maybe the "link already known" log message should be in the else of this if

plorenz: updated log message. I think the 'link already known message' is consistent. It lets you know this isn't  new link, which is true regardless of whether the link changed or not.

dovholuknf: sure, i just find it redundant is all :)

plorenz: For links I tend to err on the side of verbosity, since link events tend to be comparatively low in number.

PR #3208, opened 2025-08-09
Allow router embedders to customize configure before start. Fixes #3207

### LINKED ISSUE #3207: Allow router embedders to customize config before start

plorenz:


router/router.go
```diff
@@ -298,7 +300,17 @@ func (self *Router) GetConfig() *env.Config {
 	return self.config
 }
 
+func (self *Router) AddInitializer(f func(cfg *env.Config) error) {
```
andrewpmartinez: I am starting to believe that we need to be more diligent about documenting the intent of external-to-component or out-of-domain-obscure functions and interfaces.

andrewpmartinez (approved):

router/router.go
plorenz: I'll add some doc

plorenz: refactored to implement a small set of LifeCycle events

PR #3216, opened 2025-08-13
Refactor xgress_edge_tunnel_v2 to make it more pluggable

plorenz:
- **Refactor xgress_edge_tunnel_v2 to make it more pluggable **

### DIFF
--- a/controller/ioc/typed_registry.go
+++ b/controller/ioc/typed_registry.go
@@ -1,0 +1,173 @@
+/*
+	Copyright NetFoundry Inc.
+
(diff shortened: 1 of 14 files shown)


andrewpmartinez:
Why IoC if we have one configuration for each type? We simply don't have that much variable configuration, like a framework would. Just DI would be more tenable, but that in itself only adds more implicit behavior. The entire code base is slowly becoming hard and harder for anyone but us to decipher.

My current state is that we have outgrown the original implementation and need to pay some debt off.

PR #3220, opened 2025-08-17
Finish work of making edge listener unnecessary when using other xgress plugins that use the rdm. Fixes #3164

### LINKED ISSUE #3164: Router data model doesn't work correctly if the edge listener isn't enabled
Router data model doesn't work correctly if the edge listener isn't enabled. It seems to only sync on re-subscribe. 

We need to implement the hello response listener in all cases. We should also handle cert renewal in all cases, and not make that edge dependent.

plorenz:


router/state/certchecker.go
```diff
@@ -57,18 +65,24 @@ type CertExpirationChecker struct {
 	extender CertExtender
 }
 
-func NewCertExpirationChecker(id *identity.TokenId, edgeConfig *routerEnv.EdgeConfig, ctrls routerEnv.NetworkControllers, closeNotify <-chan struct{}) *CertExpirationChecker {
+func NewCertExpirationChecker(env CertEnv) *CertExpirationChecker {
```
andrewpmartinez, line 68: It appears the core of the work is not to depend on routerEnv.EdgeConfig and the rest is just converting to a monad;

PR #3222, opened 2025-08-17
Add indexes for authenticator username and fingerprint. Fixes #3219 Also use  externalId index in IdentityManager.

### LINKED ISSUE #3219: AuthenticatorManager ReadByFingerprint/Username should use indexes

plorenz:


andrewpmartinez (changes requested):
`"name"` used in field error for the new index lookups. Should probably be whatever field they actually are (e.g. "fingerprint").

plorenz:
> `"name"` used in field error for the new index lookups. Should probably be whatever field they actually are (e.g. "fingerprint").

Fixed

andrewpmartinez (approved):

PR #3232, opened 2025-08-20
OIDC authentication fails if the client api has a separate cert chain. Fixes #3231

### LINKED ISSUE #3231: OIDC authentication fails if the client api has a separate cert chain

plorenz:


andrewpmartinez (changes requested):
See comment. I think I have a lower level change that will cover both single and ha controllers.

andrewpmartinez (approved):

PR #3236, opened 2025-08-21
fix #3231 use root controller server certs for OIDC signing

### LINKED ISSUE #3231: OIDC authentication fails if the client api has a separate cert chain

andrewpmartinez:
- addresses HA vs non-HA signing
- fixes issue where APIs server w/ difference certs than the root identity
- adds documentation for env/appenv
- adds a new composite type for TLSCert JWT signers

### DIFF
--- a/controller/env/appenv.go
+++ b/controller/env/appenv.go
@@ -19,8 +19,6 @@ package env
 import (
 	"bytes"
 	"crypto"
-	"crypto/ecdsa"
-	"crypto/rsa"
 	"crypto/sha1"
 	"crypto/tls"
 	"crypto/x509"
@@ -113,13 +111,13 @@ type AppEnv struct {
 	InstanceId           string
 	AuthRateLimiter      rate.AdaptiveRateLimiter
 
-	serverSigner jwtsigner.Signer
-	ServerCert   *tls.Certificate
+	clientApiDefaultSigner *jwtsigner.TlsJwtSigner
 
 	TraceManager *TraceManager
 	timelineId   string
 }
 
+// GetPeerControllerAddresses returns the network addresses of peer controllers.
 func (ae *AppEnv) GetPeerControllerAddresses() []string {
 	return ae.HostController.GetPeerAddresses()
 }
@@ -145,6 +143,8 @@ func (ae *AppEnv) JwtSignerKeyFunc(token *jwt.Token) (interface{}, error) {
 	return pubKey, nil
 }
 
+// ValidateAccessToken verifies an access token and returns its claims if valid.
+// Checks token signature, audience, type, and revocation status.
 func (ae *AppEnv) ValidateAccessToken(token string) (*common.AccessClaims, error) {
 	accessClaims := &common.AccessClaims{}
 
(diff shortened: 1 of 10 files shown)


controller/webapis/client-api.go
```diff
@@ -61,7 +62,7 @@ func (factory ClientApiFactory) Validate(config *xweb.InstanceConfig) error {
 			if !clientApiFound && api.Binding() == ClientApiBinding {
 				for _, bindPoint := range webListener.BindPoints {
 					if bindPoint.Address == edgeConfig.Api.Address {
-						factory.appEnv.SetServerCert(webListener.Identity.ServerCert()[0])
+						factory.appEnv.SetClientApiDefaultCertificate(webListener.Identity.ServerCert()[0])
```
dovholuknf, line 65: i can't get behind this name

per the doc "SetClientApiDefaultCertificate configures the default JWT signer for client API operations"

seems like this should be called SetDefaultSigner or something more similar to that to me

controller/model/testing.go
```diff
@@ -19,11 +19,12 @@ package model
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"testing"
```
dovholuknf, line 22: just wondering if this should actually be named "testing_test.go" not "testing.go" or if it's somehow used elsewhere?

controller/jwtsigner/jwt.go
```diff
@@ -58,3 +77,54 @@ func (j *SignerImpl) Generate(claims jwt.Claims) (string, error) {
...
+// GetJwtSigningMethod determines the appropriate JWT signing method based on the
+// certificate's public key type and parameters.
+// For ECDSA keys, it selects ES256, ES384, or ES512 based on the curve bit size.
+// For RSA keys, it defaults to RS256.
+// Panics if the certificate has an unsupported key type or ECDSA curve size.
+func GetJwtSigningMethod(cert *tls.Certificate) jwt.SigningMethod {
+
+	var sm jwt.SigningMethod = jwt.SigningMethodNone
+
+	switch cert.Leaf.PublicKey.(type) {
```
dovholuknf, line 129: possible nil reference here, should that be checked?

controller/env/appenv.go
```diff
@@ -243,34 +245,62 @@ func (ae *AppEnv) ValidateServiceAccessToken(token string, apiSessionId *string)
...
 
-func (ae *AppEnv) GetServerCert() (serverCert *tls.Certificate, kid string, signingMethod jwt.SigningMethod) {
-	return ae.ServerCert, ae.serverSigner.KeyId(), ae.serverSigner.SigningMethod()
+// GetClientApiDefaultTlsJwtSigner returns the default JWT signer for client API operations.
+func (ae *AppEnv) GetClientApiDefaultTlsJwtSigner() *jwtsigner.TlsJwtSigner {
+	return ae.clientApiDefaultSigner
 }
 
+// GetRootTlsJwtSigner creates and returns a JWT signer using the root server certificate.
+func (ae *AppEnv) GetRootTlsJwtSigner() *jwtsigner.TlsJwtSigner {
```
dovholuknf, line 254: this function strikes me to be the sorta function that should be on the `*identity.TokenId` itself, no?

controller/webapis/client-api.go
andrewpmartinez: There are many signers for different security scopes. "DefaultSigner" is also incorrect. I'm not super interested in naming semantics - I don't believe there is any perfect name and honestly I think we can get rid of this. It simply isn't my goal with this PR.

controller/model/testing.go
andrewpmartinez: I don't have a horse in this race. It is a utility file for other tests. It doesn't hold any tests itself, so giving it the `_test.go` postfix is not correct either, as Go's testing tooling will look for tests inside of it with that naming.

controller/env/appenv.go
andrewpmartinez: It can't. The identity library parses one identity section and is unaware of other sections. It has no concept of the root controller identity. Just the identity it loads.

controller/jwtsigner/jwt.go
andrewpmartinez: Technically yes. I'll address.

PR #3253, opened 2025-09-05
Add proxy.v1 suport and add support to ER/T for it in proxy and tproxy interceptors. Fixes #3258

### LINKED ISSUE #3258: Add config type proxy.v1 so proxies can be defined dynamically for the ER/T
Add config type proxy.v1 so proxies can be defined dynamically for the ER/T.
Also suport the proxy.v1 config type when using the tproxy interceptor.

plorenz:


controller/db/migration_initialize.go
```diff
@@ -611,6 +613,49 @@ var interfacesConfigTypeV1 = &ConfigType{
 	},
 }
 
+var ProxiesV1TypeId = "proxies.v1"
```
scareything: IMO "proxy.v1" (singular) is a better fit with the other configuration type names.

controller/db/migration_initialize.go
```diff
@@ -611,6 +613,49 @@ var interfacesConfigTypeV1 = &ConfigType{
...
+					"udp",
+				},
+				"description": "Network protocol",
+			},
+		},
+		"required": []interface{}{
+			"port",
+			"protocols",
+		},
+		"properties": map[string]interface{}{
```
scareything, line 669: It would be nice to support binding to a specific IP for a given service, defaulting to loopback IPs (probably ipv4 and ipv6).

The proxy implementation currently binds all services to `net.IPv4zero`, which was OK when it was being used as a developer tool but is not the best security stance for the product.

controller/db/migration_initialize.go
plorenz: I'll update it to proxy.v1

controller/db/migration_initialize.go
plorenz: Should be straightforward to add at least a simple version of this

plorenz:
> I gather that the tunnel binding still needs to set `mode: proxy` for this to work. That's ok. I'm just thinking of ways to avoid (or help with) confusion that might arise if someone is surprised that their service isn't being "seen" by the tunneler because it isn't using the correct mode.

I'll see how much trouble it would be to add this to tproxy as well.

tunnel/intercept/proxy/proxy.go
```diff
@@ -17,49 +17,69 @@
...
 
 type interceptor struct {
 	interceptIP net.IP
 	services    map[string]*Service
+	lock        sync.Mutex
+}
+
+func NewDelegate() intercept.Interceptor {
+	return &interceptor{
+		interceptIP: net.IPv4zero,
```
scareything: 127.0.0.1 would be a better default. The all zeros address can also sneak in when the interceptor is initialized by the router in proxy mode (`router/xgress_edge_tunnel_v2/tunneler.go), but I realize that would be a change in behavior

tunnel/entities/proxies.v1
```diff
@@ -0,0 +1,34 @@
+{
```
scareything: should this file have a `.json` extension? also it's missing the binding (or whatever it ends up being named) property.

controller/db/migration_initialize.go
plorenz: done

tunnel/intercept/proxy/proxy.go
plorenz: fixed

tunnel/entities/proxies.v1
plorenz: Fixed both, forgot to update it after making changes to the code version.

scareything (approved):

PR #3273, opened 2025-09-19
Fix UPDB Authentication Lockout Logic and Improve Test Coverage

jan94:
Hi OpenZiti Devs!

I have noticed that there is bug in the way how UPDB logins are handled when ```maxAttempts``` and ```lockoutDurationMinutes``` is configured on the auth-policy attached to the respective identity. I have raised that [issue ](https://openziti.discourse.group/t/issues-observed-with-auth-policy-configuration-via-zac-in-openziti-controller-v1-5-4/4998/3 ) in the discourse forum, but it has not yet received a response, so I investigated my self and implemented this fix. 🤓 

Please have a look and consider merging it! 🙂 

Thank you and have a nice weekend
Jan 

### Summary of Changes

_UPDB Authentication Logic (controller/model/authenticator_mod_updb.go):_

- Refactored login attempt counting: Only failed authentication attempts increment the counter; successful logins reset it.
- Lockout logic now disables identities only after the configured number of failed attempts (not on successful attempts).
- Removed attempts tracking after lockout or successful authentication to prevent stale state.

_Tests (tests/auth_updb_test.go):_

- Updated and expanded test cases to match new authentication logic:
    - Added a test to ensure disabled identities cannot log in.
    - Adjusted test logic to reflect that only failed attempts count toward lockout.
    - Added checks for disabledAt and disabledUntil fields when identity is locked.
    - Verified lockout duration matches configuration.

_Test Context (tests/context.go):_
- Added fields for test user authenticator and authentication policy to support new and updated tests.
- Configured test user authentication policy with lockout settings.

### Impact

- Ensures that only failed login attempts are counted toward lockout.
- Improves reliability and coverage of authentication and lockout tests.
- This PR strengthens UPDB authentication security and test coverage, ensuring correct lockout and reset behavior.

fix openziti/ziti#3333

### DIFF
--- a/tests/authenticator_test.go
+++ b/tests/authenticator_test.go
@@ -54,7 +54,7 @@ func Test_Authenticators_AdminUsingAdminEndpoints(t *testing.T) {
 		authenticatorsBody, err := gabs.ParseJSON(resp.Body())
 		req.NoError(err)
 
-		t.Run("can see three authenticators", func(t *testing.T) {
+		t.Run("can see four authenticators", func(t *testing.T) {
 			req := require.New(t)
 			count, err := authenticatorsBody.ArrayCount("data")
 


jan94:
Side note:
I won't be available for the next two weeks, but after that I am happy to do any changes - in case necessary 🙂

jan94:
Just noticed that the test run has failed. This is most likely due to me adding an additional ```TestUser``` to the [test context](https://github.com/openziti/ziti/pull/3273/files#diff-4245c6d1cf7b448eea80e289a08024c6205d1a3d2c671f92514f96992dae6428). 

The test ```Test_Authenticators_AdminUsingAdminEndpoints``` expects to receive three authenticators from the controller, this is now different, because of the added ```TestUser```.  I have now adjusted the test accordingly and all tests again.

jan94:
You are most welcome 👍

jan94:
Is there anything I can do to help or support you with implementing the fix for this issue?

I’ve just re-tested the behavior using the latest quickstart, following the steps exactly as described in openziti/ziti#3333. The lockout behavior still appears as reported. This means users of the Ziti management API could potentially end up in an unintended permanent lockout state, since the lockout is not being reset (according to my tests).

Please let me know if there’s anything I can contribute to move this forward!

PR #3274, opened 2025-09-22
add dns-upstream option for tproxy nameserver

qrkourier:
Goal: handle unanswerable queries in the most compatible manner that enables the broadest set of use cases involving a Ziti DNS nameserver

Outcomes:
- broaden the use cases for Go tproxy w/ nameserver to include MUSL clients - they ignore any IPv4 A-record response if the parallel IPv6 AAAA-record query gets an error ([see related issue](https://github.com/openziti/ziti/issues/3075))
- enable client applications to resolve non-Ziti domain names with the same DNS configuration that resolves Ziti domain names, e.g., `127.0.0.1, 10.96.0.10`

Implementation:
- if query type is AAAA, and an A record exists in local/internal lookup, then answer NOERROR (rcode = success) with empty recordset
- add an upstream DNS nameserver option that, when set, forwards unanswerable queries to the specified `<proto>://<ip>:<port>`, and always propagates the upstream's verbatim response to the caller
- if no upstream is specified, handle unanswerable queries configurably: timeout (no response), SERVFAIL, REFUSED (default) 

Tested:
- with TCP and UDP upstream nameservers - the forwarding client is configured with the protocol from the dns-upstream option's URI scheme
- with GLIBC and MUSL clients

Note: It's still necessary to configure a second nameserver for the task/pod/service because the Ziti nameserver isn't available initially, and the tunneler itself uses the second nameserver to resolve the domain names of the controllers and routers.

### DIFF
--- a/ziti/tunnel/root.go
+++ b/ziti/tunnel/root.go
@@ -133,6 +133,7 @@ func rootPostRun(cmd *cobra.Command, _ []string) {
 		})
 
 		if err != nil {
+			log.WithError(err).Error("failed to start CLI agent")
 		}
 	}
 


tunnel/dns/server.go
```diff
@@ -148,20 +221,76 @@ func (r *resolver) ServeDNS(w dns.ResponseWriter, query *dns.Msg) {
...
 
+	r.handleUnanswerable(w, query)
+}
+
+func (r *resolver) handleUnanswerable(w dns.ResponseWriter, query *dns.Msg) {
+	switch r.unanswered {
+	case unansweredTimeout:
+		log.Tracef("unanswerable query for %s: handling with timeout (no response)", query.Question[0].Name)
+		return
+	case unansweredServfail:
```
scareything, line 275: the only difference between the srvfail and refused paths is the response code. imo it would be better to just set an rcode  variable in the case statement and not duplicate the source code that sends the response.

scareything (approved):

PR #3290, opened 2025-09-29
Update deps for v1.5.x to address CVEs. Add production flag to allow excluding demo CLI command

plorenz:


ziti/cmd/demo/root_production.go
```diff
@@ -0,0 +1,28 @@
...
+*/
+
+package demo
+
+import (
+	"github.com/openziti/ziti/ziti/cmd/common"
+	"github.com/spf13/cobra"
+)
+
+func NewDemoCmd(p common.OptionsProvider) *cobra.Command {
```
andrewpmartinez, line 26: ?

I'm not getting what this does.  I see the `production` build flag. I assume there is a NewDemoCmd with/ implementation somewhere, and this is a stand-in, but why?

plorenz: There was a request to allow building without the demo CLI subcommand, so the build doesn't include the markdown library, which has dependencies which have a (low priority) CVE without a fix. That code is never used by us, since it's image -> ascii code and we don't have any images in our ascii markdown, but some companies have a very strict no CVE policy, so this affords a workaround

plorenz: Note the build tag at the top

andrewpmartinez: I assume there is no fixed version of the lib and this is a downstream problem then?

plorenz: Correct. Specifically it's https://github.com/advisories/GHSA-q7pp-wcgr-pffx
We'll have to figure out what to do with the markdown lib, but this is a short term workaround.

PR #3296, opened 2025-09-29
stop requiring a list of services to bind; resolves #3295

### LINKED ISSUE #3295: make ziti tunnel service:port pairs optional
support the new dynamic reverse proxy ports run mode by no longer requiring a list of service:port pairs when running `ziti tunnel proxy`. Instead, without any static ports to bind, the process will attempt to bind the port specified in any dialable services with new config type `proxy.v1`

qrkourier:


dovholuknf (approved):

ziti/tunnel/root.go
```diff
@@ -190,6 +190,7 @@ func startIdentity(cmd *cobra.Command, serviceListenerGroup *intercept.ServiceLi
 		entities.InterceptV1,
 		entities.HostConfigV1,
 		entities.HostConfigV2,
+		entities.ProxyV1,
```
plorenz, line 193: Good catch!

PR #3303, opened 2025-09-30
alert events

### LINKED ISSUE #3258: Add config type proxy.v1 so proxies can be defined dynamically for the ER/T

### LINKED ISSUE #3259: Interfaces config type not added due to wrong name
Due to a copy-paste error,  the interfaces config type had the same name as ziti-tunneler-server.v1. Since that name already existed the new type wasn't added.

### LINKED ISSUE #3264: Allow routers to generate alert events in cases of service misconfiguration
Provide a general alert event mechanism that can be used for allowing routers to notify when a service is misconfigured or there is an unexpected error when intercepting or hosting services.

### LINKED ISSUE #3269: Add host-interfaces config type
There's now a config type for specifying interfaces to apply intercepts to. The same type of information is needed on the host side. Add a host-interfaces.v1 config type, so that a service can specify interfaces both on the host side and intercept side.

plorenz:
- **Fix interfaces.v1 config type name. Fixes #3259**
- **Add proxy.v1 suport and add support to ER/T for it in proxy and tproxy interceptors. Fixes #3258**
- **Add host interfaces config type. Fixes #3269**
- **Add alert events. Fixes #3264**


common/pb/ctrl_pb/ctrl.proto
```diff
@@ -320,4 +322,18 @@ message LinkStateUpdate {
...
+}
+
+message Alert {
+  string sourceType = 1;
+  string sourceId = 2;
+  int64 timestamp = 3;
+  string severity = 4;
+  string message = 5;
+  repeated string details = 6;
+  map<string, string> relatedEntities = 7;
```
andrewpmartinez, line 334: Is this being positioned as a generic alerting feature? It reads like it.

plorenz: Yes, relatively generic. The focus is on surfacing things from external components.

PR #3336, opened 2025-10-22
Remove unused update deps workflow. Remove tests from release workflow, as they will have run before merging to main

plorenz:


qrkourier (approved):
~~I agree the test job is redundant if branch checks must pass before merging to main. Can you confirm that's enforced in admin settings? I think it's under "branch protections."~~

EDIT: moved conversation to chat

Issue #2324, opened 2024-08-14
Add OIDC enrollment


andrewpmartinez:

andrewpmartinez:
This will add additional configuration to external JWT signers.

- a new option to enable enrollment (ext jwt signers will now be able to operate in authentication and/or enrollment modes)
- claim mapping for enrollment and claims mapping for authentication (

andrewpmartinez:
Also see #1352

PR #3342, opened 2025-10-27
fixes openziti/ziti#2324 add token based enrollment

andrewpmartinez:
- allows enrollment to certificate auth
- allows enrollment to ext jwt token auth
- alters ext jwt claimsProperty (maps identity id) to support JSON pointers, defaults to `/sub`
- adds ext jwt enrollToCert, enrollToToken to controller  valid enrollment end-authenticator state
- adds ext jwt enrollAuthPolicyId to map end identity auth policy to, defaults to `default`
- adds ext jwt enrollAttributeSelector, supports single field name or JSON pointer to point to a single string or array of string attributes to give the identity, defaults to no selector
- adds ext jwt enrollNameSelector, supports single field name or JSON pointer to a string field to use as the name, defaults to `/sub`
- add enrollment errors to determine if enrollment has occurred
- adds CLI support for ext jwt signer enroll flags

### DIFF
--- a/tests/enrollment_token_to_token_test.go
+++ b/tests/enrollment_token_to_token_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/golang-jwt/jwt/v5"
 	"github.com/openziti/edge-api/rest_model"
 	edgeApis "github.com/openziti/sdk-golang/edge-apis"
+	"github.com/openziti/ziti/controller/apierror"
 )
 
 // Test_EnrollmentToken_Certificate uses a token issued from a 3rd party IdP, usually a JWT, in order to enroll a client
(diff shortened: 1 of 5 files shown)


controller/model/enrollment_mod_token.go
```diff
@@ -0,0 +1,348 @@
...
+		return nil, apierror.NewInvalidEnrollmentToken()
+	}
+
+	if !verificationResult.IsValid() {
+		pfxlog.Logger().WithError(err).Errorf("token failed verification")
+		return nil, apierror.NewInvalidEnrollmentToken()
+	}
+
+	if verificationResult.NameClaimValue == "" {
+		pfxlog.Logger().Error("token verified but name claim values was empty, cannot create identities with blank names")
```
dovholuknf: "name claim values was empty" --> 'value'?

dovholuknf: also does a .trim happen at some point

controller/model/enrollment_mod_token.go
```diff
@@ -0,0 +1,348 @@
...
+		return nil, apierror.NewInvalidEnrollmentToken()
+	}
+
+	if verificationResult.IdClaimValue == "" {
+		pfxlog.Logger().Error("token verified but identity id claim values was empty, cannot check for identities with a blank value")
```
dovholuknf: same wrt values/value. seems to read odd and doesn't align to var name "IdClaimValue"

controller/model/enrollment_mod_token.go
```diff
@@ -0,0 +1,348 @@
...
+			return nil, errorz.NewUnhandled(errors.New("could not create identity"))
+		}
+
+		content = &rest_model.Empty{}
+		textContent = []byte("")
+	}
+
+	return &EnrollmentResult{
+		Identity:      newIdentity,
+		Authenticator: newAuthenticator,
```
dovholuknf, line 241: is it intended to allow newAuthenticator  to be nil here? it looks like it's possible

controller/model/enrollment_mod_token.go
andrewpmartinez: No, not our job to clean up their data. What is the space is significant?

controller/model/enrollment_mod_token.go
andrewpmartinez: Yes if the end result is token auth there is no authenticator.

controller/model/enrollment_mod_token.go
andrewpmartinez: Old value, not changing the external API and CLI for it.

controller/model/enrollment_mod_token.go
andrewpmartinez: Removed the plurals.

controller/model/enrollment_mod_token.go
andrewpmartinez: Fixed plurals.

tests/enrollment_token_to_token_test.go
```diff
@@ -0,0 +1,402 @@
...
+			ctx.Req.NotEmpty(enrollmentJwt)
+
+			clientApi := ctx.NewEdgeClientApi(nil)
+			ctx.Req.NotNil(clientApi)
+
+			err = clientApi.CompleteJwtTokenEnrollmentToTokenAuth(enrollmentJwt)
+
+			ctx.Req.Error(err)
+		})
+	})
```
dovholuknf, line 570: i see a lot of Req.NoError tests around EnrollNameClaimsSelector but are there any negative tests? What if i pass `/` or a field that doesn't exist or a number when it should be a string etc?

dovholuknf: some negative testing around EnrollAttributeClaimsSelector seems like a good idea too

ziti/cmd/edge/update_extjwtsigner.go
```diff
@@ -108,6 +115,12 @@ func newUpdateExtJwtSignerCmd(out io.Writer, errOut io.Writer) *cobra.Command {
 	cmd.Flags().StringVarP(options.ExtJwtSigner.ClientID, "client-id", "", "", "The client id for OIDC that should be used")
 	cmd.Flags().StringSliceVarP(&options.ExtJwtSigner.Scopes, "scopes", "", nil, "The scopes for OIDC that should be used")
 	cmd.Flags().StringVarP(&options.targetToken, "target-token", "", "", "The target token SDKs should use")
+	cmd.Flags().BoolVarP(options.ExtJwtSigner.EnrollToCertEnabled, "enroll-to-cert", "", false, "Enable enrollment to certificate authentication, defaults to false")
+	cmd.Flags().BoolVarP(options.ExtJwtSigner.EnrollToTokenEnabled, "enroll-to-token", "", false, "Enable enrollment to a token authentication, defaults to false")
+	cmd.Flags().StringVarP(options.ExtJwtSigner.EnrollNameClaimsSelector, "enroll-name-claims-selector", "", "", "The claims JSON pointer selector or top level property to use for the name of enrolling identities, defaults to /sub")
+	cmd.Flags().StringVarP(options.ExtJwtSigner.EnrollAttributeClaimsSelector, "enroll-attr-claims-selector", "", "", "The claims JSON pointer selector or top level property to use for the attributes of enrolling identities")
```
dovholuknf, line 121: i didn't see a test where the selector pointed to an array. Seems like EnrollNameClaimsSelector should fail but EnrollAttributeClaimsSelector should apply them all?

CHANGELOG.md
```diff
@@ -1,3 +1,123 @@
...
+### External JWT Signer Configuration
+
+External JWT signers are configured via the Edge Management API to define enrollment behavior with the following new 
+enrollment-specific properties:
+
+- **enrollToCertEnabled** - When enabled, identities can exchange a JWT token and a certificate signing request (CSR) 
+        for a client certificate during enrollment. The certificate can then be used for standard certificate-based
+        authentication.
+
+- **enrollToTokenEnabled** - When enabled, identities can use a JWT token to enroll. The current token or future tokens
```
dovholuknf, line 22: > When enabled, identities can use a JWT token to enroll

enroll **to** token? or **from** token? the words seem to be disjointed from the flag? should the flag be `enrollFromTokenEnabled` ?

dovholuknf: i am not sure if i get it entirely, but this seems like how the controller works currently? can't i do this right now (submit a jwt to get an openziti token). Is this flag going to restrict that functionality then? Is that the idea? If so, I would ask that this be defaulted to __true__ since this is how it operates right now. (again, assuming i understand it properly)

andrewpmartinez: > enroll to token? or from token? 

`enrollToCertEnabled` and `enrollToTokenEnabled` are shortened versions of "enroll to certificate authentication" and "enroll to token authentication" however the names were long. So I took out the "authentication" part.

> i am not sure if i get it entirely, but this seems like how the controller works currently? 

No, this is just-in-time identity creation via token verification for enrollment.

> can't i do this right now (submit a jwt to get an openziti token)

You can submit an enrollment token from OpenZiti on a preprovisioned identity. This allows you to submit a token from an external IdP that asserts identity. Two different token types, for different  usages, from different issuers.

> Is this flag going to restrict that functionality then?

No. This is entirely new functionality.

> Is that the idea? If so, I would ask that this be defaulted to true since this is how it operates right now. (again, assuming i understand it properly)

It will not be due to the above.

tests/enrollment_token_to_token_test.go
andrewpmartinez: Sure.

ziti/cmd/edge/update_extjwtsigner.go
andrewpmartinez: Sure. Marking this resolve as your other comment covers this.

PR #3358, opened 2025-11-10
Optimize router data model subscription code. Add additional events.

plorenz:


common/router_data_model.go
```diff
@@ -54,13 +63,254 @@ type AccessPolicies struct {
 	PostureChecks map[string]*edge_ctrl_pb.DataState_PostureCheck
 }
 
+type serviceAccess struct {
```
andrewpmartinez, line 68: Should probably add some struct/method/function doc. Claude can do it.

Without looking at the rest of the PR or where `serviceAccess` is used, it seems odd to have a struct that just counts references and indices instead of using len() on a map or the like.

common/router_data_model.go
```diff
@@ -484,62 +871,207 @@ func (rdm *RouterDataModel) queueEvent(event subscriberEvent) {
...
+
 		rdm.Identities.Upsert(model.Identity.Id, nil, func(exist bool, valueInMap *Identity, newValue *Identity) *Identity {
 			if valueInMap == nil {
-				identity = &Identity{
-					DataStateIdentity: model.Identity,
-					ServicePolicies:   cmap.New[struct{}](),
-					identityIndex:     index,
-				}
+				identity.ServicePolicies = map[string]struct{}{}
+				identity.Services = map[string]*serviceAccess{}
```
andrewpmartinez: This is a minomer. `identity.Services` isn't a collection of services. It is at best "service access stats" or "service ref counts".

common/router_data_model.go
```diff
@@ -54,13 +63,254 @@ type AccessPolicies struct {
...
+		current = &serviceAccess{}
+	}
+	current.updateServicePolicies(policy, index, add)
+	if current.HasAccess() {
+		//fmt.Printf("identity: %s, service: %s, dial: %d, bind: %d \n", self.GetId(), serviceId, current.DialPoliciesCount, current.BindPoliciesCount)
+		if !present {
+			self.Services[serviceId] = current
+		}
+	} else if present {
+		//fmt.Printf("identity: %s, service: %s, dial: %d, bind: %d (REMOVING) \n", self.GetId(), serviceId, current.DialPoliciesCount, current.BindPoliciesCount)
```
andrewpmartinez: Dead debug prints.

common/router_data_model.go
```diff
@@ -100,6 +350,11 @@ func (self *Identity) Equals(other *Identity) bool {
 		return false
 	}
 
+	if len(self.ServiceConfigs) != len(other.ServiceConfigs) {
+		log.Info("identity updated, number of service configs changed")
```
andrewpmartinez: Should this be a debug or a trace? I don't know the frequency of this output, but I figure it might be high.

Also another on line 377.

common/router_data_model.go
```diff
@@ -116,28 +371,70 @@ func (self *Identity) Equals(other *Identity) bool {
...
+func (self *Config) ToProtobuf() *edge_ctrl_pb.DataState_Config {
+	return &edge_ctrl_pb.DataState_Config{
+		Id:       self.Id,
+		TypeId:   self.TypeId,
+		Name:     self.Name,
+		DataJson: self.DataJson,
+	}
 }
 
 func (self *Config) Equals(other *Config) bool {
```
andrewpmartinez, line 463: Does the config structure not matter here for equality?

common/router_data_model.go
```diff
@@ -187,7 +531,7 @@ type RouterDataModel struct {
 	ConfigTypes      cmap.ConcurrentMap[string, *ConfigType]                        `json:"configTypes"`
 	Configs          cmap.ConcurrentMap[string, *Config]                            `json:"configs"`
 	Identities       cmap.ConcurrentMap[string, *Identity]                          `json:"identities"`
-	Services         cmap.ConcurrentMap[string, *Service]                           `json:"services"`
+	Services         cmap.ConcurrentMap[string, *Service]                           `json:"Services"`
```
andrewpmartinez: Looks like an accidental refactor from `s` -> `S`

common/router_data_model.go
```diff
@@ -377,6 +741,29 @@ func (rdm *RouterDataModel) processSubscriberEvents() {
 			return
 		case evt := <-rdm.events:
 			evt.process(rdm)
+		case <-rdm.scanSubsNotify:
+			start := time.Now()
+			count := rdm.updatedIdentities.Count()
+			for entry := range rdm.updatedIdentities.IterBuffered() {
+				rdm.updatedIdentities.Remove(entry.Key)
+
+				if entry.Val&IdentityAdded == IdentityDeleted {
```
andrewpmartinez: Space around the bitwise operator (&) or did the code formatter make it that way? I did a triple check trying to read this line.

common/router_data_model.go
```diff
@@ -1293,6 +2349,14 @@ type compareReporter struct {
...
 func (self *compareReporter) PushStep(step cmp.PathStep) {
 	self.steps = append(self.steps, step)
+	//buf := bytes.Buffer{}
+	//for idx, s := range self.steps {
+	//	if idx != 0 {
+	//		buf.WriteString(".")
+	//	}
+	//	buf.WriteString(s.String())
+	//}
+	//fmt.Println(buf.String())
```
andrewpmartinez: Dead code.

common/router_data_model.go
```diff
@@ -561,50 +1105,136 @@ func (rdm *RouterDataModel) HandleConfigTypeEvent(index uint64, event *edge_ctrl
...
 func (rdm *RouterDataModel) HandleConfigEvent(index uint64, event *edge_ctrl_pb.DataState_Event, model *edge_ctrl_pb.DataState_Event_Config) {
 	if event.Action == edge_ctrl_pb.DataState_Delete {
-		rdm.Configs.Remove(model.Config.Id)
+		rdm.Configs.RemoveCb(model.Config.Id, func(key string, v *Config, exists bool) bool {
+			if v != nil {
+				if !v.identities.IsEmpty() {
+					panic(fmt.Errorf("config %s still has identity references", key))
+				}
+				if !v.services.IsEmpty() {
+					panic(fmt.Errorf("config %s still has service references", key))
```
andrewpmartinez: I want to confrim the intention here:

1) crash the router
2) not use pfxlog.Logger().Fataf if so

common/router_data_model.go
```diff
@@ -998,33 +1701,106 @@ func CloneMap[V any](m cmap.ConcurrentMap[string, V]) cmap.ConcurrentMap[string,
...
+	}
+}
+
+func (rdm *RouterDataModel) withLockedIdentity(panicOnNil bool, identityId string, f func(identity *Identity)) {
+	if identity, _ := rdm.Identities.Get(identityId); identity != nil {
+		identity.lock.Lock()
+		defer identity.lock.Unlock()
+		f(identity)
+	} else if panicOnNil {
+		panic(fmt.Errorf("identity '%s' not found", identityId))
```
andrewpmartinez: Raw panic:

1) really want to crash the router?
2) not use pfxlog.Logger().Fatal?

common/router_data_model.go
```diff
@@ -998,33 +1701,106 @@ func CloneMap[V any](m cmap.ConcurrentMap[string, V]) cmap.ConcurrentMap[string,
...
+	}
+}
+
+func (rdm *RouterDataModel) withPostureCheck(postureCheckId string, f func(PostureCheck *PostureCheck)) {
+	if postureCheck, _ := rdm.PostureChecks.Get(postureCheckId); postureCheck != nil {
+		f(postureCheck)
+	}
+}
+
+func (rdm *RouterDataModel) SubscribeToIdentityChanges(identityId string, subscriber IdentityEventSubscriber, isRecreatableIdentity bool) error {
```
andrewpmartinez: Why the change from `isRouterIdentity` to `isRecreatableIdentity`?

I know what a `router identity` is, but a `recreatable identity` is not something I'm familiar with. Is this a new concept? Does it go beyond router identities now? Most use cases in here are `false`. I can only find `true` use cases in older code that still references the router identity as the case.

common/router_data_model.go
```diff
@@ -1035,12 +1811,48 @@ func (rdm *RouterDataModel) InheritLocalData(other *RouterDataModel) {
...
+				rdm.loadServiceConfigs(sub.Identity, identityService)
+			} else {
+				log.WithField("serviceId", serviceId).Panic("service not found by id")
+			}
+		}
+
+		for postureCheckId := range identity.PostureChecks {
+			check, ok := rdm.PostureChecks.Get(postureCheckId)
+			if !ok {
+				log.WithField("postureCheckId", postureCheckId).Panic("posture check reference but not found by id")
```
andrewpmartinez: Another panic over data issues.

I would like to know if the right thing to do is to add a path to refresh from the controller instead of going down. At least trying once.  I suppose if we are a service, we will be restarted and maybe do that? Unless corrupt data was cached to disc.

common/subscriber.go
```diff
@@ -135,21 +141,35 @@ func (self *IdentityService) Equals(other *IdentityService) bool {
 type IdentitySubscription struct {
 	IdentityId string
 	Identity   *Identity
-	Services   map[string]*IdentityService
-	Checks     map[string]*PostureCheck
+	// Some identities, like router identities, may be deleted and recreated, as the tunneler flag is toggled
+	// If IsRecreatable is true then the subscription should remain active when the identity is deleted
+	IsRecreatable bool
```
andrewpmartinez: Ah, here we go. The origin of `isRecreatable`. I still haven't answered my question in the previous comment.

common/subscriber.go
```diff
@@ -409,53 +469,57 @@ type subscriberEvent interface {
...
+		close(self.c)
+	} else {
+		fmt.Printf("updated identities count is %d\n", rdm.updatedIdentities.Count())
+		time.Sleep(time.Second * 10)
+		go func() {
+			rdm.events <- self
+		}()
+	}
+	synced := rdm.subscribersSynced.Swap(0)
+	fmt.Printf("%d subscribers synced since last check\n", synced)
```
andrewpmartinez: A random prints (517,524)?

common/subscriber_identity_test.go
```diff
@@ -0,0 +1,179 @@
...
+	event := &edge_ctrl_pb.DataState_Event{
+		Action: edge_ctrl_pb.DataState_Create,
+		Model: &edge_ctrl_pb.DataState_Event_Identity{
+			Identity: identity,
+		},
+	}
+
+	self.handleEvent(event)
+	if self.rdm.subscriptions.Count() < self.maxSubscribers {
+		fmt.Printf("subscriber count < %d, adding subscriber %s\n", self.rdm.subscriptions.Count(), identityId)
```
andrewpmartinez, line 53: Did you mean to leave prints in tests? I don't know how I feel about it, but it isn't the worst thing ever. There were others, but I commented here.

andrewpmartinez (approved):
Nothing egregious; mostly concerns about:

-  info logs in what appear to be high-traffic areas
- raw panics, and logger panics over data inconsistencies
- not having Claude generate method, struct, etc doc.

The "is recreatable identity" confuses me, as I don't know of any other "recreatable identities" other than "router identities".

common/router_data_model.go
plorenz: The  main thing we care about is if the config definition changed, which is being checked. Also checking name and type id. The other fields are internal tracking. I feel like  I'm missing something on your question. We should only be comparing when ids are already matching, but I can add a check just to be safe.

common/router_data_model.go
plorenz: code formatter does this :-/

common/subscriber_identity_test.go
plorenz: The tests are performance tests, so they're meant to be run in isolation. The fmt output is a bit easier to follow than logs.

common/router_data_model.go
plorenz: fixed

common/router_data_model.go
plorenz: fixed

common/router_data_model.go
plorenz: fixed

common/router_data_model.go
plorenz: fixed

common/router_data_model.go
plorenz: fixed

common/router_data_model.go
plorenz: removed the panics. I wanted fast fail during tests, but now that the test is running clean, we shouldn't hit this. Maybe I need to add a test mode

common/router_data_model.go
plorenz: removed

common/router_data_model.go
plorenz: reverted. I was trying to make it more generic, but since we only have the one case, more generic makes it more mysterious for no good reason

common/subscriber.go
plorenz: reverted, see above

common/subscriber.go
plorenz: test code, removed

common/router_data_model.go
plorenz: Added doc

common/router_data_model.go
plorenz: removed panics and replaced them with error messages where appropriate. I think if there's a data issue, we'd probably have to restart the controller to rebuild the router data model there, otherwise you'd get the same sequence of events that caused the issue originally

router/env/env.go
```diff
@@ -47,6 +47,7 @@ type RouterEnv interface {
 	GetCtrlRateLimiter() rate.AdaptiveRateLimitTracker
 	GetVersionInfo() versions.VersionProvider
 	GetRouterDataModel() *common.RouterDataModel
```
plorenz, line 49: WithRouterDataModel can be used to add a subscriber while being sure that the router data model won't be replaced out from under you while the subscription is being added. It's probably a very unlikely race condition, but best avoided.

common/event_cache.go
```diff
@@ -31,94 +52,36 @@ type EventCache interface {
...
+	LogSize       int                                          `json:"-"`
+	Log           []uint64                                     `json:"-"`
+	Events        map[uint64]*edge_ctrl_pb.DataState_ChangeSet `json:"-"`
 }
 
 func NewLoggingEventCache(logSize uint64) *LoggingEventCache {
 	return &LoggingEventCache{
 		HeadLogIndex: 0,
-		LogSize:      logSize,
+		LogSize:      int(logSize),
```
andrewpmartinez, line 84: Why go from uint64 to int? Do negative numbers have some important value?

common/router_data_model.go
andrewpmartinez: I was asking if there is a missed case where the ID/name remain the same, but the config's other value(s) are altered.

common/router_data_model.go
```diff
@@ -181,8 +571,7 @@ type ServicePolicy struct {
 // for the first time/after disconnects. Receive mode does not maintain an event cache and does not support replays.
 // It instead is used as a reference data structure for authorization computations.
 type RouterDataModel struct {
-	EventCache
-	listeners map[chan *edge_ctrl_pb.DataState_ChangeSet]struct{}
+	// EventCache
```
andrewpmartinez: Dead?

common/router_data_model.go
```diff
@@ -387,95 +805,86 @@ func (rdm *RouterDataModel) Stop() {
...
 			logger.
 				WithField("entry", idx).
 				WithField("action", event.Action.String()).
 				WithField("type", fmt.Sprintf("%T", event.Model)).
 				WithField("summary", event.Summarize()).
-				Debug("handling change set entry")
-			if rdm.Handle(index, event) {
-				syncSubscribers = true
-			}
+				Info("handling change set entry")
```
andrewpmartinez: Seems high volume. Was this an intentional Debug > Info?

common/router_data_model.go
```diff
@@ -1325,3 +2489,205 @@ func (self *compareReporter) Report(result cmp.Result) {
 func (self *compareReporter) PopStep() {
 	self.steps = self.steps[:len(self.steps)-1]
 }
+
+// ToJson returns a map representation of the RouterDataModel with recursively exported data
+func (rdm *RouterDataModel) ToJson() map[string]interface{} {
```
andrewpmartinez: Kind of a misnomer, this is not returning JSON. It does return a map of nested interfaces that can be serialized to JSON.

If the goal is to actually return JSON, it should probably implement the stadard JSON package interface for `json.Marshaler`.

andrewpmartinez (approved):

common/event_cache.go
plorenz: It just prevents us having to do type conversion when it's used. It's still specified as a uint, so we can't get invalid values from the config.

common/router_data_model.go
plorenz: fixed

common/router_data_model.go
plorenz: Missed that one, thank you. Fixed

common/router_data_model.go
plorenz: renamed to `ToMap` :+1:

common/router_data_model.go
plorenz: The config's data is also compared in the Equals method

PR #3386, opened 2025-11-21
Update deps, changelog and sync workflows

plorenz:


.github/workflows/release.yml
```diff
@@ -134,34 +134,6 @@ jobs:
           path: release/
           retention-days: 5
 
-  tests:
```
dovholuknf, line 137: removing all these is intentional????

.github/workflows/main.yml
```diff
@@ -169,14 +169,31 @@ jobs:
...
           go test ./... --tags apitests
 
+      - name: Run CLI tests and Integration Tests
+        if: ${{ vars.ZITI_SKIP_CLI_TESTS != 'true' }}
+        timeout-minutes: 10
+        shell: bash
+        run: |
+          mkdir cli_tests_bin
+          go build -o cli_tests_bin ./... 
+          ZITI_CLI_TEST_ZITI_BIN="$PWD/cli_tests_bin/ziti" go test ./... --tags cli_tests
```
dovholuknf, line 179: were these all merged into this branch? did you backport them?

.github/workflows/release.yml
plorenz: Only from the release flow, it's still done on the pr flows

.github/workflows/main.yml
plorenz: tests weren't backported, but the workflow still passes, I think b/c it doesn't see any tests

PR #3388, opened 2025-11-21
fixes #1330 check for SANs conflicts on startup

### LINKED ISSUE #1330: Verify server sans when specifying alt server certs
It's very easy to misconfigure certs and alt server certs with overlapping SANS. It would be helpful to introspect the certificates and alt server certs and panic on startup if the two settings allow for overlapping alt server certs. When the sans overlap, non deterministic behavior occurs making it hard to track down the actual issue

andrewpmartinez:
- requires xweb dep update

### DIFF
--- a/controller/config/config.go
+++ b/controller/config/config.go
@@ -23,10 +23,17 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
+	"math"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+
 	"github.com/hashicorp/go-hclog"
 	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/channel/v4"
 	"github.com/openziti/identity"
+	"github.com/openziti/sdk-golang/xgress"
 	"github.com/openziti/storage/boltz"
 	"github.com/openziti/transport/v2"
 	transporttls "github.com/openziti/transport/v2/tls"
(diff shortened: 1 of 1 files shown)


controller/config/config.go
```diff
@@ -198,6 +199,21 @@ func LoadConfig(path string) (*Config, error) {
...
+			conflictsStr := ""
+
+			for _, conflict := range conflicts {
+				if conflictsStr != "" {
+					conflictsStr += ", "
+				}
+				conflictsStr += conflict.Error()
+			}
+
+			return nil, fmt.Errorf("conflicting SANs found in root [identity] section server certificate, ensure each server certificate is the sole representation of each DNS/IP SAN: %s", conflictsStr)
```
dovholuknf, line 214: does it only check the root identity section? it'd have to synthesize all of them or does that happen during config-time?

qrkourier (commented):
In case of a misconfigured controller with at least one conflicting SAN, will upgrading to this revision cause the controller to abort/panic during startup, or emit a warning?

If the upgraded controller will no longer run when misconfigured, the steps to recover from the outage include reissuing the conflicting server certificates which are bound to the controller's TLS listeners, correct?

controller/config/config.go
andrewpmartinez: The other sections are xweb and handled by this PR: https://github.com/openziti/xweb/pull/31

PR #3393, opened 2025-11-24
Refine ci actions

qrkourier:


.github/workflows/release-quickstart.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     # pushes trigger a quickstart release with the same tag name and the same
     # ziti binary release version
     if: (github.repository_owner == 'openziti' || vars.ZITI_QUICKSTART_IMAGE != '') && (
-        startsWith(github.ref_name, 'v') || (
+        (startsWith(github.ref_name, 'v') && !contains(github.ref_name, '-')) || (
```
dovholuknf, line 23: is it just me or does this seems wrong already? wouldn't this capture "very-big-branch" etc?

qrkourier: Only if it was a release-triggering tag ref starting with a `v` character

PR #3436, opened 2025-12-09
Fix OIDC configuration parsing for token durations

ekoby:


dovholuknf (approved):

controller/config/config_edge.go
dovholuknf, line 295: lots of 'AccessToken' references in here? intentional?

ekoby: yeah, you need to read further down

PR #3454, opened 2025-12-15
ignore prereleases when comparing highest version

### LINKED ISSUE #3451: ci - compare only stable releases when promoting
The stable release promotion workflow is careful to avoid clobbering the Docker `:latest` stable release tag. It compares the new stable version to all previous tags to ensure it is highest before claiming latest. 

However, not all prior releases are stable, so it will fail to claim latest when it should if there's a higher prerelease. For example, 1.6.12 would have become Docker latest, but was blocked from promotion because 1.7.2 was incorrectly counted as a higher stable release when it is a prerelease.

qrkourier:
resolves #3451

### DIFF
--- a/.github/workflows/promote-downstreams.yml
+++ b/.github/workflows/promote-downstreams.yml
@@ -85,7 +85,17 @@ jobs:
           set -o pipefail
           set -o xtrace
           
-          HIGHEST_VERSION=$(git tag -l 'v*.*.*' | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -n1)
+          HIGHEST_VERSION=$(
+            gh release list --repo "${GITHUB_REPOSITORY}" \
+                            --limit 999 \
+                            --exclude-drafts \
+                            --exclude-pre-releases \
+                            --json tagName \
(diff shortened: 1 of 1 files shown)


.github/workflows/promote-downstreams.yml
```diff
@@ -85,14 +85,16 @@ jobs:
           set -o pipefail
           set -o xtrace
           
-          HIGHEST_VERSION=$(git tag -l 'v*.*.*' | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -n1)
+          HIGHEST_VERSION=$(gh release list --repo "${GITHUB_REPOSITORY}" --exclude-drafts --limit 999 --json tagName,isPrerelease --jq '.[]|select(.isPrerelease == false)|.tagName' | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -n1)
```
ekoby: FLAGS
      --exclude-drafts         Exclude draft releases
      --exclude-pre-releases   Exclude pre-releases

ekoby (commented):
`gh releases list` already has the option you want

PR #3478, opened 2025-12-23
Have identity env updates and system authenticator updates happen in …

### LINKED ISSUE #3477: Optionally enable model changes triggered by login to be non-blocking and to be droppable if the system is under load
If an identity is authenticating for the first time, or if something has changed, the sdk/env info for the identity will be updated during authentication.
Similarly, certain conditions will cause the identity authenticator to be updated.

Block of these operations can block the authentication process if the system is under load. If there are many identities authenticating at the same time, this can make a thundering herd problem worse. 

These operations should be queued and processed in the background. If the queue fills up, then the we should allow these updates to be dropped, as the values will be filled in the next time the identity authenticates.

plorenz:
…the background. Fixes #3477


controller/model/command.go
```diff
@@ -91,6 +136,31 @@ func (self *CommandManager) decodeDeleteEntityCommand(_ int32, data []byte) (com
...
+			if err := self.backgroundPool.Queue(f); err != nil {
+				return false
+			}
+		}
+		return true
+	}
+
+	f()
+	return true
+}
```
andrewpmartinez, line 162: I don't know how the queue will fail or become full, but losing these updates has some implications.

I think the most pertinent things at the top of my head are the fact that these values are shown in UIs for OS/Version info. Additionally, these values are consulted by ops during upgrades/looking for misconfigured clients.

The old posture system also uses some of these values, but I suppose that is less of a factor as that system will be deprecated at some point.

I'm not sure if these are big enough concerns to not do this, but they are worth considering.

plorenz: Made the drops disabled by default for now. As discussed we should update sdk/env info separately from auth

PR #3520, opened 2026-01-26
Update ziti version to 2.0

plorenz:


controller/events/dispatcher_alert.go
```diff
@@ -50,7 +50,7 @@ func (self *Dispatcher) registerAlertEventHandler(_ string, val interface{}, _ m
 	handler, ok := val.(event.AlertEventHandler)
 
 	if !ok {
-		return errors.Errorf("type %v doesn't implement github.com/openziti/ziti/controller/event/AlertEventHandler interface.", reflect.TypeOf(val))
+		return errors.Errorf("type %v doesn't implement github.com/openziti/ziti/v2/controller/event/AlertEventHandler interface.", reflect.TypeOf(val))
```
andrewpmartinez: This would probably be better as %T and %T instead of using %v and hard coding the other one. Can reference the nil `handler` and remove the direct reflect call.

controller/events/dispatcher_api_session.go
```diff
@@ -90,7 +90,7 @@ func (self *Dispatcher) registerApiSessionEventHandler(eventType string, val int
 	handler, ok := val.(event.ApiSessionEventHandler)
 
 	if !ok {
-		return errors.Errorf("type %v doesn't implement github.com/openziti/ziti/controller/event/ApiSessionEventHandler interface.", reflect.TypeOf(val))
+		return errors.Errorf("type %v doesn't implement github.com/openziti/ziti/v2/controller/event/ApiSessionEventHandler interface.", reflect.TypeOf(val))
```
andrewpmartinez: This would probably be better as %T and %T instead of using %v and hard coding the other one. Can reference the nil `handler` and remove the direct reflect call.

Same for the rest. I won't comment on the rest.

controller/events/dispatcher_alert.go
plorenz: I used %T for the first values, but getting the type name of an interface is fairly ugly, so I ended up just shortening the name. Did for all the dispatcher_* types.

controller/events/dispatcher_api_session.go
plorenz: Fixed, see above

Issue #3356, opened 2025-11-07
Add WWW-Authenticate Headers


andrewpmartinez:
Add WWW-Authenticate headers on standard 401 responses and other responses: `WWW-Authenticate: <auth-scheme> realm="<realm>", [parameters]`

**Use**
- Standard 401 unauthenticated responses can now include helpful support/debugging information such as which types of tokens were/were not supplied (legacy vs bearer) and why they failed (not provided, invalid, expired)
- For OIDC authentication, when an auth policy is configured with primary-ext-jwt auth and a secondary auth that matches it, the client doesn't need another JWT; it should be signaled to continue providing it. Intelligently using headers, OIDC authentication can signal to SDKs when and where additional bearer tokens are needed.
- To support secondary JWTs in general

**Scheme/Realms to use:**

- `openziti`/`zt-session` - Legacy authentication `zt-session` tokens
- `Bearer`/`openziti-oidc` - OpenZiti's new internal OIDC provider
- `Bearer`/`secondary-ext-jwt` - Secondary External JWT failure (as defined by auth policies)

**Examples:**

**401-No API Session tokens were provided:**

- `WWW-Authenticate zt-session realm="zt-session" error="missing" error_description="no token was provided"`
- `WWW-Authenticate Bearer realm="openziti-oidc" error="missing" error_description="no token was provided"`

**401 - An expired OpenZiti OIDC token was provided**

- `WWW-Authenticate Bearer realm="openziti-oidc" error="expired" error_description="the tokens was expired"`

**401 - An invalid zt-session token was provided**

- `WWW-Authenticate openziti realm="zt-session" error="invalid" error_description="the token is invalid"`

**401 - A valid API Session token was provided, but a second secondary JWT is required**

not provided 

- `WWW-Authenticate Bearer realm="openziti-secondary-ext-jwt" error="missing" error_description="this request requires an additional bearer token" id="<ext-jwt-id-ordered-csv>" " issuer="<issuer-ordered-csv>"`

or is invalid

- `WWW-Authenticate Bearer realm="openziti-secondary-ext-jwt" error="invalid" error_description="the token provided is invalid" id="<ext-jwt-id-ordered-csv>" " issuer="<issuer-ordered-csv>"`

or is expired

- `WWW-Authenticate Bearer realm="openziti-secondary-ext-jwt" error="expired" error_description="the token provided is expired" id="<ext-jwt-id-ordered-csv>" " issuer="<issuer-ordered-csv>"`

[...]

PR #3561, opened 2026-02-10
fixes openziti/ziti#3356 adds www-authenticate headers

andrewpmartinez:
- www-authenticate headers are returned on 401s from API requests
- www-authenticate headers are returned during authentication to signal addtional JWT beater tokens needed (secondary ext jwt)

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,8 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "Bash(go build:*)",
-      "Bash(grep:*)"
-    ]
-  }
-}


controller/env/appenv.go
```diff
@@ -622,173 +622,204 @@ func (a authorizer) Authorize(request *http.Request, principal interface{}) erro
...
+		securityToken.Error = errorz.NewUnauthorizedOidcInvalid()
+		return
 	}
 
 	var err error
-	rc.Identity, err = ae.GetManagers().Identity.Read(rc.Claims.Subject)
+	rc.Identity, err = ae.GetManagers().Identity.Read(securityToken.Claims.Subject)
 
 	if err != nil {
 		if boltz.IsErrNotFoundErr(err) {
```
plorenz: apiErr is set and then overwritten below

controller/env/appenv.go
```diff
@@ -883,18 +937,43 @@ func ProcessAuthQueries(ae *AppEnv, rc *response.RequestContext) {
 		extJwtAuth := extJwtAuthVal.(*model.AuthModuleExtJwt)
 		if extJwtAuth != nil {
 			authCtx := model.NewAuthContextHttp(rc.Request, model.AuthMethodExtJwt, nil, rc.NewChangeContext())
+
+			var err error
+
```
plorenz: err is nil, so this check is always false

.claude/settings.local.json
```diff
@@ -0,0 +1,7 @@
+{
```
plorenz: Should this go in your $HOME/.claude?

andrewpmartinez: It was auto generated? I don't know why. I have removed it.

plorenz (approved):

PR #3566, opened 2026-02-11
Link dialer save 'is first conn' true, so all dials claim to be first, causing potential race condition. Fixes #3565

### LINKED ISSUE #3565: Link dialer save 'is first conn' true, so all dials claim to be first, causing potential race condition
Link dialer save 'is first conn' true, so all dials claim to be first, causing potential race condition

plorenz:


router/xgress_edge_tunnel/hosted.go
```diff
@@ -564,22 +567,48 @@ func (self *HostedServiceRegistry) HandleCreateTerminatorResponse(msg *channel.M
...
+				reestablishList = append(reestablishList, terminator)
+			}
+		})
+
+		if len(reestablishList) == 0 {
+			return
+		}
+
+		// wait for verify terminator events to come in
+		time.Sleep(10 * time.Second)
```
andrewpmartinez, line 606: Waiting a static 10s in production code seems odd.

andrewpmartinez (approved):

router/xgress_edge_tunnel/hosted.go
plorenz: We have to wait some amount of time for validation events to come in from the controller and then we start trying to re-establish. In the future, we'll probably move to a different mechanism, but nothing breaks if we get the timing wrong, we just end up doing  a little extra work.

PR #3570, opened 2026-02-11
Version 2.0+ routers should not connect to controllers which do not support JWT formatted legacy sessions. Fixes #3569

### LINKED ISSUE #3569: Version 2.0+ routers should not connect to controllers which do not support JWT formatted legacy sessions

plorenz:


router/handler_ctrl/bind.go
```diff
@@ -116,12 +123,13 @@ func (self *bindHandler) BindChannel(binding channel.Binding) error {
 	binding.AddTypedReceiveHandler(newSettingsHandler(self.env))
 	binding.AddTypedReceiveHandler(newFaultHandler(self.env.GetXlinkRegistry()))
 	binding.AddTypedReceiveHandler(self.ctrlAddrChangeHandler)
+	binding.AddTypedReceiveHandler(self.clusterLeaderChangeHandler)
```
andrewpmartinez, line 126: Where did this come from? It wasn't added in this PR, so did it just exist beforehand, and we never used it?

controller/controller.go
andrewpmartinez, line 760: I thought the go convention was "error message: %w"

Its what I've been doing in all of the OpenZiti code.

plorenz: I'll start trying to internalize that

router/handler_ctrl/bind.go
plorenz: I think it was previously a copy paste error where ctrlAddrChangeHandler was added twice

PR #3572, opened 2026-02-12
Add backwards compatibility smoketest. Fixes #3571

### LINKED ISSUE #3571: Ensure 2.0 backwards compatibility with 1.6 and 1.5 using the smoketest

plorenz:


andrewpmartinez (approved):

plorenz:
> This is a security vulnerability as it allows service access tokens to be used across API Sessions. v1.1 routers don't interrogate the service access JWT to ensure it matches the current API Session, and the check in the controller here doesn't either.
> 
> Meaning someone can use just a service access token to access any service through a 1.1 router from another API Session. Service tokens have a VERY long life alone and must be compared to an API Session for proper security context, especially to make sure the API Session is still valid and matches.
> 
> Do not merge this as is.

I made some tweaks. It's now checking the submitted fingerprints against the api session for the service session. It's also limited to bind sessions, since that's the only flow that needed this tweak. Let me know if that looks better.

controller/handler_edge_ctrl/common.go
```diff
@@ -382,6 +372,38 @@ func (self *baseSessionRequestContext) loadFromBolt(sessionToken string, apiSess
...
+	if self.err != nil || !found {
+		if self.err == nil {
+			self.err = InvalidApiSessionError{}
+		}
+		logrus.
+			WithField("sessionId", self.session.Id).
+			WithField("operation", self.handler.Label()).
+			WithField("apiSessionFingerprints", apiSessionCertFingerprints).
+			WithField("clientFingerprints", fingerprints).
+			Error("matching fingerprint not found for connect")
```
andrewpmartinez: The API Session id lookup proves that the `session` originated from a still-valid `apiSession`, but not that the connecting request came from someone with a client holding that 'apiSession'. 

The fingerprint check attempts to resolve this, assuming that if the fingerprints from the certs used by the client to connect to the router match, they must be holding the API session.

That is not strictly true.

Two SDKs using the same certificate/pk and thus the same identity, but different API sessions, could connect and use each other's tokens.  This can happen in a few ways that I wouldn't recommend, but we have seen in the wild.

The result is still the same, cross-use of session tokens. Which is a problem because TOTP and posture checks are only enforced at the API Session level.

zititest/go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/openziti/foundation/v2 v2.0.87
 	github.com/openziti/identity v1.0.125
 	github.com/openziti/metrics v1.4.3
-	github.com/openziti/sdk-golang v1.4.2
```
andrewpmartinez: pseudo-version dep. Should be a clean version number.

andrewpmartinez (approved):

PR #3579, opened 2026-02-13
Add get/set db version commands

plorenz:


ziti/cmd/ops/database/db_version.go
qrkourier, line 1: It'd be useful and safe if this command could figure out whether the path arg represents a bolt db, raft db, or gzipped snapshot file produced by the snapshot command.

I'd started going there with https://github.com/openziti/ziti/pull/3468, but didn't finish writing tests.

plorenz: We couldn't get anything useful out of a raft database. I looked at the gzipped snapshot, but you can't open a bbolt db from a stream, at least not currently, you have to unzip first, so I'm not going to worry about that yet.

qrkourier: @plorenz, that makes sense about the g(unzip) and it's less important though nice to have. Importantly, I found the new command works with Raft's datastore replica file as it is, so the only enhancement that comes to mind is to accept the raft dir as an arg and look for the datastore version in the standard replica filename, `ctrl-ha.db`, which doesn't appear to be configurable in today's Ziti.

```bash
❯ ~/go/bin/ziti ops db get-db-version /tmp/ctrl.db
edge database version: 44

❯ ~/go/bin/ziti ops db get-db-version /tmp/ctrl-ha.db
edge database version: 44
```

router/state/certchecker.go
qrkourier, line 1: Are these changes related to the new datastore version command, or should this branch be based on another one other than `main`?

plorenz: It's based on the link-id-... branch with the test fixes. I've reset the base branch to clarify :+1:

ziti/cmd/ops/database/db_version.go
plorenz: Ah, gotchya. The ctrl-ha.db is exactly the same as the standard db. It only a question of whether the mutations go through raft before being applied to the db, or being applied directly/immediately.

PR #3592, opened 2026-02-17
Ctrl chan controller dials

plorenz:
Adds an opt-in beta feature where the controller can dial routers to establish control channels, reversing the traditional direction. This is useful when controllers are behind NAT/firewalls but can reach routers. The feature is disabled by default.


controller/db/router_store.go
```diff
@@ -25,21 +25,23 @@ import (
...
-	Cost        uint16       `json:"cost"`
-	NoTraversal bool         `json:"noTraversal"`
-	Disabled    bool         `json:"disabled"`
-	Interfaces  []*Interface `json:"interfaces"`
+	Name              string       `json:"name"`
+	Fingerprint       *string      `json:"fingerprint"`
+	Cost              uint16       `json:"cost"`
+	NoTraversal       bool         `json:"noTraversal"`
+	Disabled          bool         `json:"disabled"`
+	CtrlChanListeners map[string][]string `json:"ctrlChanListeners"`
```
andrewpmartinez, line 43: go fmt

controller/handler_ctrl/dialer.go
```diff
@@ -0,0 +1,501 @@
...
+	log := pfxlog.Logger().WithField("component", "ctrlDialer")
+	log.WithField("dialDelay", self.config.DialDelay).
+		WithField("groups", self.config.Groups).
+		Info("starting ctrl channel dialer")
+
+	pool, err := goroutines.NewPool(goroutines.PoolConfig{
+		QueueSize:  32,
+		MinWorkers: 0,
+		MaxWorkers: 10,
+		IdleTime:   30 * time.Second,
```
andrewpmartinez: No external config for this?

controller/handler_ctrl/dialer.go
```diff
@@ -0,0 +1,501 @@
...
+			if state.status == statusConnected || state.status == statusDialing {
+				continue
+			}
+			// already in NeedsDial, leave it on the heap
+			continue
+		}
+
+		state := &routerDialState{
+			routerId: router.Id,
+			address:  addresses[0],
```
andrewpmartinez: Multiple addresses provided, but we only ever take whatever lands at index 0?

Also, it looks like index 0 is populated back through to a map[] in the router model, which has non-deterministic ordering.

controller/handler_ctrl/dial_events.go
```diff
@@ -0,0 +1,190 @@
...
+	if !exists {
+		state = &routerDialState{
+			routerId: self.routerId,
+			address:  addresses[0],
+			nextDial: time.Now(),
+		}
+		d.states[self.routerId] = state
+		heap.Push(&d.retryQueue, state)
+	}
+}
```
andrewpmartinez, line 92: Wouldn't we also need to check for address changes here and update anything waiting to dial?

controller/handler_ctrl/dial_events.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+		state = &routerDialState{
+			routerId: self.routerId,
+			address:  addresses[0],
+		}
+		d.states[self.routerId] = state
+	}
+
+	state.connectionLost(d.config)
+	heap.Push(&d.retryQueue, state)
```
andrewpmartinez, line 54: If a scan puts a needs dial into the heap, and this also unconditionally pushes, there is a duplicate, right? I don't know if that matters.

tests/edge_router_test.go
```diff
@@ -91,6 +91,42 @@ func Test_EdgeRouter(t *testing.T) {
 		ctx.Req.False(stringz.Contains(list, role5))
 	})
 
+	t.Run("ctrlChanListeners can be created with an empty list", func(t *testing.T) {
```
andrewpmartinez, line 176: I would really prefer if we started using tests that make use of the ManagementHelperClient and ClientHelperClient exposed via ctx.NewEdgeClientApi, ctx.NewEdgeManagementApi, where possible. I haven't made any affordances for the Fabric API, but portions of this appear to be creating edge routers, which are on the edge management api.

CHANGELOG.md
```diff
@@ -745,6 +746,66 @@ established. This allows hosting applications to identify which identity initiat
...
+## Controller-Initiated Control Channel Dials (BETA)
+
+Controllers can now dial routers to establish control channels. Previously, routers were solely
+responsible for dialing controllers. This is useful in deployments where controllers are behind
+firewalls and cannot be reached by all routers, but the controllers can reach the routers.
+
+### Router Configuration
+
+Routers can configure one or more control channel listeners. Each listener specifies a bind address,
+an advertise address (reported to the controller), and optional groups for matching.
```
andrewpmartinez, line 816: "reported to the controller" or "configured on the controller"

If the router can't dial the controller, the only way this works is if the router is configured w/ a listener and then the controller is configured by an admin w/ the listener's address.


The wording is confusing and replicated below as well.

CHANGELOG.md
```diff
@@ -745,6 +746,66 @@ established. This allows hosting applications to identify which identity initiat
...
+ctrl:
+  listeners:
+    - bind: tls://0.0.0.0:6262
+      advertise: tls://router.example.com:6262
+      groups:
+        - default
+```
+
+The advertise address is stored in the router's `ctrlChanListeners` model field and reported to
+the controller. Groups default to `["default"]` if not specified.
```
andrewpmartinez: What happens if the controller is configured to connect to a router at address `A` and, after connecting, the router reports only address `B`, but that is an old address. So the controller now has a connected router, but any reconnections will be attempted to `B` and fail?

Realistically, the controller needs to be the source of truth. Not the router.

CHANGELOG.md
```diff
@@ -745,6 +746,66 @@ established. This allows hosting applications to identify which identity initiat
...
+Routers can configure one or more control channel listeners. Each listener specifies a bind address,
+an advertise address (reported to the controller), and optional groups for matching.
+
+```yaml
+ctrl:
+  listeners:
+    - bind: tls://0.0.0.0:6262
+      advertise: tls://router.example.com:6262
+      groups:
+        - default
```
andrewpmartinez, line 824: Why does the router report what group it is in? I figure that would be controller side only?

andrewpmartinez:
I would expect a config file section w/ comments in ctrl.with.edge.yml and ctrl.yml in /etc

controller/handler_ctrl/dial_events.go
plorenz: It shouldn't matter. If it's in the queue twice, the first time through it will get evaluated, and if ready for dial, will dial, which will put it in a dialing state. The second time through, if it's already dialing, it won't re-dial.

CHANGELOG.md
plorenz: The controller and router groups have to match, so they're defined on both sides.

controller/db/router_store.go
plorenz: fixed

controller/handler_ctrl/dialer.go
plorenz: fixed

controller/handler_ctrl/dialer.go
plorenz: fixed

controller/handler_ctrl/dial_state_test.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func TestDialState_ConnectionLost_NormalDisconnect(t *testing.T) {
+	cfg := testConfig()
+	state := &routerDialState{
+		routerId:    "test-router",
+		status:      statusConnected,
+		connectedAt: time.Now().Add(-time.Minute), // connected a minute ago
+		retryDelay:  30 * time.Second,             // had some previous backoff
+	}
```
andrewpmartinez, line 117: Why do some of these set `addresses: testAddresses` but other don't?

controller/handler_ctrl/dial_state.go
```diff
@@ -0,0 +1,110 @@
...
+	status       routerDialStatus
+	retryDelay   time.Duration
+	nextDial     time.Time
+	dialAttempts uint32
+	connectedAt  time.Time
+	dialActive   atomic.Bool
+}
+
+func (self *routerDialState) currentAddress() string {
+	return self.addresses[self.addrIndex%len(self.addresses)]
```
andrewpmartinez, line 61: fails on nil addresses

controller/handler_ctrl/dial_state.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (self *routerDialState) nextAddress() {
+	self.addrIndex = (self.addrIndex + 1) % len(self.addresses)
```
andrewpmartinez, line 65: fails on nil addresses

common/inspect/ctrl_dialer_inspections.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// CtrlDialerConfigDetail contains the active configuration of the ctrl dialer.
+type CtrlDialerConfigDetail struct {
+	Groups             []string `json:"groups"`
+	MinRetryInterval   string   `json:"minRetryInterval"`
+	MaxRetryInterval   string   `json:"maxRetryInterval"`
+	RetryBackoffFactor float64  `json:"retryBackoffFactor"`
+	FastFailureWindow  string   `json:"fastFailureWindow"`
+}
```
andrewpmartinez, line 39: Should this have `QueueSize` and `MaxWorkers`?

andrewpmartinez (approved):

controller/handler_ctrl/dial_events.go
plorenz: fixed

controller/handler_ctrl/dial_state.go
plorenz: we never assign an empty or nil slice to addresses. There wouldn't be a point, since there's nothing to dial

CHANGELOG.md
plorenz: updated doc to make this cleared. Also changed cli flag to reinforce this.

CHANGELOG.md
plorenz: updated doc to make this cleared. Also changed cli flag to reinforce this.

controller/handler_ctrl/dial_state_test.go
plorenz: fixed

controller/handler_ctrl/dial_state.go
plorenz: we never assign an empty or nil slice to addresses. There wouldn't be a point, since there's nothing to dial

tests/edge_router_test.go
plorenz: updated tests

common/inspect/ctrl_dialer_inspections.go
plorenz: fixed

PR #3598, opened 2026-02-21
Fix the control channel header conflicts with channel headers. Use single strategy for sharing capabilities with bit mask.

plorenz:


common/capabilities/consts.go
```diff
@@ -39,4 +39,8 @@ const (
 
 	// ControllerGroupedCtrlChan indicates support for grouped-underlay control channels
 	ControllerGroupedCtrlChan int = 5
+
+	// RouterMultiChannel indicates the router uses new (1000+) ControlHeaders IDs
+	// and supports multi-underlay control channels
+	RouterMultiChannel int = 1
```
andrewpmartinez, line 52: So does the prefix indicate who sends the header? If so, we should probably separate them and sort them based on who sends them. Based on the comments here, I have to intuit who sends and who reads.

common/ctrlchan/channel.go
```diff
@@ -194,13 +194,8 @@ func (self *BaseCtrlChannel) GetMessageSource(underlay channel.Underlay) channel
...
 
-func (self *BaseCtrlChannel) HandleTxFailed(_ channel.Underlay, sendable channel.Sendable) bool {
-	select {
-	case self.defaultMsgChan <- sendable:
-		return true
-	default:
-		return false
-	}
+func (self *BaseCtrlChannel) HandleTxFailed(_ channel.Underlay, _ channel.Sendable) bool {
+	return false
```
andrewpmartinez, line 201: Always false? This looks like it needs a comment or something.

plorenz: Added comment

common/capabilities/consts.go
plorenz: Separated the out for clarity :+1:

common/ctrlchan/channel.go
```diff
@@ -317,8 +312,8 @@ func (self *DialCtrlChannel) HandleUnderlayClose(ch channel.MultiChannel, underl
 }
 
 func (self *DialCtrlChannel) HandleUnderlayAccepted(ch channel.MultiChannel, underlay channel.Underlay) {
-	if underlayType := channel.GetUnderlayType(underlay); underlayType == ChannelTypeDefault {
-		self.hasDefaultChan.Store(true)
+	if channel.GetUnderlayType(underlay) == ChannelTypeHighPriority {
```
andrewpmartinez, line 318: Should this be encapsulated within `underlay`? underly.Type() or the like. I know it is optional, but if every underlay is untyped ("") or typed ("something") it would abstract where/what that header is.

plorenz: I thought about it, but there are a bunch of Underlay implementations and it felt clunky repeating the implementation over and over again

Issue #3597, opened 2026-02-20
Enable OIDC API by default


andrewpmartinez:
Until now, the OIDC capabilities for OpenZiti have been optional. For 2.0 they are now required and considered the default. The previous authentication API is now deprecated and will be removed in a later version, possibly v3.0. The legacy auth API remains in place to support older SDKs that do not support HA.

If the Edge OIDC API is not configured it will be auto configured by default using the following logic:

- The Edge Client API that matches the edge.api.address field will automatically have the OIDC API enabled on it, a warning log message is printed
- If no Edge Client APIs match the edge.api.address and there is only 1 Edge Client API, it will be added to that, a warning log message is printed
- If there are multiple Edge Client APIs and none of them match edge.api.address, the OIDC API is added to all of them, a warning log message is printed

PR #3605, opened 2026-02-25
fix.openziti.ziti.3597.OIDC.by.default

andrewpmartinez:
fixes https://github.com/openziti/ziti/issues/3597, enable OIDC by default

adds ability to have different sets of environment configs for tests
- adds ConfigSet struct as a configuration device for integration tests
  - allows entire environments to be defined as needed
  - original ats config set at "default-ats"
  - tests that do not specify a config set, use "default-ats" as before
- standardizes configuratoin location, naming, etc.
- adds README.md for the above
- updates testContext to now be config set aware

### DIFF
--- a/controller/env/routers.go
+++ b/controller/env/routers.go
@@ -28,6 +28,10 @@ type ApiRouterMiddleware interface {
 	AddMiddleware(ae *AppEnv)
 }
 
+type ApiRouterShutdown interface {
+	Shutdown(ae *AppEnv)
+}
+
 func AddRouter(rf ApiRouter) {
 	routers = append(routers, rf)
 }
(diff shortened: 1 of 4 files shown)


tests/configsets.go
```diff
@@ -0,0 +1,36 @@
...
+
+// DefaultATS is the standard full-stack config set used by the majority of the
+// integration test suite. It covers the controller, edge router, tunneler router,
+// transit router, and the fabric-only router pair used by link tests.
+var DefaultATS = ConfigSet{
+	Name:           "default-ats",
+	CtrlConfig:     "testdata/configs/default-ats/ctrl.yml",
+	EdgeRouter:     "testdata/configs/default-ats/edge-router.yml",
+	TunnelerRouter: "testdata/configs/default-ats/tunneler-router.yml",
+	TransitRouter:  "testdata/configs/default-ats/transit-router.yml",
```
dovholuknf, line 22: If i never saw another "transit" or "fabric" reference to a router. i'd be happy. i also see no distinction between those two imo

PR #3615, opened 2026-02-26
fixes #3609 sdk terminator race fixes and preferred-leader raft

### LINKED ISSUE #3609: Stabilize terminator creation test for 2.0

plorenz:
- removes legacy v1 terminator code path; all terminators now use v2 flow
    - refactors edgeTerminator.close() to decouple SDK notification from control plane notification
    - adds pending SDK close notification queue with retry when channel is busy
    - adds post-create inspect mechanism that verifies SDK still holds the bind after terminator creation
    - queues second post-create inspect when establishment takes >30s to catch SDK timeout races
    - detects and discards stale reordered binds on the same connection by comparing connIds
    - re-establishes replacement terminators when a delete/create race is detected
    - eliminates IsEntityPresent pre-filter in removeTerminatorsHandler to prevent raft ordering races
    - fixes ValidateTerminators to query identities from the correct manager with the correct filter field
    - adds postCreate flag to ValidateTerminatorsV2Request so routers skip redundant SDK inspect
    - returns retry-later (nil result) from router validation when inspect is temporarily unavailable
    - blocks SyncAllSubscribers until completion and guards RouterDataModel replacement with in-progress flag
    - fixes InheritLocalData to enable service access tracking for all subscribed identities
    - adds Services.Has check in GetServiceAccessPolicies to prevent false policy grants
    - validates policy-to-identity associations in ValidateServicePolicies
    - adds `ziti agent tunnel dump-sdk` command for SDK context inspection via IPC agent
    - adds `ziti fabric inspect sdk` command to query SDK context through routers
    - fixes --expected-per-host CLI flag binding in validate terminators command
    - changes bind-access-lost retry hint from NotRetriable to RetryStartOver
    - moves trace route response and xgress close handling off channel handler goroutine
    - fixes listTerminators test helper to URL-encode filter parameter
    - improves sdk-hosting-test validation resilience with login and query retries
    - adds terminator_create_flow.md documenting the full lifecycle across SDK, router, and controller
    - adds detailed logging for data model sync, service access tracking, and subscriber change detection

### DIFF
--- a/common/subscriber.go
+++ b/common/subscriber.go
@@ -18,7 +18,6 @@ package common
 
 import (
 	"encoding/json"
-	"fmt"
 	"sync"
 
 	"github.com/google/go-cmp/cmp"
@@ -348,16 +347,6 @@ func (self *IdentitySubscription) checkForChanges(rdm *RouterDataModel) {
 	for svcId, service := range oldServices {
 		newService, ok := newServices[svcId]
 		if !ok {
-			newIdentity.WithLock(func() {
-				serviceAccessKeys := make([]string, 0, len(newIdentity.ServiceAccess))
-				for k, v := range newIdentity.ServiceAccess {
-					serviceAccessKeys = append(serviceAccessKeys, fmt.Sprintf("%s(d=%d,b=%d)", k, v.DialPoliciesCount, v.BindPoliciesCount))
-				}
-				servicePolicyKeys := make([]string, 0, len(newIdentity.ServicePolicies))
-				for k := range newIdentity.ServicePolicies {
-					servicePolicyKeys = append(servicePolicyKeys, k)
-				}
-			})
 			self.notifyServiceChange(state, service, service, ServiceAccessLostEvent)
 		} else if !service.Equals(newService) {
 			self.notifyServiceChange(state, service, newService, ServiceUpdatedEvent)
--- a/router/state/manager.go
+++ b/router/state/manager.go
@@ -918,6 +918,9 @@ func (self *ManagerImpl) pubKeyLookup(token *jwt.Token) (any, error) {
 // RouterDataModel returns the current router data model containing
 // network topology and policy information.
 func (self *ManagerImpl) RouterDataModel() *common.RouterDataModel {
+	// If a router data model replace is in progress, we need to wait for it to be fully
+	// initialized before we get it, otherwise there's the possibility of race conditions
+	// Should happen rarely, so a sleep is the simplest solution
 	for self.rmdReplaceInProgress.Load() {
 		time.Sleep(100 * time.Millisecond)
 	}
--- a/router/xgress_edge/hosted.go
+++ b/router/xgress_edge/hosted.go
@@ -460,6 +460,11 @@ func (self *addSdkCloseNotificationEvent) handle(registry *hostedServiceRegistry
 }
 
 func (self *hostedServiceRegistry) addPendingSdkCloseNotification(terminator *edgeTerminator, reason string, edgeErr *EdgeError) {
+	// We can't necessarily key by terminator as terminators can be replaced and the id can be reused.
+	// We can't use the virtual conn id, because those aren't unique across routers.
+	// If we do happen to send multiple, it should be harmless, since the second close will be ignored.
+	// We could probably use a combination of terminator id and virtual conn id.
+	// Let's try that next time we run a longer test cycle
 	self.notifyCloseSet[idgen.MustNewUUIDString()] = &pendingSdkCloseNotification{
 		terminator: terminator,
 		reason:     reason,
@@ -471,7 +476,7 @@ func (self *hostedServiceRegistry) ensureSdkCloseSent(terminator *edgeTerminator
 	self.ensureSdkCloseSentInner(terminator, reason, edgeErr, false)
 }
 
-func (self *hostedServiceRegistry) ensureSdkCloseSentInner(terminator *edgeTerminator, reason string, edgeErr *EdgeError, queueDirect bool) {
+func (self *hostedServiceRegistry) ensureSdkCloseSentInner(terminator *edgeTerminator, reason string, edgeErr *EdgeError, inEventLoop bool) {
 	if terminator.GetChannel().IsClosed() {
 		return
 	}
@@ -481,7 +486,7 @@ func (self *hostedServiceRegistry) ensureSdkCloseSentInner(terminator *edgeTermi
 	}
 	queued, _ := terminator.GetDefaultSender().TrySend(closeMsg)
 	if !queued {
-		if queueDirect {
+		if inEventLoop {
 			self.addPendingSdkCloseNotification(terminator, reason, edgeErr)
 		} else {
 			self.queueSdkCloseNotification(terminator, reason, edgeErr)
@@ -977,7 +982,8 @@ func (self *hostedServiceRegistry) handleSdkReturnedInvalid(terminator *edgeTerm
 	select {
 	case result := <-event.resultC:
 		return result
-	case <-time.After(250 * time.Millisecond):
+	case <-time.After(250 * time.Millisecond): // Failsafe in case the hosting registry event loop gets too busy
+		//                                     // If we bail out, the notify process will be retried
 		if !self.terminators.Has(terminator.terminatorId) {
 			return invalidTerminatorRemoveResult{
 				existed: false,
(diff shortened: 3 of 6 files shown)


common/subscriber.go
```diff
@@ -347,6 +348,26 @@ func (self *IdentitySubscription) checkForChanges(rdm *RouterDataModel) {
...
+				log.WithField("lostServiceId", svcId).
+					WithField("lostServiceName", service.GetName()).
+					WithField("trackingEnabled", newIdentity.serviceAccessTrackingEnabled.Load()).
+					WithField("serviceAccessCount", len(newIdentity.ServiceAccess)).
+					WithField("servicePolicyCount", len(newIdentity.ServicePolicies)).
+					WithField("serviceAccess", serviceAccessKeys).
+					WithField("servicePolicies", servicePolicyKeys).
+					WithField("oldServiceCount", len(oldServices)).
+					WithField("newServiceCount", len(newServices)).
+					Warn("service access lost detected during checkForChanges")
```
andrewpmartinez: Is this really needed as a warning? It seems like this would happen normally.

router/state/manager.go
```diff
@@ -917,6 +918,9 @@ func (self *ManagerImpl) pubKeyLookup(token *jwt.Token) (any, error) {
 // RouterDataModel returns the current router data model containing
 // network topology and policy information.
 func (self *ManagerImpl) RouterDataModel() *common.RouterDataModel {
+	for self.rmdReplaceInProgress.Load() {
+		time.Sleep(100 * time.Millisecond)
```
andrewpmartinez, line 925: This seems awfully hacky.

common/subscriber.go
plorenz: Debug code from when I was trying to track down a bug where identities would temporarily lose access to services, even though the model hadn't changed. Removing.

router/state/manager.go
plorenz: It fixes a race condition where we get access to a partially initialized rdm instance. I could add more complicated machinery with a notify channel if the replace is in progress, but it's something that should happen rarely and the logic would be harder to follow and get right. I felt like this was the best solution given those circumstances. Do you have a different approach that doesn't add a bunch of complexity?

router/xgress_edge/hosted.go
```diff
@@ -740,18 +967,17 @@ func (self *hostedServiceRegistry) checkForExistingListenerId(terminator *edgeTe
...
 		resultC:    make(chan invalidTerminatorRemoveResult, 1),
-		notifyCtrl: notifyCtrl,
 	}
 	self.queue(event)
 
 	select {
 	case result := <-event.resultC:
 		return result
-	case <-time.After(100 * time.Millisecond):
+	case <-time.After(250 * time.Millisecond):
```
andrewpmartinez: Magical. I don't even know why this was increased.

router/xgress_edge/hosted.go
```diff
@@ -794,7 +1020,7 @@ func (self *inspectTerminatorsEvent) handle(registry *hostedServiceRegistry) {
 			Precedence:      terminator.precedence.String(),
 			AssignIds:       terminator.assignIds,
 			UseSdkXgress:    terminator.useSdkXgress,
-			V2:              terminator.v2,
+			V2:              true,
```
andrewpmartinez: To be clear, the terminators aren't v1/v2 its what the router supports? I am worried about mixed networks.

router/xgress_edge/listener.go
```diff
@@ -215,6 +223,114 @@ func (listener *listener) submitResponse(resultCh chan *sdkCircuitResult, result
...
 }
 
+type sdkContextResult struct {
+	identityId string
+	connId     string
+	err        error
+	context    any
+}
+
+func (listener *listener) getSdkContext(identityId string) any {
```
andrewpmartinez, line 233: Why not an interface that can do generic (for example, []byte) marshalling/unmarshalling? I understand this is probably a re-applied pattern, but I am curious.

router/xgress_edge/listener.go
```diff
@@ -335,12 +451,14 @@ func (self *edgeClientConn) handleBindAccessLost(service *common.IdentityService
 	for _, terminator := range terminators {
 		log.WithField("terminatorId", terminator.terminatorId).
 			Info("bind access to service, closing terminator")
-		err := &EdgeError{
+		edgeErr := &EdgeError{
 			Message:   "bind access lost",
 			Code:      sdkedge.ErrorCodeAccessDenied,
-			RetryHint: sdkedge.RetryNotRetriable,
+			RetryHint: sdkedge.RetryStartOver, // switch back to start over once timing issues are resolved
```
andrewpmartinez, line 457: Why the swap from "not retriable" to "start over"? Does start over mean "verify you have access again"?

router/xgress_router/factory.go
```diff
@@ -30,9 +30,11 @@ type Dialer interface {
 	IsTerminatorValid(id string, destination string) bool
 }
 
+type InspectResult int
```
andrewpmartinez: Was this ever used?

router/xgress_edge/hosted.go
```diff
@@ -421,6 +435,217 @@ func (self *hostedServiceRegistry) queueRemoveTerminatorUnchecked(terminator *ed
...
+		registry.notifyCloseSet[idgen.MustNewUUIDString()] = self.notification
+	}
+}
+
+func (self *hostedServiceRegistry) addPendingSdkCloseNotification(terminator *edgeTerminator, reason string, edgeErr *EdgeError) {
+	self.notifyCloseSet[idgen.MustNewUUIDString()] = &pendingSdkCloseNotification{
+		terminator: terminator,
+		reason:     reason,
+		edgeErr:    edgeErr,
+	}
```
andrewpmartinez, line 472: If this is called twice, does it just add two close notifies, as the UUIDs will always be different?

router/xgress_edge/hosted.go
plorenz: v1/v2 is which versions of the create terminator operation the controller supports, and since these are v2 routers which must talk to v2 controllers, we know we'll always be used the v2 create terminator operation

router/xgress_edge/hosted.go
```diff
@@ -421,6 +435,217 @@ func (self *hostedServiceRegistry) queueRemoveTerminatorUnchecked(terminator *ed
...
+}
+
+func (self *hostedServiceRegistry) ensureSdkCloseSent(terminator *edgeTerminator, reason string, edgeErr *EdgeError) {
+	self.ensureSdkCloseSentInner(terminator, reason, edgeErr, false)
+}
+
+func (self *hostedServiceRegistry) ensureSdkCloseSentInner(terminator *edgeTerminator, reason string, edgeErr *EdgeError, queueDirect bool) {
```
andrewpmartinez: `queueDirect` is a bit oblique as a naive reader. It toggles between queue and addPending.

router/state/manager.go
andrewpmartinez: That seems like a good comment for that line then.

router/xgress_edge/listener.go
plorenz: It's called from the `listener.Inspect() any` API, so it's obeying that contract. The assumption is that the types are marshallable to JSON. I don't think there's an interface for that. The inspect system has grown somewhat organically over time and could probably use an overhaul

router/xgress_edge/listener.go
plorenz: That bug that caused service loss meant that if we sent NoRetriable, the sdk would never retry. Because the sdk gets service updates from another path, it never thought it lost access. So unless you're getting access updates from the same path, there are coherency issues with sending 'do not retry'.

Yes, start over means redo access checks.

router/xgress_router/factory.go
plorenz: Removed

router/xgress_edge/hosted.go
plorenz: It does. However, we can't necessarily key by terminator as terminators can be replaced and the id can be re-used. We can't use the virtual conn id, because those aren't unique across routers. If we do happen to send multiple, it should be harmless, since the second close will be ignored. We could probably use a combination of terminator id and virtual conn id. I'll add a comment

router/xgress_edge/hosted.go
plorenz: I renamed it to `inEventLoop`

router/state/manager.go
plorenz: done

common/subscriber.go
plorenz: done

router/xgress_edge/hosted.go
plorenz: added a comment

router/xgress_edge/hosted.go
plorenz: done

PR #3631, opened 2026-03-03
Fix login certificate caching when changing controllers

dovholuknf:
- Only use cached certificates when connecting to the same controller
  - Include server certificate in file-based auth cert pool (fixes TLS verification issues surfaced with test)
  - Add tests for certificate caching and controller switching
  - also fixes verify ext-jwt-signer oidc requiring login (it shouldn't)

### DIFF
--- a/tests/cli_tests/cli_test.go
+++ b/tests/cli_tests/cli_test.go
@@ -223,10 +223,10 @@ func Test_CLI_Test_Suite(t *testing.T) {
 		log.Fatalf("controllerUnderTest start failed: %v", cutStartOverZitiErr)
 	}
 
-	//xxx testState.cliTestsOverZiti(t, zitiPath)
+	testState.cliTestsOverZiti(t, zitiPath)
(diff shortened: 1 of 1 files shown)


ziti/cmd/edge/login.go
```diff
@@ -287,8 +301,13 @@ func (o *LoginOptions) Run() error {
 	}
 
 	if ctrlUrl.Path == "" {
-		if o.FileCertCreds != nil && o.FileCertCreds.CaPool != nil {
-			host = util.EdgeControllerGetManagementApiBasePathWithPool(host, o.FileCertCreds.CaPool, &httpClient)
+		fullCaPool, caPoolErr := o.GetCaPool()
+		if caPoolErr != nil {
+			return caPoolErr
+		}
+
```
plorenz: fullCaPool should always be non-nil here

plorenz (approved):

PR #3632, opened 2026-03-03
Adaptive rate limiter tracker should use success rate instead of queue position

plorenz:


controller/command/rate_limiter.go
```diff
@@ -457,17 +538,25 @@ func NewAdaptiveRateLimitTracker(config AdaptiveRateLimiterConfig, registry metr
 }
 
 type adaptiveRateLimitTracker struct {
```
andrewpmartinez, line 601: A kind word here for the souls that need direction. A comment or two, sir, please.

controller/command/rate_limiter.go
```diff
@@ -419,18 +426,92 @@ func WasRateLimited(err error) bool {
...
+func (self *AdaptiveRateLimitTrackerConfig) SetDefaults() {
+	self.AdaptiveRateLimiterConfig.SetDefaults()
+	self.SuccessThreshold = DefaultAdaptiveRateLimiterSuccessThreshold
+	self.IncreaseFactor = DefaultAdaptiveRateLimiterIncreaseFactor
+	self.DecreaseFactor = DefaultAdaptiveRateLimiterDecreaseFactor
+	self.IncreaseCheckInterval = DefaultAdaptiveRateLimiterIncreaseCheckInterval
+	self.DecreaseCheckInterval = DefaultAdaptiveRateLimiterDecreaseCheckInterval
+}
+
+func (cfg *AdaptiveRateLimitTrackerConfig) Load(cfgmap map[interface{}]interface{}) error {
```
andrewpmartinez, line 473: Is there a reason we don't call SetDefaults () here? In the situation where a partial config is provided? Some of these values are used for division/division-like operators and will panic.

controller/command/rate_limiter.go
```diff
@@ -419,18 +426,92 @@ func WasRateLimited(err error) bool {
...
+}
+
+func (self *AdaptiveRateLimitTrackerConfig) SetDefaults() {
+	self.AdaptiveRateLimiterConfig.SetDefaults()
+	self.SuccessThreshold = DefaultAdaptiveRateLimiterSuccessThreshold
+	self.IncreaseFactor = DefaultAdaptiveRateLimiterIncreaseFactor
+	self.DecreaseFactor = DefaultAdaptiveRateLimiterDecreaseFactor
+	self.IncreaseCheckInterval = DefaultAdaptiveRateLimiterIncreaseCheckInterval
+	self.DecreaseCheckInterval = DefaultAdaptiveRateLimiterDecreaseCheckInterval
+}
```
andrewpmartinez, line 470: These two don't appear to be loaded in the config section. Intentional?

controller/command/rate_limiter.go
plorenz: SetDefaults is called in all the places where we call Load(). The reason they're not combined is that some of those call-sites have overrides for the defaults that need to be set before loading the config

controller/command/rate_limiter.go
plorenz: ah, good catch. Fixed

controller/command/rate_limiter.go
plorenz: added more doc

PR #3651, opened 2026-03-06
Unroute removed transit nodes during circuit reroute. Fixes #3650

### LINKED ISSUE #3650: Reroute doesn't proactively clean up orphaned route entries
The idle circuit detection cleans them up, but we should clean them up as soon as the new path has taken effect.

plorenz:
- Adds unrouteRemovedPathNodes to send Unroute to old path nodes not
  present in the new path after a reroute
- Calls it from rerouteCircuit and smartReroute after successfully
  routing the new path
- Uses GetConnectedRouter to get the current router object rather than
  stale path pointers
- Fixes orphaned forwarder entries on transit routers replaced during
  reroute, which caused circuit validation failures

### DIFF
--- a/.github/workflows/backward-compatibility.yml
+++ b/.github/workflows/backward-compatibility.yml
@@ -1,7 +1,6 @@
 name: Backward Compatibility Test
 
 on:
-  pull_request:
   workflow_dispatch:
 
 # cancel older, redundant runs of same workflow on same branch


controller/network/network.go
```diff
@@ -1127,6 +1130,7 @@ func (network *Network) smartReroute(circuit *model.Circuit, cq *model.Path, dea
 	if circuit.Rerouting.CompareAndSwap(false, true) {
 		defer circuit.Rerouting.Store(false)
 
+		oldPath := circuit.Path
 		circuit.Path = cq
 		circuit.UpdatedAt = time.Now()
```
andrewpmartinez, line 1136: Does it matter that `circuit.Path` is overwritten here and never reverted if `retry` is set to true?

controller/network/network.go
```diff
@@ -1108,6 +1109,8 @@ func (network *Network) rerouteCircuit(circuit *model.Circuit, deadline time.Tim
 				}
```
andrewpmartinez, line 1110: seems odd here that we erred in sending a routing message, but keep going.

controller/network/network.go
plorenz: Yes. If the re-route fails, the path is in an unknown state. It may have succeeded but timed out, it maybe have partially succeeded or it may have failed. So `circuit.Path` is untrustworthy at this point, reverting it is unlikely to be beneficial.

controller/network/network.go
plorenz: Seems wrong, fixing

PR #3664, opened 2026-03-09
fixes #3658 #3648 fallback to legacy headers on dial; fix circuitId in tunnel

### LINKED ISSUE #3648: tunnel: myCopy logs router ID as circuitId, causing misleading debug output
## Description

In `tunnel/tunnel.go`, the `Run()` function starts two `myCopy` goroutines — one for each direction of the data copy. The client-to-ziti direction correctly passes `zitiConn.GetCircuitId()`, but the ziti-to-client direction passes `zitiConn.GetRouterId()`:

```go
go myCopy(clientConn, zitiConn, doneSend, halfClose, zitiConn.GetCircuitId())
go myCopy(zitiConn, clientConn, doneRecv, halfClose, zitiConn.GetRouterId())
```

Inside `myCopy()`, this value is logged under the field name `"circuitId"`. This means every "stopping pipe", "copy failed", or "half-close" log entry from the receive direction shows the **router identity** (e.g., `AZD.aVO2ob`) instead of the actual circuit ID.

## Impact

This makes it extremely difficult to debug connection issues because:

1. The router identity appears as a constant "circuit ID" across all connections through that router, making it look like a single persistent zombie circuit
2. Operators investigating connection failures waste significant time trying to track down a "zombie circuit" that doesn't exist
3. Log correlation between the two copy directions is impossible since they log different IDs for the same circuit

## Expected behavior

Both `myCopy` goroutines should log the actual circuit ID for the connection. The router identity should be logged separately if needed.

## Steps to reproduce

1. Run `ziti tunnel tproxy` with any identity
2. Make a connection to any Ziti service
3. Observe that "tunnel started" and "stopping pipe" (send direction) log the real circuit ID
4. Observe that "stopping pipe" (receive direction) logs the router identity as `circuitId`

## Environment

- ziti v2.0.0-pre2
- sdk-golang v1.4.2
- Linux tproxy mode

### LINKED ISSUE #3658: Raft cluster join fails with "hello message too big" when using long hostnames
Observed error when joining controllers to a raft cluster using `ziti agent cluster add`:

```
hello message too big. header len: 4461, body len: 16
```

The joining controller receives only an EOF on the client side; the "hello message too big" error appears in the target controller's logs.

Worked around the issue by using shorter hostnames (e.g. `127.0.0.1` instead of `ziti-controller1.127.0.0.1.sslip.io`).

Reproduced in the HA quickstart test (`quickstart/test/ha-test.sh`) by changing `--ctrl-address` from `127.0.0.1` to sslip.io-style hostnames. The test passes with short hostnames and fails with long ones.

plorenz:
- On Raft mesh dial failure (e.g. "hello message too big"), retries with SigningCertHeader stripped so only the smaller
  legacy header is sent, enabling joins against old controllers with the 4k hello limit
 - Fixes incorrect routerId passed to myCopy in tunnel (should be circuitId); resolves misleading log output
 - Updates deps: channel v4.3.8, sdk-golang v1.5.2, and several minor bumps
- **Fallback to legacy headers on dial failure, in case of size constraints. Fixes #3658**
- **Pass circuitId, not routerId to myCopy in tunnel. Fixes #3648**

### DIFF
--- a/controller/raft/mesh/mesh.go
+++ b/controller/raft/mesh/mesh.go
@@ -61,6 +61,8 @@ const (
 	RaftDisconnectType = 2052
 
 	ChannelTypeMesh = "ctrl.mesh"
+
+	RecentDialInterval = 5 * time.Minute
 )
 
 type Peer struct {
@@ -473,11 +475,10 @@ func (self *impl) GetOrConnectPeer(address string, timeout time.Duration) (*Peer
 	// If so, strip the new signing cert header so only the legacy header is sent, allowing
 	// the connection to succeed with old controllers that enforce the smaller hello limit.
 	self.lock.RLock()
-	if rec := self.dialRecords[address]; rec != nil && time.Since(rec.lastAttempt) < 5*time.Minute {
-		if rec.peerVersion == nil || func() bool {
-			hasMin, _ := rec.peerVersion.HasMinimumVersion("v2.0.0")
-			return !hasMin
-		}() {
+	if rec := self.dialRecords[address]; rec != nil && time.Since(rec.lastAttempt) < RecentDialInterval {
+		if rec.peerVersion == nil {
+			delete(headers, SigningCertHeader)
+		} else if hasMin, _ := rec.peerVersion.HasMinimumVersion("v2.0.0"); !hasMin {
 			delete(headers, SigningCertHeader)
 		}
 	}
--- a/controller/raft/raft.go
+++ b/controller/raft/raft.go
@@ -57,6 +57,11 @@ import (
 	"github.com/teris-io/shortid"
 )
 
+const (
+	LeaderCheckInterval = 5 * time.Second
+	MeshCleanupInterval = time.Minute
+)
+
 type ClusterEvent uint32
 
 func (self ClusterEvent) String() string {
@@ -816,30 +821,17 @@ func (self *Controller) eventLoop() {
 		self.handleClusterStateChange(ClusterEventIsLeaderless, eventState)
 	}
 
-	leaderTicker := time.NewTicker(5 * time.Second)
+	leaderTicker := time.NewTicker(LeaderCheckInterval)
 	defer leaderTicker.Stop()
 
-	dialCleanupTicker := time.NewTicker(time.Minute)
+	dialCleanupTicker := time.NewTicker(MeshCleanupInterval)
 	defer dialCleanupTicker.Stop()
-
-	first := false
 
 	for {
 		select {
 		case observation := <-self.clusterEvents:
 			self.processRaftObservation(observation, eventState)
 		case <-leaderTicker.C:
-			if first {
-				// delay this check because it seems like raft generates observations for leader state, so if we do this
-				// first we're going to get duplicates
-				if self.Raft.State() == raft.Leader {
-					if wasLeader := self.isLeader.Swap(true); !wasLeader {
-						self.handleClusterStateChange(ClusterEventLeadershipGained, eventState)
-					}
-				}
-				first = false
-			}
-
 			if !eventState.warningEmitted && !eventState.hasLeader && time.Since(eventState.noLeaderAt) > self.Config.WarnWhenLeaderlessFor {
 				pfxlog.Logger().WithField("timeSinceLeader", time.Since(eventState.noLeaderAt).String()).
 					Warn("cluster running without leader for longer than configured threshold")


controller/raft/raft.go
```diff
@@ -800,52 +800,55 @@ type clusterEventState struct {
...
-				self.processRaftObservation(observation, eventState)
-			case <-ticker.C:
-				if first {
-					// delay this check because it seems like raft generates observations for leader state, so if we do this
-					// first we're going to get duplicates
-					if self.Raft.State() == raft.Leader {
-						if wasLeader := self.isLeader.Swap(true); !wasLeader {
-							self.handleClusterStateChange(ClusterEventLeadershipGained, eventState)
-						}
+	first := false
```
andrewpmartinez: When is first ever not false? It seems like the block at lines 832-841 is just never hit.

controller/raft/mesh/mesh.go
```diff
@@ -461,6 +469,20 @@ func (self *impl) GetOrConnectPeer(address string, timeout time.Duration) (*Peer
...
+	// If so, strip the new signing cert header so only the legacy header is sent, allowing
+	// the connection to succeed with old controllers that enforce the smaller hello limit.
+	self.lock.RLock()
+	if rec := self.dialRecords[address]; rec != nil && time.Since(rec.lastAttempt) < 5*time.Minute {
+		if rec.peerVersion == nil || func() bool {
+			hasMin, _ := rec.peerVersion.HasMinimumVersion("v2.0.0")
+			return !hasMin
+		}() {
+			delete(headers, SigningCertHeader)
+		}
```
andrewpmartinez: Why is an IIFE being used here? It seems to serve no purpose other than to be weird-looking.

controller/raft/mesh/mesh.go
```diff
@@ -461,6 +469,20 @@ func (self *impl) GetOrConnectPeer(address string, timeout time.Duration) (*Peer
 		headerProvider.Apply(headers)
 	}
 
+	// Check if a recent dial to this address failed (e.g., hello too large for an old peer).
+	// If so, strip the new signing cert header so only the legacy header is sent, allowing
+	// the connection to succeed with old controllers that enforce the smaller hello limit.
+	self.lock.RLock()
+	if rec := self.dialRecords[address]; rec != nil && time.Since(rec.lastAttempt) < 5*time.Minute {
```
andrewpmartinez: Magic number.

controller/raft/mesh/mesh.go
```diff
@@ -523,6 +545,12 @@ func (self *impl) GetOrConnectPeer(address string, timeout time.Duration) (*Peer
 		peer.Version = versionInfo
 		peer.SigningCerts = []*x509.Certificate{underlay.Certificates()[0]}
 
+		self.lock.Lock()
+		if rec := self.dialRecords[address]; rec != nil {
+			rec.peerVersion = versionInfo
```
andrewpmartinez, line 551: If rec is a pointer, is setting peerVersion here before we hit a success case (no error) safe?

controller/raft/raft.go
```diff
@@ -800,52 +800,55 @@ type clusterEventState struct {
...
-		}
+	if eventState.hasLeader {
+		self.handleClusterStateChange(ClusterEventHasLeader, eventState)
+	} else {
+		self.handleClusterStateChange(ClusterEventIsLeaderless, eventState)
+	}
 
-		ticker := time.NewTicker(time.Second * 5)
-		defer ticker.Stop()
+	leaderTicker := time.NewTicker(5 * time.Second)
```
andrewpmartinez: Magic number.

controller/raft/raft.go
```diff
@@ -800,52 +800,55 @@ type clusterEventState struct {
...
+	defer leaderTicker.Stop()
 
-		first := false
+	dialCleanupTicker := time.NewTicker(time.Minute)
```
andrewpmartinez: Magic number.

controller/raft/raft.go
plorenz: This has been wrong since I first added it, and it doesn't seem to be causing any problems, so I think I'll just remove the block

controller/raft/raft.go
```diff
@@ -800,52 +800,55 @@ type clusterEventState struct {
...
 
-				if !eventState.warningEmitted && !eventState.hasLeader && time.Since(eventState.noLeaderAt) > self.Config.WarnWhenLeaderlessFor {
-					pfxlog.Logger().WithField("timeSinceLeader", time.Since(eventState.noLeaderAt).String()).
-						Warn("cluster running without leader for longer than configured threshold")
-					eventState.warningEmitted = true
-				}
+			if !eventState.warningEmitted && !eventState.hasLeader && time.Since(eventState.noLeaderAt) > self.Config.WarnWhenLeaderlessFor {
+				pfxlog.Logger().WithField("timeSinceLeader", time.Since(eventState.noLeaderAt).String()).
+					Warn("cluster running without leader for longer than configured threshold")
+				eventState.warningEmitted = true
```
andrewpmartinez, line 838: What resets this if ever? What if no leader happens, then warning, then comes back, then no leader happens again?

controller/raft/mesh/mesh.go
plorenz: I wasn't happy with it either way. With it, it's a single if, without it, I have to check peerVersion == nil and then do a separate if for the HasMinVersion, because that function has a multi-return. This is probably harder to read, though. I'll switch it back.

controller/raft/mesh/mesh.go
plorenz: If you mean safe from a concurrency standpoint, all accesses are protected by a lock.
If you mean safe from a 'is this a successful connection' case, then we're just plucking the version number out. The connection has been successful enough that we've been able to exchange headers, which is all we need to establish the version. This whole thing is for a very small window, when we're upgrading a cluster from pre-2.0 to 2.0+.

controller/raft/raft.go
plorenz: It gets reset in the leader observer, whenever we detect that the cluster has a leader

controller/raft/raft.go
plorenz: fixed

controller/raft/mesh/mesh.go
plorenz: fixed

controller/raft/mesh/mesh.go
plorenz: fixed

controller/raft/raft.go
plorenz: fixed

controller/raft/mesh/mesh.go
andrewpmartinez: More like what if we pick the version, the connection dies due to an upgrade/downgrade into a restart, and we never detect the version change.

plorenz: It will resolve eventually. The cluster can't live with mixed version for long because it goes into read-only mode. So either everything will get downgrade or everything will get upgraded and in either case, state will settle back to a good state. This is just an effort to prevent a short-term bad state during an upgrade.

PR #3686, opened 2026-03-13
Keep controller mesh fully connected, as much as possible. Fixes #3684

### LINKED ISSUE #3684: Keep controller mesh fully connected, as much as possible
The mesh currently follows raft conventions, which means it tries to fully connect during elections, but only maintains connections to the leader when there is a leader.
We should try to maintain the full mesh and allow raft to manage it's virtual connections on top of our fully connected mesh. This will present a less confusing picture to users, who expect a fully connected mesh, and will be advantageous if we want to start using gossip to propagate router data.

plorenz:


controller/raft/mesh/peer_dialer.go
```diff
@@ -0,0 +1,306 @@
...
+							WithField("address", addr).
+							Error("error closing channel to removed peer")
+					}
+				}()
+			}
+		}
+	}
+
+	newCount := 0
+	spreadInterval := 50 * time.Millisecond
```
andrewpmartinez, line 148: Hard-coded? I don't see mutations, constant?

controller/raft/mesh/peer_dialer.go
```diff
@@ -0,0 +1,306 @@
...
+	if !state.dialActive.CompareAndSwap(false, true) {
+		return // already being dialed
+	}
+
+	state.status = peerStatusDialing
+
+	go self.doDial(state)
+}
+
+func (self *PeerDialer) doDial(state *peerDialState) {
```
andrewpmartinez, line 235: Do we need synchronization on `state`? This code doesn't own it, and it can be updated.

controller/raft/mesh/mesh.go
andrewpmartinez, line 555: Is this ever used?

controller/raft/mesh/peer_dial_state_heap.go
```diff
@@ -0,0 +1,45 @@
...
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package mesh
+
+type peerDialStateHeap []*peerDialState
```
andrewpmartinez, line 19: Can this accumulate duplicates? A heap w/ no index.

controller/raft/mesh/mesh.go
```diff
@@ -432,17 +446,67 @@ func (self *impl) Dial(address raft.ServerAddress, timeout time.Duration) (net.C
...
+
+	select {
+	case peer := <-waiter:
+		return peer, nil
+	case <-time.After(timeout):
+		return nil, errors.Errorf("timeout waiting for peer at %v", address)
+	case <-self.closeNotify:
+		return nil, errors.New("mesh closed while waiting for peer")
+	}
+}
```
andrewpmartinez, line 498: Do we need to clean up peerWaiters? At least on timeout?

andrewpmartinez (changes requested):
Unsure of sync issues, heap bloat, and peer wait buildup.

controller/raft/mesh/peer_dialer.go
plorenz: added a comment

plorenz: also pulled out into a constant

controller/raft/mesh/peer_dialer.go
plorenz: non-atomic state is now read before being dispatched to a different goroutine

controller/raft/mesh/mesh.go
plorenz: fixed, it's now used correctly

controller/raft/mesh/peer_dial_state_heap.go
plorenz: index added

controller/raft/mesh/mesh.go
plorenz: fixed

plorenz:
Ready for re-review

andrewpmartinez (approved):

PR #3700, opened 2026-03-19
Coalesce OIDC JWT revocations to reduce controller write pressure. Fixes #3681

### LINKED ISSUE #3681: coalesce OIDC JWT revocations to reduce controller write pressure
Add time-bucketed batching of OIDC JWT token revocation writes because bulk revocation events may create excessive controller write pressure.

- Batch revocation writes into a single write per configurable time window instead of one write per token
- Add `revocationBucketInterval` config field (time-setting format, default `1m`) to set the bucket window size
- Add `revocationMinTokenLifetime` config field (time-setting format, default unset) to skip revocation entries for tokens expiring within that duration
- Validate that `revocationMinTokenLifetime` is less than 50% of the configured refresh token lifetime; reject invalid configuration on startup

plorenz:
- adds DeleteRevocationsBatchCommand so expired-revocation cleanup goes
  through raft as a single log entry per batch
- adds CreateRevocationsBatchCommand for batched revocation creation
  through raft
- moves refresh-token revocations from synchronous inline creation to a
  background batcher that flushes on a configurable interval, removing
  the database and raft as a bottleneck on token refreshes
- skips revocation creation for tokens expiring within a configurable
  threshold (revocationMinTokenLifetime), since they become invalid on
  their own
- validates that revocationMinTokenLifetime is less than 50% of the
  configured refresh token lifetime
- makes the revocation enforcer frequency configurable and restricts it
  to run only on the raft leader
- adds tests for multi-batch delete, batched create with router RDM
  propagation, and skip-threshold behavior
- adds new configuration tunables under edge.oidc: revocationBucketInterval,
  revocationMinTokenLifetime, revocationBucketMaxSize, revocationMaxQueued,
  revocationEnforcerFrequency


controller/model/revocation_manager.go
```diff
@@ -138,3 +141,146 @@ func (self *RevocationManager) Unmarshall(bytes []byte) (*Revocation, error) {
...
+	}
+	return self.Dispatch(cmd)
+}
+
+// ApplyCreateBatch persists a batch of revocations in a single DB transaction.
+func (self *RevocationManager) ApplyCreateBatch(cmd *CreateRevocationsBatchCommand, ctx boltz.MutateContext) error {
+	return self.GetDb().Update(ctx, func(ctx boltz.MutateContext) error {
+		for _, rev := range cmd.Revocations {
+			if _, err := self.createEntityInTx(ctx, rev); err != nil {
+				return err
```
dovholuknf, line 227: ApplyDeleteBatch doesn't work the same way as this - is that difference intentional? here the whole create is aborted where delete seems to do its best and carry on?

controller/oidc_auth/config.go
```diff
@@ -39,6 +39,11 @@ type Config struct {
 	RedirectURIs         []string
 	PostLogoutURIs       []string
 
+	RevocationMinTokenLifetime time.Duration
+	RevocationBucketInterval   time.Duration
+	RevocationBucketMaxSize    int
+	RevocationMaxQueued        int
+
```
dovholuknf, line 46: should this contain RevocationEnforcerFrequency? or was that omitted on purpose? i see controller/config/config_edge.go has five new things but see only four here?

controller/webapis/oidc-api.go
```diff
@@ -118,6 +118,10 @@ func NewOidcApiHandler(serverConfig *xweb.ServerConfig, ae *env.AppEnv, options
 	oidcConfig.AccessTokenDuration = ae.GetConfig().Edge.Oidc.AccessTokenDuration
 	oidcConfig.RefreshTokenDuration = ae.GetConfig().Edge.Oidc.RefreshTokenDuration
 	oidcConfig.IdTokenDuration = ae.GetConfig().Edge.Oidc.IdTokenDuration
+	oidcConfig.RevocationMinTokenLifetime = ae.GetConfig().Edge.Oidc.RevocationMinTokenLifetime
+	oidcConfig.RevocationBucketInterval = ae.GetConfig().Edge.Oidc.RevocationBucketInterval
+	oidcConfig.RevocationBucketMaxSize = ae.GetConfig().Edge.Oidc.RevocationBucketMaxSize
+	oidcConfig.RevocationMaxQueued = ae.GetConfig().Edge.Oidc.RevocationMaxQueued
```
dovholuknf, line 125: same comment on RevocationEnforcerFrequency here?

controller/oidc_auth/storage.go
```diff
@@ -351,9 +355,16 @@ func (s *HybridStorage) start() {
 				}
 			}
 		}()
+		s.batcher.Start()
```
dovholuknf, line 358: i mean i know "it will never happen".... and seems to be 'previous code' (not related to this pr) but if Clean takes more than 10 seconds, should "something" happen? a panic, stop the clean, a backoff, something?  is it possible to flood auth requests in a DoS type of attack here? Maybe this is already accounted for elsewhere? just something that kinda catches myattention so making a comment to look at it....

controller/server/controller.go
```diff
@@ -47,11 +47,11 @@ type Controller struct {
...
-	policyMinFreq          = 1 * time.Second
-	policyMaxFreq          = 1 * time.Hour
-	policyAppWanFreq       = 1 * time.Second
-	policySessionFreq      = 5 * time.Second
-	policyRevocationFreq   = 1 * time.Minute
+	policyMinFreq               = 1 * time.Second
+	policyMaxFreq               = 1 * time.Hour
+	policyAppWanFreq            = 1 * time.Second
+	policySessionFreq           = 5 * time.Second
+	policyRevocationFreqDefault = 1 * time.Minute
```
dovholuknf, line 54: anything that should be configuration related, it'd be swell if we could keep defaults in a single place like `ziti\constants\constants.go`. These and the config changes seem like changes we should add to the config generation template (not for this pr i suppose just informational/aspirational)

controller/oidc_auth/revocation_batcher.go
```diff
@@ -0,0 +1,142 @@
...
+			log.Info("revocation batcher shutting down, flushing remaining entries")
+			b.drainLoop()
+			return
+		}
+	}
+}
+
+// drainLoop flushes one batch at a time until fewer than maxBatchSize entries
+// remain. Each batch is dispatched through raft without holding the lock.
+func (b *RevocationBatcher) drainLoop() {
```
dovholuknf, line 93: so technically if there are more pending than the batch size this won't 'drain fully' right? i don't see a test that makes sure this works right. also if that's the case doesn't the b.drainLoop from the b.closeNotify branch above possibly leave some behind?

dovholuknf: what i mean -- does the b.closeNotify b.drainLoop need to process them all FULLY (not just one iteration)

controller/model/revocation_manager.go
plorenz: It's a reasonable question. Thinking about it, there's not much difference in the outcome, as changes will get rolled back either way. In the delete case you'll get a list of all problematic ids, for create, just the first. The chances of either failing are very low. If they did fail, it's more likely to be a whole-db issue, like being out of space, in which case neither behavior is particularly advantageous.

controller/oidc_auth/config.go
plorenz: It's just that it's not needed here. Where RevocationEnforcerFrequency is used, it has access to the original config, so we don't need to pass it through here.

controller/webapis/oidc-api.go
plorenz: same as above

controller/oidc_auth/storage.go
plorenz: According to the docs: 
```
The ticker will adjust the time interval or drop ticks to make up for slow receivers.
```
It's also cleaning in-memory structures, so it shouldn't be a concern.

controller/server/controller.go
plorenz: config in general needs an overhaul. Hopefully soon

controller/oidc_auth/revocation_batcher.go
plorenz: It should process one batch if there at least one thing to process. If there's more than batch size left, it should process more until there's less than batch size left. The remainder should be left for the next interval, otherwise we may be doing more, smaller batches.

PR #3707, opened 2026-03-21
Increase link payload/ack queue sizes and make them configurable. Fixes #3706

### LINKED ISSUE #3706: Increase link payload/ack queue sizes and make them configurable
The link payload and ack sender queues on grouped (multi-underlay) links are hardcoded to small buffer sizes (payload: 64, ack: 4). Under sustained throughput the payload queue can fill, creating backpressure on the forwarding path. The ack queue at only 4 slots is particularly tight and can delay acknowledgments, which in turn stalls the sender side.

These values should be raised and made tunable so operators can adjust them for their workload without a code change.

- Default payload sender queue size is increased from 64 to at least 128
- Default ack sender queue size is increased from 4 to at least 64
- Both values are configurable in the router config under the `link` section (`payloadSenderQueueSize`, `ackSenderQueueSize`)
- Invalid config values (non-integer, less than 1) produce a clear startup error
- Existing router configs with no explicit queue size settings continue to work with the new defaults

plorenz:


router/xlink_transport/channel.go
```diff
@@ -17,11 +17,11 @@ const (
 	ChannelTypeDefault string = "link.default"
 )
 
-func NewBaseLinkChannel(underlay channel.Underlay) *BaseLinkChannel {
+func NewBaseLinkChannel(underlay channel.Underlay, payloadSenderQueueSize, ackSenderQueueSize int) *BaseLinkChannel {
```
andrewpmartinez, line 20: Any reason not to send in an options struct?

plorenz: The dial and listen sides have different config patterns, so we don't have a common type to pass in :-/

PR #3725, opened 2026-03-26
Add CreateCircuitV3 for RDM-authorized circuit creation. Fixes #3721

### LINKED ISSUE #3721: Add CreateCircuitV3 to controller
CreateCircuitV3 allows the router to provide a circuit id. This will support circuit based (as opposed to connection id base) routing in the SDK, by letting the router select the circuit and send it to the sdk ahead of time. This lets the sdk set up routing before the circuit is established. We also no longer require the session token, as the router can validate the connect request and ensure that the identity has the necessary access.

plorenz:
- adds CreateCircuitV3 message type and handler for routers that have
  already authorized dials locally via RDM, bypassing service session
  tokens in favor of identity ID, service ID, and pre-assigned circuit ID
- renames CreateCircuitRequest/Response to CreateCircuitV2Request/V2Response
  for clarity now that V3 exists
- adds CircuitManager.Reserve to atomically claim circuit IDs before routing,
  preventing collisions on pre-assigned IDs
- extends CreateCircuitParams with GetCircuitId so V3 can supply a
  pre-assigned circuit ID (falls back to UUID generation when empty)
- fixes IsDialableByIdentity which was incorrectly calling IsBindableByIdentity
- extracts V2 handler into its own file create_circuit_v2.go
- adds CreateCircuitV3RequestType/ResponseType (20222/20223) to edge_ctrl protobuf
- registers V3 handler in controller server


common/ctrl_msg/messages.go
```diff
@@ -93,35 +96,35 @@ func NewRouteResultFailedMessage(sessionId string, attempt int, rerr string) *ch
 	return msg
 }
 
-type CreateCircuitRequest struct {
+type CreateCircuitV2Request struct {
```
dovholuknf, line 99: I take it there's no "v1" to worry about and renaming this is 'fine'... i don't really know so ... :/

controller/handler_ctrl/circuit_request.go
```diff
@@ -147,3 +147,7 @@ func (self *circuitParams) GetLogContext() logcontext.Context {
 func (self *circuitParams) GetDeadline() time.Time {
 	return self.deadline
 }
+
+func (self *circuitParams) GetCircuitId() string {
+	return ""
```
dovholuknf, line 152: is this really always empty on purpose?

tests/create_circuit_v3_test.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+type circuitV3Error struct {
+	msg  string
+	code uint32
+}
+
+func (e *circuitV3Error) Error() string {
+	return e.msg
+}
```
dovholuknf, line 197: > fixes IsDialableByIdentity

should there be some sort of test for that? did I miss it or is it somewhere else maybe?

common/ctrl_msg/messages.go
plorenz: There's a v1 version in protobuf. I renamed this because it's actually the v2 version, so the lack of a v2 was confusing me :)

controller/handler_ctrl/circuit_request.go
plorenz: It's a new parameter, so in most cases/implementations it will be empty. Only the new v3 path allows passing the circuit id in.

tests/create_circuit_v3_test.go
plorenz: It's tested by the new integ test. It will fail if this fix is reverted.

PR #3745, opened 2026-04-02
config type target

### LINKED ISSUE #3743: Implement controller managed configuration for routers

### LINKED ISSUE #3744: Add a target field to config type
Currently we only support tying configs to services. In the future we want to allow creating configurations for routers and maybe tunnelers. If we add a target field to the config type, we can ensure that users don't try to assign configs to entities of the wrong type. This will also help UIs list only the appropriate config types for a given entity.

The target should be immutable once created to avoid confusion.

plorenz:
- **Add target field to config types. Fixes #3744**


common/pb/edge_cmd_pb/edge_cmd.proto
```diff
@@ -173,6 +173,7 @@ message ConfigType {
   string name = 2;
   bytes schema = 3;
   map<string, TagValue> tags = 4;
+  optional string target = 5;
```
andrewpmartinez, line 176: Why optional? I assume this is for backward compatibility, but `optional` makes `string` be `*string` and doesn't really do much else besides letting us detect whether it was explicitly set or not. I don't see a reason to detect whether a setting is set or not.

I assume that the default is a target value of "service," and we could create an enum that does this:

```
enum ConfigTargetType {
  CONFIG_TYPE_SERVICE = 0;
  CONFIG_TYPE_ROUTER = 1;
}
```

Then the config type can be the enum type and default cleanly to 0 and not have to deal w/ pointer refs.

doc/design/anycast.md
```diff
@@ -0,0 +1,226 @@
...
+Currently `verifyRouter()` requires a synchronous RPC to the controller during link setup. We
+can eliminate this round-trip by distributing router certificate fingerprints to all routers
+ahead of time. Two options:
+
+1. **PeerStateChange messages.** Add fingerprints to the `PeerStateChange` data alongside
+   the existing router ID, version, and listeners. Simple, but means fingerprints are sent
+   with every peer state update.
+2. **Router data model.** Store fingerprints on the router data model and let them propagate
+   through the normal data model sync. Less chatty, since fingerprints only change on router
+   re-enrollment or cert rotation, not on every connect/disconnect.
```
andrewpmartinez, line 137: I'd vote for RDM, there are other related changes that could be done, but don't solve this exact issue. They have trade-offs.

doc/design/anycast.md
```diff
@@ -0,0 +1,226 @@
+# Anycast Support for Router Links
```
andrewpmartinez, line 1: Did you mean to include this on this PR? Seems out of place.

ziti/cmd/edge/list.go
```diff
@@ -1529,13 +1529,14 @@ func runListConfigTypes(o *api.Options) error {
...
 	t.SetStyle(table.StyleRounded)
-	t.AppendHeader(table.Row{"ID", "Name", "Schema"})
+	t.AppendHeader(table.Row{"ID", "Name", "Target", "Schema"})
 
 	for _, entity := range children {
 		wrapper := api.Wrap(entity)
 		t.AppendRow(table.Row{
 			wrapper.String("id"),
 			wrapper.String("name"),
+			wrapper.String("target"),
```
andrewpmartinez, line 1539: Mismatched columns? 4 columns, only outputting 3 values?

controller/internal/routes/config_type_api_model.go
```diff
@@ -34,7 +34,8 @@ func MapCreateConfigTypeToModel(configType *rest_model.ConfigTypeCreate) *model.
 		BaseEntity: models.BaseEntity{
 			Tags: TagsOrDefault(configType.Tags),
 		},
-		Name: stringz.OrEmpty(configType.Name),
+		Name:   stringz.OrEmpty(configType.Name),
+		Target: configType.Target,
```
andrewpmartinez, line 38: Isn't this optional in the API def, and it isn't defaulted here? Who uses NIL set target types? The API should be clear, but if backward compatibility is the goal, it seems reasonable to default to "service".

controller/model/edge_service_model.go
```diff
@@ -61,24 +61,31 @@ func (entity *EdgeService) toBoltEntityForCreate(tx *bbolt.Tx, env Env) (*db.Edg
...
+		configType, _ := configTypeStore.LoadById(tx, config.TypeId)
+		configTypeName := "<not found>"
+		if configType != nil {
+			configTypeName = configType.Name
+			if configType.Target == nil || *configType.Target != db.ConfigTypeTargetService {
+				msg := fmt.Sprintf("config %v has config type %v which does not target services",
+					config.Name, configTypeName)
+				return errorz.NewFieldError(msg, "configs", entity.Configs)
 			}
+		}
```
andrewpmartinez, line 86: So if it is nil, we don't validate it? So we allow nil values and don't validate their target type?

andrewpmartinez (changes requested):
I think overall, the PRs dealing with target type for configs need more effort to default to 'service' for backward compatibility. Allowing nil values and such is just a recipe for disaster.

common/pb/edge_cmd_pb/edge_cmd.proto
plorenz: Fixed. Made target immutable. On create, default target to 'service' if not provided.

doc/design/anycast.md
plorenz: removed doc for now

doc/design/anycast.md
plorenz: removed doc for now

ziti/cmd/edge/list.go
plorenz: fixed

controller/internal/routes/config_type_api_model.go
plorenz: fix. Empty targets are no longer allowed

controller/model/edge_service_model.go
plorenz: Fixed, nil targets not allowed

changelogs/CHANGELOG.2.0.md
```diff
@@ -0,0 +1,1539 @@
+# Release 2.0.0
```
andrewpmartinez, line 1: I'm lost as to why this "2.0.0" would be added if we already released it.

controller/db/config_type_store.go
```diff
@@ -18,29 +18,48 @@ package db
...
 const (
 	FieldConfigTypeSchema = "schema"
+	FieldConfigTypeTarget = "target"
+
+	// ConfigTypeTargetService indicates a config type used for service configuration.
+	ConfigTypeTargetService = "service"
+	// ConfigTypeTargetRouter indicates a config type used for router configuration.
+	ConfigTypeTargetRouter = "router"
+	// ConfigTypeTargetOther indicates a config type used for purposes other than service or router configuration.
+	ConfigTypeTargetOther = "other"
```
andrewpmartinez, line 40: Am I losing my mind, or is this not documented anywhere? CLI, notes, etc. Also, what is it for?

controller/model/edge_service_model.go
```diff
@@ -61,24 +61,31 @@ func (entity *EdgeService) toBoltEntityForCreate(tx *bbolt.Tx, env Env) (*db.Edg
...
 		}
-		conflictConfig, found := typeMap[config.TypeId]
-		if found {
-			configTypeName := "<not found>"
-			if configType, _ := env.GetStores().ConfigType.LoadById(tx, config.TypeId); configType != nil {
-				configTypeName = configType.Name
+
+		configType, _ := configTypeStore.LoadById(tx, config.TypeId)
+		configTypeName := "<not found>"
+		if configType != nil {
```
andrewpmartinez, line 79: At this point, we have a config and are loading the config type.

If the `config type == nil`, we don't validate it anymore, but also don't reject it or error.

andrewpmartinez (approved):
Comments about the change log, probably just need to rebase/merge main.

The edge_service_model.go comment about configType==nil is ....odd? I'll let you decide what the right thing is, but reading the code, it makes no sense to load something w/o a related config type. I also assume that key ref's would stop this from happening.

changelogs/CHANGELOG.2.0.md
plorenz: I need to rebase to get the delta into the updated CHANGELOG.md

controller/model/edge_service_model.go
plorenz: Configs don't require a config type. Historically we've allowed services to be linked to configs with no type, so I didn't change this behavior. I would guess that no one uses this feature, and we could change it. On the other hand, I'm not sure this looseness has any real downside, so I'm inclined to leave it as is.

controller/db/config_type_store.go
plorenz: Fixed. Added to changelog and cli. Configs could be used outside the context of services (and now routers) as a general store for management information. Since config types now require a target, this allows admins/devs to create config types that are explicitly not service oriented.

PR #3759, opened 2026-04-03
bug fixes

### LINKED ISSUE #3753: SPIFFE trust domain prefix check has swapped HasPrefix arguments
## Summary

In `controller/config/config.go`, the `strings.HasPrefix` calls for SPIFFE trust domain validation have swapped arguments:

```go
if !strings.HasPrefix("spiffe://", trustDomain) {
```

This checks whether the literal `"spiffe://"` starts with `trustDomain`, which is backwards. The result is that the `spiffe://` prefix is always prepended, producing malformed URIs like `spiffe://spiffe://example.com`.

Appears twice: once for `trustDomain` and once for `additionalTrustDomains`.

## Impact

Only affects non-HA configurations that set `trustDomain` with an explicit `spiffe://` prefix in the config. The URI gets double-prefixed, which could cause SPIFFE ID validation failures.

## Fix

Swap to `strings.HasPrefix(trustDomain, "spiffe://")`.

### LINKED ISSUE #3754: dialFailed drops applyFailed parameter, preventing duplicate link retry jitter
## Summary

In `router/link/link_registry.go`, the `dialFailed` method accepts an `applyFailed bool` parameter but never passes it to the event struct:

```go
func (self *linkRegistryImpl) dialFailed(state *linkState, applyFailed bool) {
    self.queueEvent(&updateLinkStatusToDialFailed{
        linkState: state,
        // applyFailed not set
    })
}
```

The `applyFailed` field on `updateLinkStatusToDialFailed` is always `false`, so the caller that passes `true` (to add random jitter when a duplicate link redial fails) has no effect.

## Impact

Duplicate link situations never get the intended retry jitter, which could lead to synchronized redial storms between routers.

## Fix

Set the `applyFailed` field on the event struct.

### LINKED ISSUE #3755: commandHandler config read from wrong scope
## Summary

In `controller/config/config.go`, the `commandHandler` config key is read from the top-level `cfgmap` instead of `submap` (the cluster config section):

```go
if value, found := cfgmap["commandHandler"]; found {
```

All surrounding reads in the cluster config block use `submap`. A `commandHandler` key nested under `cluster` in the config file is never found.

## Impact

The `commandHandler.maxQueueSize` setting under `cluster` is silently ignored. Only affects HA configurations that attempt to tune the command handler queue size.

## Fix

Change `cfgmap` to `submap`.

### LINKED ISSUE #3756: TLS handshake rate limiter timeout check reads from wrong config scope
## Summary

In `controller/config/config.go`, `loadTlsHandshakeRateLimiterConfig` checks for a user-specified `timeout` in the wrong map:

```go
_, timeoutSpecified = cfgmap["timeout"]
```

Here `cfgmap` is the `tls` config section, but `timeout` is a key inside the `rateLimiter` subsection (`submap`).

## Impact

A user-specified `tls.rateLimiter.timeout` is loaded by `rateLimitConfig.Load(submap)` but the `timeoutSpecified` flag is never set, so the loaded value is always overwritten with the default (handshakeTimeout + 2s). The timeout configuration is effectively ignored.

## Fix

Change `cfgmap["timeout"]` to `submap["timeout"]`.

### LINKED ISSUE #3757: Mesh peer signing cert from header is overwritten by TLS underlay cert
## Summary

In `controller/raft/mesh/mesh.go`, the `AcceptUnderlay` bind handler parses a signing cert from the hello header, but then unconditionally overwrites `peer.SigningCerts` with the TLS underlay cert:

```go
// Line ~871: parses header cert
peer.SigningCerts = append(peer.SigningCerts, cert)

// Line ~891: overwrites it
peer.SigningCerts = []*x509.Certificate{underlay.Certificates()[0]}
```

The dialer side (`GetOrConnectPeer`) also ignores the header cert entirely.

## Impact

The explicitly advertised signing cert from the hello header is never used. Both sides always fall back to the TLS underlay cert. This matters when the signing cert differs from the TLS transport cert.

## Fix

Prefer the signing cert from the header when available, falling back to the TLS underlay cert. Apply consistently on both dialer and acceptor sides.

plorenz:
- **Fix swapped HasPrefix args in SPIFFE trust domain check. Fixes #3753**
- **Fix dialFailed dropping applyFailed parameter. Fixes #3754**
- **Fix commandHandler config read from wrong scope. Fixes #3755**
- **Fix TLS rate limiter timeout check reading from wrong scope. Fixes #3756**
- **Fix mesh peer signing cert from header being overwritten. Fixes #3757**
- **Fix minor config parsing bugs**
- **Fix data race on heartbeat lastResponse and peer logger label**

### DIFF
--- a/controller/handler_ctrl/bind.go
+++ b/controller/handler_ctrl/bind.go
@@ -17,6 +17,7 @@
 package handler_ctrl
 
 import (
+	"sync/atomic"
 	"time"
 
 	"github.com/openziti/ziti/v2/common/pb/ctrl_pb"
@@ -98,8 +99,8 @@ func (self *bindHandler) BindChannel(binding channel.Binding) error {
 		ch:                       binding.GetChannel(),
 		latencySemaphore:         concurrenz.NewSemaphore(2),
 		closeUnresponsiveTimeout: self.heartbeatOptions.CloseUnresponsiveTimeout,
-		lastResponse:             time.Now().Add(self.heartbeatOptions.CloseUnresponsiveTimeout * 2).UnixMilli(), // wait at least 2x timeout before closing
 	}
+	cb.lastResponse.Store(time.Now().Add(self.heartbeatOptions.CloseUnresponsiveTimeout * 2).UnixMilli()) // wait at least 2x timeout before closing
 	channel.ConfigureHeartbeat(binding, self.heartbeatOptions.SendInterval, self.heartbeatOptions.CheckInterval, cb)
 
 	xctrlDone := make(chan struct{})
@@ -122,7 +123,7 @@ func (self *bindHandler) BindChannel(binding channel.Binding) error {
 type heartbeatCallback struct {
 	latencyMetric            metrics.Histogram
 	queueTimeMetric          metrics.Histogram
-	lastResponse             int64
+	lastResponse             atomic.Int64
 	ch                       channel.Channel
 	latencySemaphore         concurrenz.Semaphore
 	closeUnresponsiveTimeout time.Duration
@@ -136,12 +137,12 @@ func (self *heartbeatCallback) HeartbeatRespTx(int64) {}
 
 func (self *heartbeatCallback) HeartbeatRespRx(ts int64) {
 	now := time.Now()
-	self.lastResponse = now.UnixMilli()
+	self.lastResponse.Store(now.UnixMilli())
 	self.latencyMetric.Update(now.UnixNano() - ts)
 }
 
 func (self *heartbeatCallback) timeSinceLastResponse(nowUnixMillis int64) time.Duration {
-	return time.Duration(nowUnixMillis-self.lastResponse) * time.Millisecond
+	return time.Duration(nowUnixMillis-self.lastResponse.Load()) * time.Millisecond
 }
 
 func (self *heartbeatCallback) CheckHeartBeat() {
--- a/controller/handler_peer_ctrl/bind.go
+++ b/controller/handler_peer_ctrl/bind.go
@@ -17,6 +17,7 @@
 package handler_peer_ctrl
 
 import (
+	"sync/atomic"
 	"time"
 
 	"github.com/michaelquigley/pfxlog"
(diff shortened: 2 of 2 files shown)


controller/raft/mesh/mesh.go
```diff
@@ -544,7 +544,14 @@ func (self *impl) GetOrConnectPeer(address string, timeout time.Duration) (*Peer
 		}
 
 		peer.Version = versionInfo
-		peer.SigningCerts = []*x509.Certificate{underlay.Certificates()[0]}
```
andrewpmartinez, line 547: Who provides the header? If we're unsure of its source, it is a spoofing opportunity.

andrewpmartinez: Oh, I see, this is the signing cert.

controller/raft/mesh/mesh.go
```diff
@@ -544,7 +544,14 @@ func (self *impl) GetOrConnectPeer(address string, timeout time.Duration) (*Peer
...
+		if certHeader, found := headerWithFallback(peer.Channel.Underlay().Headers(), SigningCertHeader, LegacySigningCertHeader); found {
+			if cert, err := x509.ParseCertificate(certHeader); err == nil {
+				peer.SigningCerts = []*x509.Certificate{cert}
+			}
+		}
+		if len(peer.SigningCerts) == 0 {
+			peer.SigningCerts = []*x509.Certificate{underlay.Certificates()[0]}
+		}
```
andrewpmartinez: This should be reverted. Dialing side should use the remote controller's leaf server cert.

controller/raft/mesh/mesh.go
```diff
@@ -888,7 +888,14 @@ func (self *impl) AcceptUnderlay(underlay channel.Underlay) error {
...
 		peer.Version = versionInfo
-		peer.SigningCerts = []*x509.Certificate{underlay.Certificates()[0]}
+		if certHeader, found := headerWithFallback(ch.Underlay().Headers(), SigningCertHeader, LegacySigningCertHeader); found {
+			if cert, err := x509.ParseCertificate(certHeader); err == nil {
+				peer.SigningCerts = []*x509.Certificate{cert}
+			}
+		}
+		if len(peer.SigningCerts) == 0 {
+			peer.SigningCerts = []*x509.Certificate{underlay.Certificates()[0]}
+		}
```
andrewpmartinez, line 891: The changes to accept are fine and actually fix a bug, as long as the dialing side sets the header.

controller/raft/mesh/mesh.go
plorenz: fixed

PR #3761, opened 2026-04-03
[Backport-1.6] Fix revocation build-up in db and rdm. Fixes #3760

### LINKED ISSUE #3760: [Backport-1.6] revocation build-up in db and rdm
Backport of #3673 to 1.6

**Experienced**

Revocation records accumulate in the controller's bbolt database and in each router's in-memory data model. Networks with active OIDC usage and frequent token refreshes will see growth in both controller storage and router memory over time.

**Expected**

Revocation records should be purged once they are no longer required.

**Reproduction**

1. Configure a controller with OIDC enabled (default in 2.0+).
2. Enroll one or more identities using the SDK.
3. Allow SDKs to run for an extended period, performing normal token refresh cycles.
4. Query the `revocations` bucket in the bbolt database. The record count will grow monotonically with no upper bound.

plorenz:
- fixes RevokeToken double-write: adds return after JWTID-keyed revocation save, preventing fallthrough write with raw JWT string as unreachable key 
- fixes TerminateSession key mismatch: stores revocation by identityId alone, matching the Subject-based lookup in ValidateAccessToken 
- fixes RevocationDelete sync action: passes DataState_Delete instead of DataState_Create so routers evict the entry from their data model 
- adds RevocationManager.DeleteExpired: batch-deletes expired revocations in batches of 500 until none remain 
- adds RevocationEnforcer: periodic policy runner (every 1 minute) that calls DeleteExpired and records metrics

### DIFF
--- a/controller/oidc_auth/storage.go
+++ b/controller/oidc_auth/storage.go
@@ -755,7 +755,7 @@ func (s *HybridStorage) TokenRequestByRefreshToken(_ context.Context, refreshTok
 // TerminateSession implements the op.Storage interface
 func (s *HybridStorage) TerminateSession(_ context.Context, identityId string, clientID string) error {
 	now := time.Now()
-	return s.saveRevocation(NewRevocation(identityId+","+clientID, now.Add(s.config.MaxTokenDuration())))
+	return s.saveRevocation(NewRevocation(identityId, now.Add(s.config.MaxTokenDuration())))
 }
 
 // GetRefreshTokenInfo implements the op.Storage interface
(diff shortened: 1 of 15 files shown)


controller/oidc_auth/storage.go
```diff
@@ -755,7 +755,7 @@ func (s *HybridStorage) TokenRequestByRefreshToken(_ context.Context, refreshTok
 // TerminateSession implements the op.Storage interface
 func (s *HybridStorage) TerminateSession(_ context.Context, identityId string, clientID string) error {
 	now := time.Now()
-	return s.saveRevocation(NewRevocation(identityId+","+clientID, now.Add(s.config.MaxTokenDuration())))
```
andrewpmartinez, line 758: I was specifically looking for this change. This fixes a bug w/ revocations.

PR #3775, opened 2026-04-06
fixes #3770 close channel on invalid token update requests

### LINKED ISSUE #3770: Update Token Requests Should Close Channel Connections If Invalid
In routers, when an OIDC token update message is handled, if the token is invalid (cryptographically, API Session ID, etc.), the channel should be closed. Channels are closed when the current registered token expires, but token update messages that appear forged/invalid/incongruent should always close the channel.

*COAs*
- if a token's API Session ID doesn't match, close channel, return error, "invalid token, API session id does not match"
- if a token fails cryptographically, close channel, return error, "invalid token, invalid signature"
- if a token can't be parsed, close channel, return error "invalid token, could not be parsed"

andrewpmartinez:
- closes channel when token update contains an unparseable bearer token
  - closes channel when token update fails JWT signature validation
  - closes channel when token update API session ID does not match
  - fixes missing return after API session ID mismatch check, which allowed mismatched tokens to fall through to processing

### DIFF
--- a/router/xgress_edge/listener.go
+++ b/router/xgress_edge/listener.go
@@ -1354,11 +1354,11 @@ func (self *edgeClientConn) processTokenUpdate(req *channel.Message, ch channel.
 		reply := sdkedge.NewUpdateTokenFailedMsg(retErr)
 
 		retErr.ApplyToMsg(reply)
-		reply.ReplyTo(req)
 
-		if err := ch.Send(reply); err != nil {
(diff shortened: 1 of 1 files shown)


router/xgress_edge/listener.go
```diff
@@ -1374,13 +1374,14 @@ func (self *edgeClientConn) processTokenUpdate(req *channel.Message, ch channel.
 		if err := ch.Send(reply); err != nil {
```
plorenz: Can switch these to SendAndWaitForWire

andrewpmartinez: Done

Issue #3543, opened 2026-02-04
New Circuit Failure code for sockets not available


mikegorman-nf:
It appears as though if an Edge Router runs out of sockets to provide a NAT instance, the error code returned is ERR_ROUTER_DIAL_TIMED_OUT, which I assume is because it times out waiting for a socket to become available.  If there is a way to capture the specific event of socket blocking, it would be a great help in troubleshooting particular event types moer quickly.

Issue #3364, opened 2025-11-12
Make no such host error specific


mikegorman-nf:
Currently, when a service is routed via FQDN to a service, and the binding host cannot resolve the FQDN, the error code is ROUTER_ERR_DIAL_TIMED_OUT.  If that were a specific messages such as ROUTER_ERR_NO_SUCH_HOST, it would be more useful in reviewing and troubleshooting the network behaviors.

Issue #2888, opened 2025-03-13
New specific Error code for port not allowed


mikegorman-nf:
A failure of the port validation against the allowed range results in a ROUTER_ERR_GENERIC failure cause.  It would be very useful to have a separate error code for this, as it indicates a misconfiguration, or an attempt to reach ports out of policy, possibly signaling malicious intent.

Log entry:
`Mar 10 17:47:45 ip-10-19-232-153 ziti-controller[264984]: {"_channels":["selectPath"],"apiSessionId":"cm81ryt2wosivgoqkt9nb5b9m","attemptNumber":1,"circuitId":"BwPDd2inOV","error":"error creating route for [s/BwPDd2inOV] on [r/XXB0Jd8KwN] (error creating route for [c/BwPDd2inOV]: port 7070 is not in allowed port ranges)","file":"github.com/openziti/ziti/controller/network/network.go:597","func":"github.com/openziti/ziti/controller/network.(*Network).CreateCircuit","level":"warning","msg":"route attempt for circuit failed","serviceId":"vngGz6W0LDMh6BFYTUGO7","serviceName":"service-name","sessionId":"cm81s3pwqospjgoqklwj2dqmb","time":"2025-03-10T17:47:45.346Z"}`

Issue #2859, opened 2025-02-28
Cerate specific error code for DNS failed resolution


mikegorman-nf:
An error in DNS resolution on the hosting device should have it's own specific error code to aid in troubleshooting.  Example message below.

Feb 27 00:48:12 ip-10-19-177-14 ziti-router[945622]: {"_context":"ch{edge}-\u003eu{classic}-\u003ei{Kxjv}","chSeq":3127866,"connId":5678,"edgeSeq":0,"error":"exceeded maximum [2] retries creating circuit [c/NSv8suZnx]: error creating route for [s/NSv8suZnx] on [r/KZQGYWXKxv] (error creating route for [c/NSv8suZnx]: dial tcp: lookup readacted.servicename..com on 172.17.0.1:53: no such host)","file":"github.com/openziti/ziti/router/xgress_edge/listener.go:199","func":"github.com/openziti/ziti/router/xgress_edge.(*edgeClientConn).processConnect","level":"warning","msg":"failed to dial fabric","time":"2025-02-27T00:48:12.944Z","type":"EdgeConnectType"}

PR #3792, opened 2026-04-09
Add specific circuit failure error codes. Fixes #3717, #3543, #3364, #2888, #2859, #1580

### LINKED ISSUE #3717: Generic error message for specific error
A specific error, "Rejected by application" is being reported as a ROUTER_ERROR_GENERIC failure cause.  It would be much more useful to network operations if that had its own error code reported in the circuit events.

`received failed route status from [r/vLXmAIR9BQ] for attempt [#1] of [s/xE8IxNgKM] (error creating route for [c/xE8IxNgKM]: failed to establish connection with terminator address 4R3AZdA9NxNTXjhogUNYGD. error: (rejected by application))`

plorenz:
- adds ErrorType constants for rejected-by-application, DNS resolution failed,
  port not allowed, invalid link destination, and resources not available
- adds corresponding CircuitFailureCause strings reported in circuit events
- extracts classifyDialError() in route handler to map dial errors to specific
  error codes using typed errors, syscall constants, and string matching
- detects DNS errors via *net.DNSError and string fallback for ER/T hosted
  services where errors are serialized through the SDK message protocol
- detects resource exhaustion via EMFILE, ENFILE, ENOBUFS syscall errors
- introduces InvalidLinkDestinationError typed error in forwarder package
- adds unit tests covering all 16 classification cases
- adds integration tests for rejected-by-application (SDK host),
  DNS resolution failed, connection refused, and port not allowed
  (ER/T host mode) with circuit event verification
- adds CreateEnrollAndStartTunnelerEdgeRouterWithCfgTweaks to test context

### DIFF
--- a/controller/network/routesender.go
+++ b/controller/network/routesender.go
@@ -189,6 +189,7 @@ func (self *routeSender) handleRouteSend(attempt uint32, path *model.Path, strat
 			self.serviceCounters.ServiceDialOtherError(terminator.GetServiceId())
 			failureCause = CircuitFailureRouterErrPortNotAllowed
 		case ctrl_msg.ErrorTypeInvalidLinkDestination:
+			self.serviceCounters.ServiceDialOtherError(terminator.GetServiceId())
 			failureCause = CircuitFailureRouterErrInvalidLinkDest
 		case ctrl_msg.ErrorTypeResourcesNotAvailable:
 			self.serviceCounters.ServiceDialOtherError(terminator.GetServiceId())
--- a/controller/network/routesender.go
+++ b/controller/network/routesender.go
@@ -179,6 +179,20 @@ func (self *routeSender) handleRouteSend(attempt uint32, path *model.Path, strat
 		case ctrl_msg.ErrorTypeConnectionRefused:
 			self.serviceCounters.ServiceTerminatorConnectionRefused(terminator.GetServiceId(), terminator.GetId())
 			failureCause = CircuitFailureRouterErrDialConnRefused
+		case ctrl_msg.ErrorTypeRejectedByApplication:
+			self.serviceCounters.ServiceDialOtherError(terminator.GetServiceId())
+			failureCause = CircuitFailureRouterErrRejectedByApp
+		case ctrl_msg.ErrorTypeDnsResolutionFailed:
+			self.serviceCounters.ServiceDialOtherError(terminator.GetServiceId())
+			failureCause = CircuitFailureRouterErrDnsResolutionFailed
+		case ctrl_msg.ErrorTypePortNotAllowed:
+			self.serviceCounters.ServiceDialOtherError(terminator.GetServiceId())
+			failureCause = CircuitFailureRouterErrPortNotAllowed
+		case ctrl_msg.ErrorTypeInvalidLinkDestination:
+			failureCause = CircuitFailureRouterErrInvalidLinkDest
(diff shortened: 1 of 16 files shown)


router/handler_ctrl/route.go
andrewpmartinez, line 241: Is this function wrong? This is any net error, and it looks like an ordered switch statement is used to help this function, when I think we can just augment this function:

```go
func isNetworkTimeout(err error) bool {
    var netErr net.Error
    return errors.As(err, &netErr) && netErr.Timeout()
}
```

It looks like this entire implementation is fine, but the ordering is fragile. I think this would help reduce fragility.

controller/network/routesender.go
```diff
@@ -179,6 +179,20 @@ func (self *routeSender) handleRouteSend(attempt uint32, path *model.Path, strat
...
+			self.serviceCounters.ServiceDialOtherError(terminator.GetServiceId())
+			failureCause = CircuitFailureRouterErrRejectedByApp
+		case ctrl_msg.ErrorTypeDnsResolutionFailed:
+			self.serviceCounters.ServiceDialOtherError(terminator.GetServiceId())
+			failureCause = CircuitFailureRouterErrDnsResolutionFailed
+		case ctrl_msg.ErrorTypePortNotAllowed:
+			self.serviceCounters.ServiceDialOtherError(terminator.GetServiceId())
+			failureCause = CircuitFailureRouterErrPortNotAllowed
+		case ctrl_msg.ErrorTypeInvalidLinkDestination:
+			failureCause = CircuitFailureRouterErrInvalidLinkDest
```
andrewpmartinez, line 193: Is there a reason this one case doesn't call `self.serviceCounters.ServiceDialOtherError()`?

router/handler_ctrl/classify_dial_error_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	t.Run("resource exhaustion ENOBUFS", func(t *testing.T) {
+		err := fmt.Errorf("dial failed: %w", syscall.ENOBUFS)
+		require.Equal(t, byte(ctrl_msg.ErrorTypeResourcesNotAvailable), classifyDialError(err))
+	})
+
+	t.Run("timeout", func(t *testing.T) {
+		err := fmt.Errorf("dial failed: %w", syscall.ETIMEDOUT)
+		require.Equal(t, byte(ctrl_msg.ErrorTypeDialTimedOut), classifyDialError(err))
+	})
```
andrewpmartinez, line 67: Missing the *net.Op variant which is what go actually throws in most paths.

`*net.OpError` wrapping `syscall.ETIMEDOUT`.

Example:  `&net.OpError{Err: &os.SyscallError{Err: syscall.ETIMEDOUT}}`

andrewpmartinez (approved):
Small improvements suggested.

controller/network/routesender.go
plorenz: Yes. It's not a dial error, it's a route failure caused by a link going away, so it doesn't count as a service dial error. However, I think semantically we're using ServiceDial to be the circuit dial, not the dial on the router, so it probably should update that. fixing.

router/handler_ctrl/classify_dial_error_test.go
plorenz: added a test for this

router/handler_ctrl/route.go
plorenz: fixed, good catch

PR #3803, opened 2026-04-14
Update sdk-golang and channel to more recent versions

plorenz:


go.mod
```diff
@@ -5,6 +5,8 @@ go 1.26.1
 // pinned
 replace github.com/michaelquigley/pfxlog => github.com/michaelquigley/pfxlog v0.6.10
 
+replace github.com/openziti/foundation/v2 => github.com/openziti/foundation/v2 v2.0.86
+
```
andrewpmartinez, line 9: Intentional?

andrewpmartinez (approved):

go.mod
plorenz: yes, had to pin this because newer versions of foundation introduced a breaking change with the errors type

PR #3805, opened 2026-04-14
fixes #3734 enforce client certificate proof-of-possession for OIDC sessions

### LINKED ISSUE #3734: Enforce client certificate proof-of-possession on controller REST API for OIDC sessions
Enforce `z_cfs` (certificate fingerprint) and SPIFFE ID validation on controller REST API requests authenticated with OIDC JWT access tokens, so that cert-authenticated sessions require the matching client certificate to be presented on every API call — not just at the router.

**COAs**

- when `z_cfs` is non-empty, require the TLS peer certificate to match by fingerprint or SPIFFE ID
- enforce on management, fabric, client API
- return `401 Unauthorized` on failure

andrewpmartinez:
- adds verifyCertProofOfPossession() in resolveOidcSession() to require TLS client cert matching a z_cfs fingerprint or SPIFFE ID when the OIDC token was issued with cert bindings
- adds z_cae (CertAllowExpired) claim from auth policy, propagated through token issuance and refresh
- adds TrustCache.VerifyClientCert() with tiered pool matching (first-party roots, trust anchors, third-party) and TTL cache
- adds WrapIdentityWithCertValidation() on the router to verify client certs at the TLS level against RDM PublicKeys
- adds IsFirstPartyCert() on the router to gate SPIFFE ID matching by checking whether the cert chains to the controller root CA
- adds VerifySpiffeId() in common/spiffehlp with SpiffeMatchApiSession, SpiffeMatchIdentity, and SpiffeMatchNone return types
- adds SPIFFE IDs to OTT and token enrollment certs (/identity/<id>)
- enforces cert expiry by match type: API session certs must be valid, fingerprint-matched certs respect z_cae, legacy sessions skip checks
- shallow-copies leaf certs before overriding time fields in all cert verification paths to avoid races on shared x509.Certificate pointers
- fixes controllerRootCache setting inited=true before the ctrl channel is available, which permanently cached the failure

### DIFF
--- a/router/state/cert_validating_identity.go
+++ b/router/state/cert_validating_identity.go
@@ -1,0 +1,122 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package state
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"time"
+
+	"github.com/michaelquigley/pfxlog"
+	"github.com/openziti/identity"
+	"github.com/openziti/ziti/v2/common/pb/edge_ctrl_pb"
+)
+
+// certValidatingIdentity wraps an identity.Identity to add client certificate chain verification
+// at the TLS level via a VerifyConnection callback. It dynamically builds a CA pool from the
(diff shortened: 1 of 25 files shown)


router/state/cert_validating_identity.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// Add the router identity's CA bundle as intermediates. The RDM PublicKeys typically
+	// contain only root CAs, but client certs may be signed by an intermediate CA that
+	// is part of the router's trust bundle.
+	if idCa := self.Identity.CA(); idCa != nil {
+		intermediatePool = idCa
+	}
+
+	// Also add any additional certs from the TLS peer chain as intermediates.
+	for _, cert := range state.PeerCertificates[1:] {
```
plorenz, line 101: Adding certs to a shared pool (the identity CA)

plorenz (approved):
Overall looks good. Just had the concern with the shared pool and the test failure. I looked at that quickly, looked odd, I didn't see a specific test that failed.

router/state/cert_validating_identity.go
andrewpmartinez: Fixed

PR #3821, opened 2026-04-20
Rework connect event handling to ensure serialized-per-router handling

plorenz:


CHANGELOG.md
```diff
@@ -117,29 +117,35 @@ when running HA. Legacy API and service session are now deprecated and will be r
...
 * [Multiple DNS upstreams](#multiple-dns-upstreams) - the tunneler can now fan out recursive queries to several upstream resolvers in parallel
 
 ## Connect Events Pool
 
-The controller now uses a shared, bounded goroutine pool to process identity
-connect/disconnect events from routers. Previously each router connection spawned
-a dedicated goroutine that was never cleaned up on disconnect, leaking a goroutine
-per reconnect cycle. Under churn (e.g., chaos testing with hundreds of routers) this
-could accumulate tens of thousands of leaked goroutines and destabilize the controller.
+The controller now uses per-router, single-worker goroutine pools to process identity
```
andrewpmartinez, line 169: Is it really per-router? Or is it per-channel to a router? I suppose it is kind of the same thing. I don't know if a router might have more than 1 connection to a controller.

andrewpmartinez:
Is this testable? At least unit for FIFO ordering?

andrewpmartinez (approved):

CHANGELOG.md
plorenz: Routers can only have one channel (they can have multiple connections, but those will be underlays for the same channel). Controllers don't allow two channels for the same router id.

plorenz:
> Is this testable? At least unit for FIFO ordering?

I'm not sure what we'd test. There's a single channel of events and at most one goroutine processing them. If we need more testing it would be generic testing of the goroutines library, to ensure it's living up to its API.

PR #3826, opened 2026-04-21
Support `ziti agent inspect` so ziti processes can be inspected via IPC. Fixes #3824. Fixes #2049

### LINKED ISSUE #2049: The ziti agent command should have a controller connection status
The ziti agent CLI command has subcommands to get the current link dump, which can imply the connection to the controller, but it would be great to have an explicit command to show if a process is connected to a controller.  In HA, maybe if run against the controller process it would show the connections to other controllers?

### LINKED ISSUE #3824: Allow calling inspect using the IPC agent on the controller, router and go tunnel
This lets admins inspect local processes in cases where the controller may be not be reachable.

plorenz:
- adds `ziti agent inspect <value>...` CLI that sends an InspectRequest directly to a ziti process over its agent IPC channel and pretty-prints JSON values in the response
- accepts app id 0 on controller and router agent channels so a single inspect command works against any process type
- adds Controller.agentOpInspect, backed by a new InspectionsManager.InspectLocal that runs inspect processing on the local controller only, without fanning out to routers or peer controllers
- extracts the router inspect handler into a new router/inspect package and stores a single shared instance on Router, reused by both the control channel and the agent IPC channel
- adds RouterEnv.GetInspectHandler and RouterEnv.GetXgressListeners so the control channel bind pulls the shared handler from env
- removes the now-redundant InspectRouterEnv interface from handler_ctrl/bind.go
- adds tunnel HandleAgentAsyncOp with support for stackdump and sdk inspect keys

### DIFF
--- a/router/agent.go
+++ b/router/agent.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/channel/v4"
+	"github.com/openziti/ziti/v2/common/agentid"
 	"github.com/openziti/ziti/v2/common/handler_common"
 	"github.com/openziti/ziti/v2/common/pb/ctrl_pb"
 	"github.com/openziti/ziti/v2/common/pb/mgmt_pb"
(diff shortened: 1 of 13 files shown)


ziti/tunnel/root.go
```diff
@@ -133,7 +133,8 @@ func rootPostRun(cmd *cobra.Command, _ []string) {
 			AppAlias:        cliAgentAlias,
 			AppType:         "tunnel",
 			CustomOps: map[byte]func(c net.Conn) error{
-				AgentDump: handleAgentDump,
+				AgentDump:          handleAgentDump,
+				agent.CustomOpAsync: HandleAgentAsyncOp,
```
andrewpmartinez: I noticed a single space is missing here. Looks like a missed `go fmt`.

router/agent.go
```diff
@@ -73,9 +74,9 @@ func (self *Router) HandleAgentAsyncOp(conn net.Conn) error {
 	}
 	appId := appIdBuf[0]
 
-	if appId != AgentAppId {
+	if appId != 0 && appId != AgentAppId {
```
andrewpmartinez: Magic number 0.

andrewpmartinez:
Changelog update needed?

andrewpmartinez (approved):
Just small things, nothing that stops this.

ziti/tunnel/root.go
plorenz: fixed and hopefully fixed the tooling gap that let that through

router/agent.go
plorenz: added a shared constant

plorenz:
> Changelog update needed?

Added

PR #3836, opened 2026-05-01
OIDC authentication fablab test

plorenz:


zitirest/clients.go
andrewpmartinez, line 1: This module zitirest and this type should probably be removed or at least made private/moved if possible. I see it is mostly used by a small subset of integration tests and the chaos tests you authored.
The module name makes it look like this is a generic facility to be used but it really isn't and has a narrow scope.
I don't know what the right move is. Using the SDK's lower-level auth modules is better than improving this in tandem. Those do have some testing capabilities built into them, but depending on how this is used, it might not be a drop-in.
I am tired of seeing new "rest clients" dropped in all over the code base and slowly falling out of line w/ the SDK implementation. I know this one has been around for a bit, but I'm le-tired.

I do see this is for the fabric API, but it should be consildated w/ the other clients.

controller/handler_mgmt/validate_identity_connection_statuses.go
```diff
@@ -214,8 +214,22 @@ func (handler *validateIdentityConnectionStatusesHandler) validRouterEdgeConnect
...
 	lock.Lock()
+
 	for identityId, connMap := range m {
 		if routerConn := connMap[router.Id]; routerConn != nil {
+			// The controller marks tunneler-enabled routers' own identities as
+			// connected (via the broker), but the router's edge connection tracker
+			// doesn't include the internal tunneler binding. Skip this check when
+			// the identity is the router itself and it's a tunneler-enabled router.
+			if identityId == router.Id {
+				if er, _ := handler.appEnv.Managers.EdgeRouter.Read(router.Id); er != nil && er.IsTunnelerEnabled {
```
andrewpmartinez: error swallowed instead of saying the ER doesn't exist?

ziti/constants/constants.go
```diff
@@ -19,16 +19,15 @@ package constants
...
 const (
-	OpenZitiOrg             = "openziti"
-	ZITI                    = "ziti"
-	ZROK                    = "zrok"
-	CaddyOrg                = "caddyserver"
-	Caddy                   = "caddy"
-	ZITI_CONTROLLER         = "ziti-controller"
-	ZITI_ROUTER             = "ziti-router"
-	ZITI_TUNNEL             = "ziti-tunnel"
-	ZITI_EDGE_TUNNEL        = "ziti-edge-tunnel"
```
andrewpmartinez, line 30: This is considered an API breaking change and warrants a changelog.md for 2.0.

andrewpmartinez (commented):
The test LGTM. The REST client OIDC auth stuff makes me sad.

zitirest/clients.go
plorenz: Fixed, used sdk-golang where possible and split the rest between tests/ and zititest/ so it's not public

controller/handler_mgmt/validate_identity_connection_statuses.go
plorenz: fixed

ziti/constants/constants.go
plorenz: Are we considering all of ziti public API? I thought we were limiting our concern there to sdk-golang and the REST APIs?

dovholuknf (approved):

PR #3840, opened 2026-05-04
fixes #3809 support CSR submission during OIDC authentication

### LINKED ISSUE #3809: Support CSR submission during OIDC authentication for session-bound certificates
Allow non-cert authenticated identities to submit a CSR with their credentials during OIDC authentication because it enables certificate-based channel communication for endpoints that authenticate via external IdPs without requiring pre-provisioned client certificates.

- accept an optional CSR during OIDC authentication when both primary and secondary auth are provided
- sign the CSR and generate a certificate with a one-year lifetime
- bind the certificate to the API session using a SPIFFE ID derived from the identity and API session
- return the signed certificate in the authentication response
- authentication events should indicate whether the identity authenticated with a pre-existing certificate or with a non-cert method that had a certificate generated and bound to the API session at authentication time

andrewpmartinez:
- accepts an optional CSR during OIDC login (all auth methods) and signs it into a session-bound certificate with a SPIFFE ID derived from the identity and API session
- returns the signed certificate PEM as a top-level "session_cert" field in the token endpoint JSON response (CodeExchange, RefreshToken, TokenExchange)
- adds cert-binding verification on RefreshToken and TokenExchange: if z_cfs is present the peer cert fingerprint must match (strict), otherwise falls back to SPIFFE ID verification
- supports cert rotation via csr_pem form parameter on refresh and token exchange; replaces the session cert fingerprint while preserving the authenticating cert fingerprint
- adds AuthCertFingerprints (z_acfs) claim to track permanent auth cert fingerprints separately from rotatable session cert fingerprints
- only the leaf certificate fingerprint is added to z_cfs and z_acfs, intermediates are never included
- invalid CSR returns 400 Bad Request in OIDC error format
- propagates updated CustomClaims (including CertFingerprints) from access token to renewed refresh token so rotated fingerprints are enforced on subsequent refreshes
- adds CertGenerated field to ApiSessionEvent
- advertises OIDC_AUTH_WITH_CSR controller capability when OIDC is enabled
- adds unit tests for verifyCertBinding (fingerprint, SPIFFE ID, edge cases) and CsrPem field parsing
- adds integration tests for initial CSR auth (updb, cert, ext-jwt), cert-binding on refresh/exchange, CSR rotation with cert auth, SPIFFE fallback and z_cfs transition, and CSR property forging resistance

### DIFF
--- a/controller/event/api_session.go
+++ b/controller/event/api_session.go
@@ -79,11 +79,14 @@ type ApiSessionEvent struct {
 
 	// The IP address from which the identity to connected to require the api session.
 	IpAddress string `json:"ip_address"`
+
+	// CertGenerated is true when a certificate was generated from a CSR during authentication.
+	CertGenerated bool `json:"cert_generate"`
 }
 
 func (event *ApiSessionEvent) String() string {
-	return fmt.Sprintf("%v.%v id=%v timestamp=%v token=%v identityId=%v ipAddress=%v",
-		event.Namespace, event.EventType, event.Id, event.Timestamp, event.Token, event.IdentityId, event.IpAddress)
+	return fmt.Sprintf("%v.%v id=%v timestamp=%v token=%v identityId=%v ipAddress=%v certGenerated=%v",
+		event.Namespace, event.EventType, event.Id, event.Timestamp, event.Token, event.IdentityId, event.IpAddress, event.CertGenerated)
 }
 
 type ApiSessionEventHandler interface {
--- a/controller/event/api_session.go
+++ b/controller/event/api_session.go
@@ -81,7 +81,7 @@ type ApiSessionEvent struct {
 	IpAddress string `json:"ip_address"`
 
 	// CertGenerated is true when a certificate was generated from a CSR during authentication.
-	CertGenerated bool `json:"cert_generate"`
+	CertGenerated bool `json:"cert_generated"`
 }
 
 func (event *ApiSessionEvent) String() string {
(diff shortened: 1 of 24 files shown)


controller/internal/routes/version_router.go
```diff
@@ -150,6 +150,7 @@ func (ir *VersionRouter) buildVersions(ae *env.AppEnv) *rest_model.Version {
 
 	if oidcEnabled {
 		v.Capabilities = append(v.Capabilities, string(rest_model.CapabilitiesOIDCAUTH))
+		v.Capabilities = append(v.Capabilities, "OIDC_AUTH_WITH_CSR")
```
andrewpmartinez: Magic value is removed once https://github.com/openziti/edge-api/pull/190 is merged and tagged.

controller/event/api_session.go
```diff
@@ -79,11 +79,14 @@ type ApiSessionEvent struct {
 
 	// The IP address from which the identity to connected to require the api session.
 	IpAddress string `json:"ip_address"`
+
+	// CertGenerated is true when a certificate was generated from a CSR during authentication.
+	CertGenerated bool `json:"cert_generate"`
```
plorenz: Should be `json:"cert_generated"`

controller/oidc_auth/server.go
```diff
@@ -161,10 +212,86 @@ func (s *server) RefreshToken(ctx context.Context, r *op.ClientRequest[oidc.Refr
 		return nil, err
 	}
 
+	// Cert-binding verification: the TLS peer cert must match the token's cert fingerprints
+	// or SPIFFE ID. This prevents use of stolen refresh tokens.
```
plorenz: would it be an error for this type assertion to fail?

controller/event/api_session.go
andrewpmartinez: Fixed

controller/oidc_auth/server.go
andrewpmartinez: Sure should be. Fixed.

plorenz (approved):

PR #3858, opened 2026-05-07
fixes #3857 include intermediates in router first-party cert check

### LINKED ISSUE #3857: Router rejects API session certificate after HA failover for ext-jwt identities without enrollment
**Experienced**
External JWT signer identities without enrollment certificates cannot connect to services after an HA controller failover. After the tunnel restarts and re-authenticates via the external IdP, the edge router rejects the API session certificate with "invalid client certificate for api session." Identities enrolled via OTT JWT work correctly through the same failover scenario.

**Expected**
External JWT signer identities using API session certificates should be able to connect to services regardless of which controller in the HA cluster signed the certificate.

**Reproduction**
1. Deploy a 3-member HA controller cluster with each controller using its own intermediate CA for enrollment signing
2. Configure an external JWT signer (e.g. Keycloak) without JIT enrollment
3. Pre-create an identity with an external ID mapped to the JWT signer
4. Authenticate from an edge client — services connect successfully
5. Stop the controller that was the leader during step 4
6. Wait for leader election to complete on a remaining controller
7. Restart the tunnel on the client — re-authentication succeeds and services appear available
8. Attempt to connect to a service — the terminating router rejects with "invalid client certificate for api session"

andrewpmartinez:
- updates IsFirstPartyCert to accept the full TLS peer chain and populate VerifyOptions.Intermediates from peerCerts[1:]
- updates the state.Manager interface signature to match
- updates validateApiSession to pass the full peer chain instead of only the leaf

### DIFF
--- a/router/state/cert_origin.go
+++ b/router/state/cert_origin.go
@@ -42,10 +42,14 @@ type controllerRootCache struct {
 	inited   bool
 }
 
-// IsFirstPartyCert determines if a client certificate was issued by the internal (first-party) CA
-// by checking whether it chains to the same root CA as the controller connection. The controller
-// root CA is extracted from any ctrl channel and cached for subsequent calls.
-func (self *ManagerImpl) IsFirstPartyCert(clientCert *x509.Certificate) bool {
+// IsFirstPartyCert reports whether the leaf of peerCerts chains to a controller-trusted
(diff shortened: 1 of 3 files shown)


dovholuknf (approved):
no relevant test?

PR #3865, opened 2026-05-13
e2ee: make sure public key returned in DialSuccess is propagated

### LINKED ISSUE #3864: e2ee: allow hosting SDK to return e2ee public key in the dial response
A hosting SDK can return an e2ee public key in its DialSuccess response on an encryption-required service, and that key is used for the resulting session instead of the terminator's bind-time e2ee key.

**Conditions of acceptance**
- A hosting SDK may include an e2ee public key in DialSuccess.
- When present, the initiator uses that key for the session in place of the bind-time terminator key.
- When absent, behavior is unchanged: the bind-time terminator key is used.
- An encryption-required service with no key in DialSuccess and no bind-time key still fails the circuit as it does today.

ekoby:
fixes #3864 propagate hosting SDK public key from DialSuccess

- reads PublicKeyHeader from the dial reply in all three xgress_edge dial paths (Dial, dialSdkXgress, dialLegacy)
- returns it via PeerData so it flows through RouteResultSuccess and CircuitSuccess to the initiator SDK
- splits dialLegacy's previous combined `return nil, SendState(...)` so the success path can return PeerData while the SendState error path is preserved

### DIFF
--- a/router/xgress_edge/dialer.go
+++ b/router/xgress_edge/dialer.go
@@ -198,7 +198,11 @@ func (dialer *dialer) Dial(params xgress_router.DialParams) (xt.PeerData, error)
 	}
 	log.Debug("dial success")
 
-	return nil, nil
+	peerData := xt.PeerData{}
+	if pk := reply.Headers[edgeSdk.PublicKeyHeader]; len(pk) > 0 {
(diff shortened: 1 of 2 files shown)


ekoby:
superceded by #3874

PR #3869, opened 2026-05-14
Refcount shared intercept hostnames so iptables rules are installed for every service

### LINKED ISSUE #3867: Tunneler skips iptables rules for services sharing an intercept hostname
## Summary

When two or more services share an intercept hostname (the `addresses` field
of `intercept.v1`), only the first service registered at tunneler startup gets
an `NF-INTERCEPT` iptables rule installed. Subsequent services sharing the
hostname get a `tproxy` localhost listener but no kernel rule routing traffic
to it. Packets to the CGNAT IP miss `NF-INTERCEPT` and are silently dropped
(e.g. by a default-DROP `INPUT` policy). Clients see TCP connection timeouts;
there is no error logged at dial time and no error returned to the controller.

`tunnel/intercept/iputils.go:getDnsIp` returns early when the hostname already
has an allocated IP, skipping the `addrCB` callback that installs the iptables
rule. The only diagnostic signature is a single `info`-level
`hostname <host> already assigned (<ip>)` line at registration time.

## Reproducing

1. Configure two services with `intercept.v1` `addresses: ["shared.example"]`
   and different `portRanges` (e.g. service A on port 4723, service B on port
   8088).
2. Restart `ziti-router` so both services are registered fresh.
3. `iptables -t mangle -L NF-INTERCEPT -n -v` shows a rule for whichever
   service was processed first and no rule for the other.
4. `ss -tulnp` shows one more `127.0.0.1:<port>` listener than the chain has
   rules — the orphan listener belongs to the broken service.

The log signature for the broken service is a `tproxy listening on
tcp:127.0.0.1:<port>` line with no matching `Adding rule iptables ...
--on-port=<port>` line.

plorenz:
Fixes #3867

## Summary
- When two or more services share an intercept hostname, `getDnsIp` returned early on the second service and skipped the `addrCB` callback that installs the iptables rule. The result: a tproxy listener with no kernel rule routing traffic to it, so packets to the CGNAT IP were silently dropped.
- This change invokes `addrCB` and registers a cleanup action on the existing-hostname path too, and refcounts hostname allocations so the resolver entry / IP are only released when the last service using them is cleaned up.
- The refcount key is canonicalized (lowercased) to match the resolver's case-insensitive view, so two services configured with different-case spellings of the same hostname share one allocation.

### DIFF
--- a/tunnel/dns/refcount.go
+++ b/tunnel/dns/refcount.go
@@ -65,9 +65,9 @@ func (self *RefCountingResolver) RemoveHostname(s string) net.IP {
 		return nil
 	}
 
-	// count <= 1 covers both the last reference and hostnames never added
-	// through this layer (e.g. wildcard-domain hostnames, which are added
-	// directly on the wrapped resolver but cleaned up through this one)
+	// count <= 1 covers both the last reference and, defensively, hostnames
+	// never added through this layer (e.g. when the wrapped AddHostname
+	// failed, so the count was never incremented)
 	delete(self.names, key)
 	return self.wrapped.RemoveHostname(s)
 }
--- a/tunnel/dns/refcount.go
+++ b/tunnel/dns/refcount.go
@@ -2,19 +2,27 @@ package dns
 
 import (
 	"net"
-
-	cmap "github.com/orcaman/concurrent-map/v2"
+	"strings"
+	"sync"
 )
 
+// NewRefCountingResolver wraps resolver so the underlying hostname is only
+// removed once the last caller releases it. Successive AddHostname calls for
+// the same hostname bump a reference count; RemoveHostname decrements it and
+// only forwards to the wrapped resolver when the count reaches zero.
 func NewRefCountingResolver(resolver Resolver) Resolver {
 	return &RefCountingResolver{
-		names:   cmap.New[int](),
+		names:   map[string]int{},
 		wrapped: resolver,
 	}
 }
 
+// RefCountingResolver reference counts AddHostname/RemoveHostname calls per
+// hostname. The lock is held across the wrapped resolver calls so the count
+// and the wrapped resolver's state can't diverge under concurrent use.
 type RefCountingResolver struct {
-	names   cmap.ConcurrentMap[string, int]
+	lock    sync.Mutex
+	names   map[string]int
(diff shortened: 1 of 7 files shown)


tunnel/intercept/iputils.go
```diff
@@ -81,13 +95,25 @@ func cleanUpFunc(hostname string, resolver dns.Resolver) func() {
...
+	// resolver's (resolver internally lowercases). Two services configured
+	// with the same hostname in different cases must share one allocation.
+	canonical := strings.ToLower(host)
+
+	// If the hostname already has an allocated IP, reuse it. Even though no new
+	// IP is allocated, we still must invoke addrCB so the interceptor installs
+	// its per-service iptables rule; otherwise the tproxy listener exists with
+	// no kernel rule pointing at it and the service is silently unreachable.
+	if foundIP, found := resolver.LookupIP(canonical + "."); found {
+		addr := &net.IPNet{IP: foundIP, Mask: net.CIDRMask(len(foundIP)*8, len(foundIP)*8)}
```
andrewpmartinez: So does this addr construction differ from the one on 132?

If they have the same result, they should be formed the same.

plorenz: they both now use a hostMask method

tunnel/dns/refcount.go
```diff
@@ -35,31 +43,33 @@ func (self *RefCountingResolver) RemoveDomain(name string) {
...
+		self.names[key] = count - 1
+		return nil
 	}
-	return nil
+
+	// count <= 1 covers both the last reference and hostnames never added
+	// through this layer (e.g. wildcard-domain hostnames, which are added
+	// directly on the wrapped resolver but cleaned up through this one)
+	delete(self.names, key)
+	return self.wrapped.RemoveHostname(s)
```
andrewpmartinez: Speaking of wildcards, what happens if a wildcard hostname and a non-wild hostname both overlap?

e.g., one service at  `*.example.com` and another service at `test.example.com`

Looks like this might early cleanup on the collision. Not sure if it is a problem.

andrewpmartinez (approved):
Just the wildcard comment. Not sure if it matters.

PR #3870, opened 2026-05-14
[Backport-1.6] Refcount shared intercept hostnames so iptables rules are installed for every service

### LINKED ISSUE #3868: [Backport-1.6] Tunneler skips iptables rules for services sharing an intercept hostname
Backport of #3867 to `release-v1.6.x`.

plorenz:
Fixes #3868. Backport of #3869 to `release-v1.6.x`.

## Summary
- When two or more services share an intercept hostname, `getDnsIp` returned early on the second service and skipped the `addrCB` callback that installs the iptables rule. The result: a tproxy listener with no kernel rule routing traffic to it, so packets to the CGNAT IP were silently dropped.
- This change invokes `addrCB` and registers a cleanup action on the existing-hostname path too, and refcounts hostname allocations so the resolver entry / IP are only released when the last service using them is cleaned up.
- The refcount key is canonicalized (lowercased) to match the resolver's case-insensitive view, so two services configured with different-case spellings of the same hostname share one allocation.


andrewpmartinez:
This backport needs to brought inline w/ https://github.com/openziti/ziti/pull/3869

plorenz:
> This backport needs to brought inline w/ #3869

I'll update it once the source is merged so I know the code won't change

plorenz:
> This backport needs to brought inline w/ #3869

Done

andrewpmartinez (approved):

PR #3892, opened 2026-05-28
setup-cli: enhance ziti CLI version selection

ekoby:
- adds actual/resolved ziti cli version as output
- support regex version matching

examples:
`version: 'v2'` - will match latest release with `v2.` prefix including `v2.0.0-pre11`
`version: 'v2.0.[0-9]+$'` - will match latest `v2.0.*` release ignoring all `-preXX` versions

sample downstream usage is here https://github.com/openziti/ziti-sdk-c/actions/runs/26596699921

### DIFF
--- a/setup-cli/action.yml
+++ b/setup-cli/action.yml
@@ -7,8 +7,7 @@ inputs:
     description: |
       version of ziti CLI to download,
       use tag regex (like v1.6) or special values:
-      'stable' (default) means latest release, 
-      `latest` means most recent tagged release,
+      'stable' (default) means latest release (GitHub marked `latest`), 
       regex can be used to pick latest release matching the pattern,
       for example, `2.0.[0-9]+$` will match latest 2.0.x release ignoring any 2.0.x-pre* releases
 outputs:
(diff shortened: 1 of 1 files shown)


setup-cli/action.yml
```diff
@@ -6,24 +6,41 @@ inputs:
     default: stable
     description: |
       version of ziti CLI to download,
-      use tag (like v1.6.7) or special values:
+      use tag regex (like v1.6) or special values:
       'stable' (default) means latest release, 
-      `latest` means most recent tagged release
+      `latest` means most recent tagged release,
```
dovholuknf: if this is really 'most recent', it means if we tag 1.6.16 and it'll show up ahead of '2.0.10'. that would be confusing

setup-cli/action.yml
```diff
@@ -6,24 +6,41 @@ inputs:
...
       shell: bash
+      id: version
       run: |
         ver=${{ inputs.version }}
         if [[ ${{ inputs.version }} == "stable" ]]; then
+           echo "fetching stable version"
            ver=$(gh release view --repo openziti/ziti | awk '/^tag/{ print $2 }')
         elif [[ ${{ inputs.version }} == "latest" ]]; then
+          echo "fetching latest releases version"
            ver=$(gh release list --limit 1 --repo openziti/ziti --json "tagName" --jq ".[0].tagName")
```
dovholuknf: this should be converted to use semver imo

setup-cli/action.yml
ekoby: I'll just scrub `latest` altogether -- since it is very non-deterministic

dovholuknf (approved):

setup-cli/action.yml
ekoby: there is an output line below indicating resolved version

PR #3895, opened 2026-05-28
Import agent library into common/agent and consolidate channel-upgrade code

plorenz:
## Summary

Two things ride together here:

1. The design proposal for the upcoming slog logging refactor lands under
   `doc/design/`. The docs cover phases that are **not** implemented in this
   PR; only the agent groundwork is.
2. The agent groundwork itself: absorb `github.com/openziti/agent` into ziti
   as `common/agent`, then de-duplicate the four copies of the channel-upgrade
   plumbing into one shared helper.

This is a no-behavior-change preparation step. The follow-up PR will add the
typed v2 log-level channel commands on top.

Fixes #3893, fixes #3894.

## What's in this PR

**Design proposal** (`doc/design/`):

- `agent-capabilities.md` — capability discovery, AppInfoV2, v2 log-level commands.
- `logging-refactor.md` — the slog-based async logging refactor that builds on top.
- `logging-refactor-progress.md` — phased plan tying it all together.

Only Phase 1 (the df mirror, external) and the agent groundwork from Phase 2
(this PR) are touched here.

**Absorb `github.com/openziti/agent` into ziti** (#3893)

The library is openziti-owned but, per the GitHub dependency graph, used only
by ziti. The separate-repo boundary buys a cross-repo release dance and not
much else. The 5 `.go` files move into `common/agent`; the ~21 in-repo
importers re-point; the external dependency leaves both `go.mod` files. Wire
shape is identical (`Magic`, op constants, socket prefix). Mirrors the recent
storage-library absorption.

**Consolidate channel-upgrade plumbing** (#3894)

The conn-to-channel upgrade for the agent IPC was copy-pasted in four spots
(controller, router, tunnel, demo) plus the CLI-side dialer. They collapse
into `agent.HandleChannelConnection` (server) and `agent.NewChannel` /
`agent.ConnToChannel` / `agent.MakeChannelRequest` (client). Each app's
handler shrinks to a one-line delegation.

## Backwards compatibility

No behavior change. The wire is unchanged across every existing command.
`ziti agent ps`, `set-log-level`, `inspect`, the router debug commands, the
cluster commands all keep working bit-for-bit identically.

PR #3903, opened 2026-05-29
Add agent IPC capability discovery and channel-based log-level commands

### LINKED ISSUE #3902: Add agent IPC capability discovery and channel-based log-level commands
## Why

This is the follow-up to #3893 / #3894, which absorbed the agent library and
consolidated the duplicated channel-upgrade plumbing. With that prep done, we
can add the typed v2 log-level channel commands and the capability discovery
they need to coexist with old clients.

The framed `set-log-level` / `set-channel-log-level` / `clear-channel-log-level`
commands carry a `logrus.Level` byte, which welds the wire to logrus. New
channel-based commands carrying string level names let the broader slog
logging refactor (`doc/design/logging-refactor.md`) plug in without grafting
new shapes onto the legacy bytes.

## Scope

- **Agent-local `LogLevel` enum** in `common/agent`, with `String` /
  `ParseLogLevel`. Used at the callback boundary so `common/agent` stays
  agnostic of any specific logging library.
- **Capability registry** (`AgentLoggingSlogLevels`, names table, mask /
  string-list / bit-from-string helpers) and `RegisterAppCapabilities` for app
  strings. Both freeze on `Listen`; late registration panics.
- **`AppInfoV2`** framed command (op `0x16`) returning identity plus the two
  capability lists. Legacy `AppInfo` is untouched; old servers close with no
  write on the unknown op, and the client falls back cleanly.
- **Capability bitmask in the channel hello** for completeness, plus
  `CapabilitiesFromHeaders` to decode it.
- **Three v2 channel commands** `SetLogLevelV2`, `SetChannelLogLevelV2`,
  `ClearChannelLogLevelV2`, using content-type IDs in the agent-reserved
  `30000+` band, with parameters carried as channel string headers.
- **`LogLevelCallbacks` + `RegisterLogLevelHandlers`**, which both the v2 and
  legacy framed handlers route through. Embedders that don't register get
  today's logrus / pfxlog behavior unchanged.
- **`HandleChannelConnection` auto-binds the v2 handlers** on every agent
  channel when callbacks are registered.
- **ziti registration**: controller, router, and tunnel call
  `agent.RegisterLogLevelHandlers(agentlog.DefaultLogLevelCallbacks())` at
  startup, driving logrus + pfxlog (matching today's behavior).
- **CLI v2 selection**: `agent set-log-level` / `set-channel-log-level` /
  `clear-channel-log-level` prefer the v2 channel command when the target
  advertises `logging.slog-levels`, falling back to framed otherwise.
- Developer note (`common/agent/README.md`) and an in-process end-to-end test.

## Not in scope

The slog-based logging refactor itself (`AsyncHandler`, async sink, named
loggers, format compatibility). That sits on top of this work in a separate
follow-up.

## Done when

- Old clients (pre-PR) work against the new server unchanged.
- A new CLI against an old server falls back to the framed commands cleanly.
- A new CLI against a new server uses the v2 channel commands.
- The in-process end-to-end test covers `AppInfoV2`, the three v2 commands,
  and the framed-through-callback path.

plorenz:
## Summary

Follow-up to #3895, adding capability discovery and the typed v2 log-level
channel commands on top of the absorbed `common/agent`. Old clients keep
working unchanged; new clients prefer the typed v2 commands when the target
advertises them.

Fixes #3902.

**Stacked on #3895** — this PR's base is `agent-prep`, so the diff shows only
the 8 commits of new work. Merge #3895 first, then this rebases cleanly onto
`main`.

## What's in this PR

**Capability primitive** (`common/agent`)

A net-new `AppInfoV2` framed command (op `0x16`) returns identity plus two
capability lists: `agent_capabilities` (package-owned) and `app_capabilities`
(registered by the embedding application). Legacy `AppInfo` (`0xa`,
`map[string]string`) is untouched. A new client hitting an old server reads a
clean zero-byte EOF on the unknown op and falls back. Capabilities also ride
the channel hello as a `*big.Int` bitmask.

Advertisement is conditional: a capability is only emitted when its handler is
registered. `RegisterAppCapabilities` and `RegisterLogLevelHandlers` freeze
once `Listen` starts (panic on late registration).

**Three v2 log-level channel commands**

`SetLogLevelV2`, `SetChannelLogLevelV2`, `ClearChannelLogLevelV2`. Content-type
IDs in the agent-reserved `30000+` band (clear of ctrl_pb's 1000s and
mgmt_pb's 10000s, both of which ride the agent channel). Parameters are
carried as channel string headers.

`common/agent` stays logger-agnostic: it defines its own `LogLevel` enum,
parses the wire string once, and hands callbacks an `agent.LogLevel`. The
embedding application maps onto its loggers.

**Callbacks**

`LogLevelCallbacks` + `RegisterLogLevelHandlers`. Both the v2 channel handlers
and the legacy framed handlers route through these callbacks. Embedders that
don't register get today's logrus / pfxlog behavior unchanged.
`HandleChannelConnection` auto-binds the v2 handlers on every agent channel
when callbacks are registered.

**ziti wiring**

`common/agentlog.DefaultLogLevelCallbacks` is the one place mapping
`agent.LogLevel` onto logrus and pfxlog. Controller, router, and tunnel
register it at startup, before `Listen`. The CLI commands
(`agent set-log-level`, `set-channel-log-level`, `clear-channel-log-level`)
pick the v2 channel command when the target advertises `logging.slog-levels`,
falling back to the framed command otherwise. The `clear-channel-log-level`
two-arg positional-address form stays framed (it doesn't use the standard
targeting flags).

**Developer note and tests**

`common/agent/README.md` documents the two-tier capability model and how to
extend it (add an agent capability, register an app capability, add a v2
channel command, preserve wire compatibility). An in-process end-to-end test
exercises `AppInfoV2`, each of the three v2 channel commands, and the framed
path-through-the-callbacks.

## Backwards compatibility

- Legacy framed `set-log-level` / `set-channel-log-level` /
  `clear-channel-log-level` stay forever and now route through the same
  callbacks when registered, with a logrus/pfxlog fallback when not.
- Legacy `AppInfo` (`map[string]string`) is byte-identical to before, so old
  `ziti agent ps` / `list` keeps working against new servers.
- An old `ziti agent` CLI against a new server uses the legacy framed
  commands; the v2 capability simply isn't observed.

## Not in scope

The slog-based logging refactor itself (`AsyncHandler`, async sink, named
loggers, format compatibility). That sits on top of this work and will be a
separate follow-up.

### DIFF
--- a/common/agent/README.md
+++ b/common/agent/README.md
@@ -1,0 +1,112 @@
+# common/agent
+
+This package is the in-tree home of ziti's IPC agent: the `gops`-style framed
+protocol, the channel-transport upgrade, and the capability-advertisement
+primitive. It was absorbed from `github.com/openziti/agent`; see
+[doc/design/agent-capabilities.md](../../doc/design/agent-capabilities.md) for
+the full design and rationale. This note is the practical "how do I extend it"
+guide.
+
+## Capabilities: two tiers
+
+A process advertises what it supports through two capability lists, returned by
+the `AppInfoV2` command and (for agent capabilities) carried as a bitmask in the
+channel hello:
+
+- **Agent capabilities** are owned by this package. They describe features the
+  agent machinery itself provides, like the v2 log-level commands. They have a
+  stable bit position (for the hello bitmask) and a stable hierarchical-dotted
+  string name (for `AppInfoV2`). Today there is one: `logging.slog-levels`.
+- **App capabilities** are owned by the embedding application. The package does
+  not interpret them, it just passes the strings through. ziti can register
+  strings of its own (e.g. `ziti.something`) at startup; none are registered
+  yet.
+
+The two live in separate fields, so even an identical string in both namespaces
+cannot collide.
+
+A capability is only advertised when its handler is actually wired up. This
+keeps a binary from claiming a feature it did not install: a controller built
+without the logging wiring simply does not advertise `logging.slog-levels`, and
+capability-aware clients fall back to the legacy commands.
+
+## Adding an agent capability
+
+Agent capabilities are defined here, in `capabilities.go`:
+
+1. Add a bit constant. Bits are append-only and never renamed; if a capability
+   needs to change shape, add a new one.
+
+   ```go
+   const (
+       CapabilityLoggingSlogLevels int = 1
+       CapabilitySomethingNew      int = 2 // new
+   )
+   ```
+
+2. Add its canonical name to the `agentCapabilityNames` table. The string is the
+   identifier clients match on; the prefix should name the subsystem
+   (`logging.*`, `ipc.*`, ...).
+
+   ```go
+   var agentCapabilityNames = map[int]string{
+       CapabilityLoggingSlogLevels: "logging.slog-levels",
+       CapabilitySomethingNew:      "something.new",
+   }
+   ```
+
+3. Mark it active only when its handler is registered, by calling
+   `markAgentCapabilityActive(bit)` from the registration entry point (the way
+   `RegisterLogLevelHandlers` does for `CapabilityLoggingSlogLevels`). That is what
+   makes advertisement conditional. Registration must happen before
+   `Listen`; the capability set freezes once the listener starts.
+
+Clients check an agent capability by bit:
+
+```go
+if opts.HasAgentCapability(agent.CapabilityLoggingSlogLevels) { ... }
+```
+
+## Registering an app capability
+
+App capabilities need no changes here. The application keeps its own constants
+in its own package and registers the strings at startup, before `agent.Listen`:
+
+```go
+agent.RegisterAppCapabilities("ziti.something")
+```
+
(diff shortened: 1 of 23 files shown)


common/agent/README.md
```diff
@@ -0,0 +1,110 @@
...
+## Adding an agent capability
+
+Agent capabilities are defined here, in `capabilities.go`:
+
+1. Add a bit constant. Bits are append-only and never renamed; if a capability
+   needs to change shape, add a new one.
+
+   ```go
+   const (
+       AgentLoggingSlogLevels int = 1
```
andrewpmartinez: This is an example, but `AgentLoggingSlogLevels` appears to be an invalid existing constant.

andrewpmartinez (approved):

common/agent/README.md
plorenz: fixed doc drift cause by renames

PR #3905, opened 2026-05-29
Add common/logging async slog handler and chain composition

### LINKED ISSUE #3904: Add slog AsyncHandler in preparation for moving to slog for logging
## Why

ziti logs through pfxlog/logrus today, which holds a single process-wide
mutex across the format+write of every log call. Under chaos load that
serializes every goroutine that wants to log behind one mutex; we've seen
13 goroutines blocked on it at once during fablab runs.

The fix, designed in `doc/design/logging-refactor.md`, is to land `log/slog`
as the API contract behind a single async sink. This issue lands the
foundation of that sink — the `AsyncHandler`, its level helpers, and the
handler-chain wrappers for `WithAttrs`/`WithGroup` — as a new `common/logging`
package. Nothing migrates yet; the bridge from pfxlog/logrus and the
migration of call sites come in follow-up issues.

## Scope

- **Custom `slog.Level` constants** for `Trace=-8`, `Fatal=12`, `Panic=16`,
  plus `LevelName(slog.Level) string` and `ParseLevel(string) (slog.Level,
  error)` as the single source of truth for level names on the wire.
- **`AsyncOptions`** struct (`QueueSize`, `BlockThreshold`,
  `SummaryInterval`) with `Validate() error`, plus `AddFlags` /
  `OptionsFromFlags` for cobra integration (depends only on `spf13/pflag`).
- **`AsyncHandler`** — bounded queue, drain goroutine, per-level atomic drop
  counters, summary-line emission, `downstreamMu` shared with `SyncEmit`,
  non-blocking `Close` with bounded-drain contract.
- **`SyncEmit(ctx, r)`** for Fatal/Panic durability: bypasses the queue,
  writes through the downstream handler synchronously.
- **Handler-chain wrappers** `boundHandler` and `groupedHandler` implementing
  `WithAttrs`/`WithGroup`. `boundHandler.WithAttrs` returns a sibling
  boundHandler (same parent) rather than wrapping the receiver, so repeated
  `slog.Logger.With` does not grow chain depth. `WithGroup("")` and
  `WithAttrs(nil)` are no-ops on all three handler types.

## Not in scope

- The `logrus -> slog` bridge and `logging.Install`.
- The named-logger registry (`logging.For(name)`) and per-name overrides.
- Format compatibility coercion (pfxlog-shape JSON via `ReplaceAttr`).
- Migrating any call sites.

All in follow-up issues.

## Done when

- `common/logging` package exists with the level helpers, `AsyncOptions`,
  `AsyncHandler`, `SyncEmit`, and the `boundHandler`/`groupedHandler` chain.
- Tests cover: normal flow, full-queue drop with summary line, full-queue
  block at threshold, `Close` non-blocking and idempotent, concurrent
  `Handle` racing `Close` never panics, `SyncEmit` serializes through
  `downstreamMu`, `LevelName`/`ParseLevel` round-trip, `AsyncOptions.Validate`
  rejects bad values, the four worked examples from the design doc for the
  chain (`With.WithGroup.With`, `WithGroup.With`, nested groups, basic),
  no-nesting on repeated `With`, and sibling loggers do not leak attrs.
- No behavior change anywhere outside `common/logging`.

plorenz:
## Summary

Adds the `common/logging` package, the async slog sink the broader logging
refactor lands on. Nothing migrates yet: no logrus/pfxlog bridge, no named
logger registry, no call-site migration. This PR is just the foundation
(handler + level helpers + handler chain), self-contained inside
`common/logging` and not imported by anything else in the repo yet.

Fixes #3904.

**Stacked on #3895 + #3903.** This PR's base is `agent-capabilities`, so the
diff shows only the new `common/logging` work. Once those merge to `main`,
this rebases cleanly and the issue auto-close link will activate.

## What's in this PR

**Level helpers** (`levels.go`)

Custom `slog.Level` constants for `Trace=-8`, `Fatal=12`, `Panic=16`, plus
`LevelName(slog.Level) string` and `ParseLevel(string) (slog.Level, error)`
as the single source of truth for canonical lowercase wire names. `warn`
and `warning` both accepted on input.

**`AsyncOptions` + flag binding** (`options.go`)

`QueueSize`, `BlockThreshold`, `SummaryInterval` with `Validate`, defaults of
`4096`/`Warn`/`5s`, and `AddFlags`/`OptionsFromFlags` for cobra integration
(`spf13/pflag` only, no cobra dep in the package).

**`AsyncHandler`** (`handler.go`)

Bounded async `slog.Handler`:

- Hands records to a single drain goroutine and onto a downstream handler
  under a shared mutex.
- Records at or above `BlockThreshold` block on the queue (with a
  `closeNotify` escape so shutdown can't deadlock); records below it drop
  when full and bump a per-level atomic counter.
- The drain emits a drop-summary record on each `SummaryInterval` tick when
  any counter is non-zero. Downstream errors are counted (and logged once
  to `os.Stderr` to avoid slog recursion) and appear in the same summary
  line.
- `Close` signals shutdown and returns immediately; the drain final-flushes
  records that beat the close, emits a final summary if drops occurred,
  and closes `drainDone` for tests.
- `SyncEmit` bypasses the queue for Fatal/Panic durability — synchronous
  write through the downstream handler under the shared mutex.

**Handler-chain wrappers** (`chain.go`)

- `boundHandler` prepends bound attrs to every record before delegating.
  `WithAttrs` returns a new boundHandler whose parent is the receiver's
  parent (not the receiver), so repeated `slog.Logger.With` produces
  sibling boundHandlers at the same chain depth rather than stacking
  wrapper-on-wrapper.
- `groupedHandler` wraps record attrs in `slog.Group(name, ...)` before
  delegating. A subsequent `WithAttrs` creates a `boundHandler` whose
  parent is the `groupedHandler`, so the attrs land inside the group.
- `WithGroup("")` and `WithAttrs(nil)` return the receiver on all three
  handler types so no-op `.With()` / `.WithGroup("")` allocate nothing.

## Not in scope

- The `logrus -> slog` bridge and `logging.Install`.
- The named-logger registry (`logging.For(name)`) and per-name overrides.
- Format compatibility coercion (pfxlog-shape JSON via `ReplaceAttr`).
- Migrating any call sites.

All in follow-up PRs.

### DIFF
--- a/common/logging/handler.go
+++ b/common/logging/handler.go
@@ -1,0 +1,281 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package logging
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"os"
+	"slices"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// AsyncHandler is a bounded, async slog.Handler that hands records to a single
+// drain goroutine and onto a downstream handler. Records at or above the
+// configured BlockThreshold block when the queue is full; records below it
+// drop and are counted toward a periodic summary line.
+//
+// AsyncHandler is the root of the slog handler chain: WithAttrs returns a
+// boundHandler that prepends bound attrs to records, and WithGroup returns a
+// groupedHandler that wraps record attrs in slog.Group(name, ...). Both
+// wrappers delegate back to AsyncHandler.Handle, which is where the queueing
+// and drain dispatch live.
+type AsyncHandler struct {
+	opts         AsyncOptions
+	downstream   slog.Handler
+	queue        chan queuedRecord
+	closeNotify  chan struct{}
+	drainDone    chan struct{}
+	closed       atomic.Bool
+	downstreamMu sync.Mutex
+	dropCounts   [7]atomic.Int64
+	drainErrors  atomic.Int64
+	// windowStart is the start of the current summary window. It is only
+	// accessed by the drain goroutine after the handler is constructed.
+	windowStart time.Time
+}
+
+// queuedRecord carries the originating call's context alongside the record so
+// downstream handlers that consult ctx values (tracing, OTel) see them across
+// the async hop.
+type queuedRecord struct {
+	ctx    context.Context
+	record slog.Record
+}
+
+// NewAsyncHandler returns an AsyncHandler that drains records onto downstream.
+// It validates opts and starts the drain goroutine; on validation failure it
+// returns an error and no resources are leaked.
+func NewAsyncHandler(downstream slog.Handler, opts AsyncOptions) (*AsyncHandler, error) {
+	if downstream == nil {
+		return nil, errors.New("downstream handler must not be nil")
+	}
+	if err := opts.Validate(); err != nil {
+		return nil, err
+	}
+	h := &AsyncHandler{
+		opts:        opts,
+		downstream:  downstream,
+		queue:       make(chan queuedRecord, opts.QueueSize),
+		closeNotify: make(chan struct{}),
+		drainDone:   make(chan struct{}),
+		windowStart: time.Now(),
+	}
+	go h.drain()
+	return h, nil
+}
+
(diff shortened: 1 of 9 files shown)


common/logging/handler.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	go h.drain()
+	return h, nil
+}
+
+// Enabled returns true. AsyncHandler does no level gating itself; downstream
+// handlers (or callers) decide what to filter. This keeps Handle on the
+// hot-path of every slog call without an extra atomic load.
+func (h *AsyncHandler) Enabled(context.Context, slog.Level) bool {
+	return true
```
andrewpmartinez, line 94: Not a pro-slog guy yet, but I think this causes record allocation and queueing. It might be the cost of doing business while we transition, though.

andrewpmartinez (approved):

common/logging/handler.go
plorenz: The comment is misleading, I'm fixing that. The log level enabled check ends up in namedHandler.Enabled in the registry, so it does happen before slog constructs them. This stays stays gate-free because everything reaching it has already passed a level check (registry for direct slog, logrus for bridged).

PR #3907, opened 2026-05-30
Add named-logger registry, logrus bridge, and pfxlog-shape JSON

### LINKED ISSUE #3906: Add named-logger registry, logrus bridge, and pfxlog-shape JSON
## Why

Follow-up to #3904, which landed the AsyncHandler foundation and the slog
handler-chain wrappers. This issue adds the next two pieces that round out
the `common/logging` package as a self-contained slog sink:

- **Named-logger registry** with per-channel runtime level control.
  Operators can change the log level of a specific subsystem at runtime
  without affecting the rest of the process; the design doc explains why
  this matters for surgical debugging during incidents.
- **logrus -> slog bridge** with `Install`, `SyncEmit`, and pfxlog-shape
  JSON output. After this lands, any logrus call site (which is most of
  ziti today) can route through the slog sink with no source change at the
  call site; the agent IPC v2 log-level commands from #3902 already drive
  the callbacks that `Install` configures.

Nothing in the ziti binaries opts in yet; that comes in the follow-up
issue along with the df/dl pretty handler and the golden-file format tests.

## Scope

**Named-logger registry**

- `Registry` (exported, for tests) with global `*slog.LevelVar`, a per-name
  override map of `*slog.LevelVar`, a logger cache, and the root handler.
- `For(name) *slog.Logger` returns a cached logger whose handler chain
  binds `channel: name` as the first attr; panics on empty name.
- `SetGlobalLevel` / `SetNamedLevel` / `ClearNamedLevel` for runtime
  control; override clear deliberately tracks the live global rather than
  snapshotting it, so a later global change still reaches the
  previously-overridden name.
- `namedHandler` chain wrapper that gates `Enabled` on the
  registry-resolved level (override or live global) and forwards `Handle`
  to the registry's root.
- Package-level default Registry behind an `atomic.Pointer`; `Configure`
  swaps it and clears the logger cache. Anything called before `Configure`
  panics with a clear message.

**logrus -> slog bridge**

- `slogBridge` is a `logrus.Hook` that copies each `logrus.Entry` into a
  `slog.Record` and dispatches via `RootHandler.Handle`, except for Fatal
  and Panic which go through `SyncEmit` so they're durable before the
  process exits.
- `Install` / `InstallTo` redirect logrus output to `io.Discard`, replace
  its formatter with a noop, set its level to the mapped equivalent, and
  add the bridge as a hook. logrus's own pre-filter still applies, so
  below-threshold records never reach the bridge.
- `noopFormatter` is the logrus formatter logrus uses post-Install.
- Bidirectional level mapping between `logrus.Level` and `slog.Level` for
  the seven canonical names; non-canonical slog levels bucket into the
  canonical level whose value they most recently exceeded.
- `SyncEmit` writes through the root handler synchronously, type-asserting
  to `*AsyncHandler` when possible to bypass the queue under the same
  mutex the drain uses.

**Format compatibility**

- `ReplaceAttr` coerces JSON output to the pfxlog shape downstream parsers
  expect: lowercase level via `LevelName`, nested `source` attr suppressed
  at the top level.
- `sourceFlattener` chain wrapper decodes the record's PC into flat `file`
  and `func` attrs, matching pfxlog's existing JSON shape. Bridged records
  arrive with PC == 0 (logrus supplies file/func in Entry.Data), so the
  bridge path produces the same flat keys without a PC decode.
- `BuildHandler(out, opts)` constructs the production chain
  (AsyncHandler -> sourceFlattener -> JSONHandler over `out`) for callers
  that want the whole pipeline pre-wired.

## Not in scope

- The df/dl pretty handler integration and TTY-vs-redirected format
  default.
- Wiring `Install` into ziti's controller + router `main` and adding the
  AsyncOptions flags to the cobra commands.
- Golden-file format tests with the full fixture set.
- Startup invariant tests.
- The agent v2 callbacks driving `SetGlobalLevel` and `SetNamedLevel`
  (Phase 7).

All follow-up.

## Done when

- `common/logging` has `Registry`, `namedHandler`, `For` / `Configure` /
  `SetGlobalLevel` / `SetNamedLevel` / `ClearNamedLevel`, `slogBridge` /
  `Install` / `InstallTo` / `SyncEmit` / `RootHandler`, `ReplaceAttr` /
  `sourceFlattener` / `BuildHandler`.
- Tests cover the acceptance gates from the design doc.
- Nothing outside `common/logging` imports the new types yet.

plorenz:
## Summary

Stacked on #3905. Adds the next layer of `common/logging`: the named-logger
registry with runtime per-channel level control, the `logrus -> slog`
bridge plus `Install`, and the pfxlog-shape JSON output via `ReplaceAttr`
and `sourceFlattener`. Nothing in the ziti binaries opts in yet; this PR
is still self-contained inside `common/logging`.

Fixes #3906.

**Base is `slog-async-handler`** so the diff shows only the new work. Once
#3905 merges to `main`, this rebases cleanly.

## What's in this PR

**Named-logger registry** (`registry.go`)

- `Registry` (exported, for tests) holds the global `*slog.LevelVar`, a
  per-name override map of `*slog.LevelVar`, the root handler, and a
  Logger cache.
- `For(name)` returns a cached `*slog.Logger` whose handler chain binds
  `channel: name` as the first attr; panics on empty.
- `SetGlobalLevel` / `SetNamedLevel` / `ClearNamedLevel` for runtime
  control. Clear deliberately tracks the live global rather than
  snapshotting it, so later global changes still propagate to a
  previously-overridden name.
- `namedHandler` chain node gates `Enabled` on the registry-resolved
  level (override or live global) and forwards to the root.
- The package-level default `Registry` is a plain package var created at
  init with a stderr bootstrap root, so `For` and the other package-level
  entry points work before `Configure` runs, including at package-init
  time. The `atomic.Pointer` lives on the `Registry`'s root field, so
  `Configure` swaps the root handler in place via `SetRoot`, leaving the
  logger cache and per-name overrides intact.

**logrus -> slog bridge** (`bridge.go`)

- `slogBridge` (a `logrus.Hook`) copies each `logrus.Entry` into a
  `slog.Record` and dispatches via `RootHandler.Handle`. Fatal and Panic
  route through `SyncEmit` for durability before the process exits.
- `Install` / `InstallTo` redirect logrus to `io.Discard` with a noop
  formatter, set its level, and register the bridge. logrus's own
  pre-filter still keeps below-threshold records from reaching the
  bridge.
- Bidirectional level mappings between `logrus.Level` and `slog.Level`
  for the seven canonical names.
- `SyncEmit` package function type-asserts the root to `*AsyncHandler` so
  Fatal/Panic records flush the queued records and then write
  synchronously, falling back to `Handle` when the root is any other
  handler.

**Format compatibility** (`format.go`)

- `ReplaceAttr` produces pfxlog-shape JSON: lowercase level via
  `LevelName`, nested source suppressed at the top level.
- `sourceFlattener` decodes the record's PC into flat `file` and `func`
  attrs. Direct slog records carry a clean PC and are decoded here.
  Bridged-logrus records instead arrive with `PC == 0`: logrus's
  `Entry.Caller.PC` is a symbolized frame PC that does not re-decode
  reliably, so the bridge derives `file`/`func` from the resolved
  `Entry.Caller` and attaches them as attrs, which `sourceFlattener`
  passes through unchanged.
- `BuildHandler(out, opts)` constructs the production chain
  (`AsyncHandler -> sourceFlattener -> JSONHandler`) over a
  caller-supplied `io.Writer` (default `os.Stderr`).

## Not in scope

- df/dl pretty handler integration and TTY-vs-redirected format default.
- Wiring `Install` into ziti's controller + router `main` and adding the
  AsyncOptions flags to cobra commands.
- Golden-file format tests with the full fixture set.
- Startup invariant tests.
- The agent v2 callbacks driving `SetGlobalLevel` and `SetNamedLevel`.

All follow-up.

### DIFF
--- a/common/logging/bridge.go
+++ b/common/logging/bridge.go
@@ -1,0 +1,190 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package logging
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log/slog"
+
+	"github.com/sirupsen/logrus"
+)
+
+// noopFormatter is the formatter logrus uses after Install. The actual
+// dispatch happens in the slogBridge hook; the formatter's job is just to
+// return an empty payload so logrus's own write (to io.Discard) does nothing.
+type noopFormatter struct{}
+
+func (noopFormatter) Format(*logrus.Entry) ([]byte, error) {
+	return nil, nil
+}
+
+// slogBridge is the logrus.Hook that copies each logrus.Entry into a
+// slog.Record and dispatches it onto the configured root handler.
+// Fatal and Panic records bypass the async queue via SyncEmit so they are
+// durable before logrus exits the process.
+type slogBridge struct{}
+
+func (slogBridge) Levels() []logrus.Level {
+	return logrus.AllLevels
+}
+
+func (slogBridge) Fire(e *logrus.Entry) error {
+	level := logrusToSlog(e.Level)
+	// logrus has already resolved the caller into e.Caller (Function, File,
+	// Line). We must NOT forward e.Caller.PC and let a downstream handler
+	// re-decode it: that PC is a symbolized runtime.Frame.PC, and re-running it
+	// through runtime.CallersFrames resolves to the wrong function under inline
+	// expansion (e.g. logrus.NewEntry instead of the real call site). Instead
+	// the bridge carries the resolved location as flat file/func attrs and
+	// leaves PC == 0, which is the contract sourceFlattener and PrettyHandler
+	// expect for bridged records; direct slog records keep their PC and are
+	// decoded by those handlers instead.
+	r := slog.NewRecord(e.Time, level, e.Message, 0)
+	if e.Caller != nil {
+		r.AddAttrs(
+			slog.String("file", fmt.Sprintf("%s:%d", e.Caller.File, e.Caller.Line)),
+			slog.String("func", e.Caller.Function),
+		)
+	}
+	for k, v := range e.Data {
+		r.AddAttrs(slog.Any(k, v))
+	}
+	ctx := e.Context
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	if level >= LevelFatal {
+		return SyncEmit(ctx, r)
+	}
+	return RootHandler().Handle(ctx, r)
+}
+
+// Install wires logrus.StandardLogger() into the async slog sink: it
+// Configure's the default Registry with handler, sets the global slog level
+// (which drives both slog and logrus in lockstep), redirects logrus output to
+// io.Discard, replaces its formatter with a noop, enables ReportCaller so the
+// bridge can forward a real PC to slog, and registers the slogBridge hook so
+// every legacy log call routes through slog.
+//
+// Registering the hook replaces logrus's entire hook set, so any logrus hooks
+// added before Install are discarded. ziti adds none; a caller that does must
+// re-add them after Install.
+//
+// Calling Install more than once just replaces state; tests can call it
+// freely. Anything that runs after Install must not later mutate logrus's
+// output, formatter, or ReportCaller, or the bridge invariant breaks.
+func Install(handler slog.Handler, initialLevel slog.Level) {
+	InstallTo(logrus.StandardLogger(), handler, initialLevel)
+}
+
+// InstallTo is the testable form of Install: it wires the given *logrus.Logger
+// instead of the global standard logger. Production code calls Install; tests
+// pass logrus.New() so they can exercise the bridge without mutating global
+// logrus state.
+func InstallTo(target *logrus.Logger, handler slog.Handler, initialLevel slog.Level) {
+	Configure(handler)
+	target.SetOutput(io.Discard)
+	target.SetFormatter(noopFormatter{})
+	target.SetReportCaller(true)
+	// Replace existing hooks before adding ours so repeated Install calls
+	// don't stack multiple slogBridge instances on the same logger. The
+	// quickstart exercises this by running the controller and router from
+	// the same process, each through their own PreRun.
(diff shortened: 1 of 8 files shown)


common/logging/bridge.go
```diff
@@ -0,0 +1,173 @@
...
+func InstallTo(target *logrus.Logger, handler slog.Handler, initialLevel slog.Level) {
+	Configure(handler)
+	target.SetOutput(io.Discard)
+	target.SetFormatter(noopFormatter{})
+	target.SetReportCaller(true)
+	// Replace existing hooks before adding ours so repeated Install calls
+	// don't stack multiple slogBridge instances on the same logger. The
+	// quickstart exercises this by running the controller and router from
+	// the same process, each through their own PreRun.
+	target.ReplaceHooks(logrus.LevelHooks{})
```
andrewpmartinez, line 109: Not an issue, but this does remove all hooks from any other caller. Since this is documented as a `testable` version, I don't suppose it is a problem.

Worth documenting, though.

common/logging/registry.go
```diff
@@ -0,0 +1,301 @@
...
+// same instance throughout the test binary; only its internals are
+// reset.
+func resetDefaultForTest() {
+	defaultRegistry.mu.Lock()
+	defaultRegistry.overrides = map[string]*slog.LevelVar{}
+	defaultRegistry.loggerCache = map[string]*slog.Logger{}
+	defaultRegistry.mu.Unlock()
+	defaultRegistry.SetRoot(discardHandler{})
+	defaultRegistry.global.Set(slog.LevelInfo)
+}
```
andrewpmartinez: Can't these be moved to the _test file?

common/logging/format.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (h *sourceFlattener) Handle(ctx context.Context, r slog.Record) error {
+	if r.PC == 0 {
+		return h.parent.Handle(ctx, r)
+	}
+	frames := runtime.CallersFrames([]uintptr{r.PC})
+	frame, _ := frames.Next()
+	r2 := slog.NewRecord(r.Time, r.Level, r.Message, 0)
+	r2.AddAttrs(slog.String("file", fmt.Sprintf("%s:%d", frame.File, frame.Line)))
```
andrewpmartinez, line 81: Does this trim the file names like the older logs? It looks like a full path here.

andrewpmartinez (approved):
Some small nits.

Unverified findings sent in private.

common/logging/bridge.go
plorenz: Added a note on the Install method

common/logging/registry.go
plorenz: Moved

common/logging/format.go
plorenz: Only the console logger trimmed, the json logger didn't, so this maintains that behavior. The pfxlog equivalent in the follow-up pr needs to maintain the trim functionality

PR #3911, opened 2026-06-01
Install slog and route agent log-level callbacks through common/logging

### LINKED ISSUE #3910: Install slog and route agent log-level callbacks through common/logging
Phase 7 of the logging refactor described in [doc/design/logging-refactor.md](doc/design/logging-refactor.md) and [doc/design/logging-refactor-progress.md](doc/design/logging-refactor-progress.md). Stacks on #3906.

## Scope

Wire the foundation into the ziti binaries so slog becomes the default logger under the hood, while keeping operator-visible output the same. Specifically:

- `common/logging.SetGlobalLevel` becomes the lockstep entry point that drives both the default Registry global and `logrus.SetLevel(slogToLogrus(level))`. Operator surfaces (agent callbacks, `--verbose`, future flags) go through this one function so the two worlds never drift.
- `common/agentlog.DefaultLogLevelCallbacks` is rewritten to call `logging.SetGlobalLevel` / `SetNamedLevel` / `ClearNamedLevel` instead of mutating logrus and pfxlog directly.
- Controller, router, and tunnel PreRun call `logging.Install` with a handler built by `logging.BuildHandlerForFormat` so the bridge invariant is in place before the agent registers callbacks. AsyncOptions queue/threshold/summary flags are exposed via `logging.AddFlags` on each persistent flag set.
- A df/dl pretty handler (`logging.BuildPrettyHandler`) is added so the default look-and-feel matches the pre-slog binaries. `--log-formatter=json` continues to produce JSON via the existing `BuildHandler`.

## Behavior change

Per-channel overrides become slog-only. `ziti agent set-channel-log-level <name> debug` lifts that channel above the global only for code that calls `logging.For(name)`; legacy `pfxlog.Logger()` / `pfxlog.ChannelLogger(name)` call sites stay at the global level. This is the migration carrot from the design: convert a package to `logging.For` to gain per-channel debug overrides.

Output format is preserved.

## Acceptance

- `TestInstallInvariant` — post-Install, `logrus.StandardLogger` has Out=io.Discard, noop formatter, ReportCaller=true, and level mirrors the slog global; subsequent `SetGlobalLevel` keeps the two in lockstep.
- `TestEndToEnd_AgentSetLogLevel` — the `ziti agent set-log-level info` path moves both globals; pfxlog Info and `logging.For` Info both flow through the bridge; Debug stays filtered.
- `TestPerChannelOverride_AppliesToSlogOnly_NotPfxlog` — slog channel override lifts `logging.For("test.gossip").Debug` above the global; same call against `pfxlog.Logger()` / `pfxlog.ChannelLogger("test.gossip")` stays filtered; Clear restores the channel to the global.
- Existing `TestEdgeQuickstartAutomated` covers the same-process controller + router Install path (idempotent thanks to `InstallTo`'s `ReplaceHooks`).

plorenz:
## Summary

Stacked on #3907. Wires the `common/logging` foundation from the prior
stack into the ziti binaries so slog becomes the default logger under
the hood, while keeping operator-visible output the same.

Fixes #3910.

**Base is `slog-named-loggers`** so the diff shows only the new work.
Once #3907 merges to `main`, this rebases cleanly.

## What's in this PR

**Lockstep `SetGlobalLevel` + agent callback rewire** (`common/logging`,
`common/agentlog`)

- The package-level `logging.SetGlobalLevel(level)` now drives both the
  default Registry global and `logrus.SetLevel(slogToLogrus(level))`,
  so operator surfaces (agent callbacks, `--verbose`, future flags)
  set both worlds in one call.
- `agentlog.DefaultLogLevelCallbacks` is rewritten to call
  `logging.SetGlobalLevel` / `SetNamedLevel` / `ClearNamedLevel`
  instead of mutating logrus and pfxlog directly. A small
  `agentToSlog` helper maps the seven canonical levels with an Info
  fallback for forward-compat.

**Bespoke pretty handler** (`common/logging`)

- New `BuildPrettyHandler(out, opts, *PrettyOptions)` wraps the
  hand-rolled `logging.PrettyHandler` (a direct port of pfxlog's pretty
  output) in the `AsyncHandler` chain. We dropped the df/dl pretty handler
  after review found behavior gaps its options couldn't reach (blank
  labels for the non-standard Trace/Fatal/Panic levels, among others), so
  there is no `github.com/michaelquigley/df` dependency. See "Why we
  hand-roll the pretty handler" in `doc/design/logging-refactor.md`.
- New `BuildHandlerForFormat(out, opts, format)` picks the leaf
  handler by name; `""` and `"pfxlog"` produce pretty output, `"text"`
  is pretty with color off, `"json"` is the existing
  `BuildHandler` JSON path.

**Binary wiring** (`ziti/run`, `ziti/tunnel`)

- `Options.PreRun` (controller, router) and `rootPreRun` (tunnel) build
  the slog chain via `logging.BuildHandlerForFormat` driven by the
  existing `--log-formatter` flag, then call `logging.Install`.
- `--verbose` now seeds the slog initial level to Debug instead of
  mutating logrus directly.
- AsyncOptions queue / threshold / summary flags exposed via
  `logging.AddFlags` on each persistent flag set.

## Behavior change

Per-channel overrides become **slog-only**.
`ziti agent set-channel-log-level <name> debug` lifts the channel
above the global only for code that calls `logging.For(name)`; legacy
`pfxlog.Logger()` / `pfxlog.ChannelLogger(name)` call sites stay at
the global level. This is the migration carrot from the design:
convert a package to `logging.For` to gain per-channel debug
overrides. Output format is preserved (pfxlog-shape pretty by
default; JSON via `--log-formatter=json`).

### DIFF
--- a/common/logging/pretty.go
+++ b/common/logging/pretty.go
@@ -1,0 +1,302 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package logging
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"runtime"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+)
+
+// ANSI color sequences matching the mgutz/ansi values the old pfxlog
+// formatter used, so colored output looks the same as the pre-slog binaries.
+const (
+	ansiRed        = "\033[31m"
+	ansiYellow     = "\033[33m"
+	ansiWhite      = "\033[37m"
+	ansiBlue       = "\033[34m"
+	ansiLightBlack = "\033[90m"
+	ansiCyan       = "\033[36m"
+	ansiLightCyan  = "\033[96m"
+	ansiDefaultFg  = "\033[39m"
+)
+
+// Special attr keys carried over from pfxlog: ChannelsKey holds a []string of
+// channel names rendered as |a, b| after the function, ContextKey holds a
+// string rendered as [ctx]. Both are excluded from the fields block.
+const (
+	ChannelsKey = "_channels"
+	ContextKey  = "_context"
+)
+
+// PrettyOptions configures PrettyHandler's human-readable output.
+type PrettyOptions struct {
+	// AbsoluteTime renders the record's time as a wall-clock timestamp using
+	// TimestampFormat instead of seconds since StartTimestamp.
+	AbsoluteTime bool
+
+	// StartTimestamp is the baseline for the relative [seconds] time column.
+	// DefaultPrettyOptions sets it to the start of today in local time,
+	// matching pfxlog's StartingToday behavior.
+	StartTimestamp time.Time
+
+	// TimestampFormat is the time layout used when AbsoluteTime is set.
+	TimestampFormat string
+
+	// TrimPrefix is removed from the front of function names before they are
+	// rendered. DefaultPrettyOptions sets "github.com/openziti/".
+	TrimPrefix string
+
+	// UseColor enables ANSI coloring of every colored segment (level label,
+	// timestamp, function, fields). When false the output contains no escape
+	// sequences at all.
+	UseColor bool
+}
+
+// DefaultPrettyOptions returns PrettyOptions matching the pre-slog pfxlog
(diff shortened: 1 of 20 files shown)


common/logging/pretty.go
```diff
@@ -0,0 +1,308 @@
...
+// defaults: relative time since the start of today, "github.com/openziti/"
+// trimmed from function names, and color decided by the PFXLOG_USE_COLOR
+// environment variable when set, otherwise by whether out is a terminal.
+func DefaultPrettyOptions(out io.Writer) *PrettyOptions {
+	now := time.Now()
+	return &PrettyOptions{
+		StartTimestamp:  time.Date(now.Year(), now.Month(), now.Day(), 0, 0, 0, 0, now.Location()),
+		TimestampFormat: "2006-01-02 15:04:05.000",
+		TrimPrefix:      "github.com/openziti/",
+		UseColor:        useColor(out),
```
andrewpmartinez: useColor will result in color terms in scenarios that previously wouldn't. Not sure it is a bug, but it would appear that color will be used in more places than before.

andrewpmartinez (approved):

common/logging/pretty.go
plorenz: fixed, to make it consistent with existing behavior

PR #3913, opened 2026-06-01
fix oidc auth fails with wildcard server-cert SANs

### LINKED ISSUE #3891: oidc auth fails with wildcard server-cert SANs
changes to the way ziti login works have caused a regression and it is no longer possible to log in to the controller with a LE-backed alt-server-cert. 

**Steps to reproduce:**

* get a LE cert
* configure controller with LE cert
* try to use LE-based url to ziti login
* observe: "failed to verify certificate"

dovholuknf:
fix openziti/ziti#3891: oidc auth fails with wildcard server-cert SANs. add wildcard-cert discovery test + keep wildcard SANs as issuers


controller/webapis/oidc-api.go
```diff
@@ -233,9 +233,9 @@ func getPossibleIssuers(id identity.Identity, bindPoints []xweb.BindPoint) []oid
 		}
 		curServerCert := curServerCertChain[0]
 		for _, dnsName := range curServerCert.DNSNames {
-			if strings.HasPrefix(dnsName, "*.") {
-				continue // wildcard DNS names produce unusable OIDC issuer URLs
-			}
+			// Certs with wildcard DNS SANs (e.g. "*.example.com") are valid and at request time
+			// issuer URL is built from the request's Host (e.g. https://foo.example.com/oidc)
+			// the literal "*" should never be used in an issuer URL (see isWildcardIssuer).
```
andrewpmartinez: `isWildcardIssuer` doesn't seem to exist?

tests/testdata/configs/wildcard-oidc-server/ctrl.yml
```diff
@@ -0,0 +1,62 @@
+v: 3
+
+db: ${ZITI_TEST_DB}
+
+# Models a controller fronted by a wildcard (e.g. LetsEncrypt) certificate supplied via alt_server_certs.
+# The primary server_cert is an ordinary cert (SANs: localhost / loopback IPs), standing in for a
+# quickstart/host cert. The alt server cert's only SAN is the wildcard *.wildcard.test, standing in for the
+# LetsEncrypt cert. The edge-oidc API derives its OIDC issuers from the SANs of ALL active server certs
+# (identity.GetX509ActiveServerCertChains, which includes alt certs), so the wildcard becomes an issuer
+# served via a request-host-derived issuer URL. See controller/oidc_auth/provider.go (isWildcardIssuer /
```
andrewpmartinez: Again, I do not seem to be able to find `isWIldCardIssuer`

tests/testdata/configs/README.md
```diff
@@ -43,6 +43,43 @@ testdata/configs/
...
+  -key tests/testdata/ca/intermediate/private/ctrl.key.pem \
+  -subj "/C=US/ST=North Carolina/L=Charlotte/O=NetFoundry/OU=Ziti Fabric/CN=ctrl" \
+  -out /tmp/wildcard.csr
+MSYS_NO_PATHCONV=1 openssl x509 -req -in /tmp/wildcard.csr \
+  -CA tests/testdata/ca/intermediate/certs/intermediate.cert.pem \
+  -CAkey tests/testdata/ca/intermediate/private/intermediate.key.decrypted.pem \
+  -CAserial /tmp/wildcard.srl -CAcreateserial \
+  -days 3650 -sha256 -extfile /tmp/wildcard-san.cnf -extensions v3_req \
+  -out tests/testdata/ca/intermediate/certs/ctrl-server-wildcard.cert.pem
+```
```
andrewpmartinez: Remove this. Not necessary.

controller/webapis/oidc-api.go
dovholuknf: oh yah - i'll clean the comment up. it was a func that was there but was only used in one place so i removed it

controller/oidc_auth/provider.go
```diff
@@ -157,7 +158,16 @@ func newOidcProvider(_ context.Context, issuer string, oidcConfig Config) (op.Op
...
 
-	handler, err := op.NewProvider(config, oidcConfig.Storage, op.StaticIssuer(issuer))
+	// A wildcard issuer (e.g. https://*.example.com/oidc) can't be a static URL; derive it from the
+	// request host. The dispatcher only routes here when the host matches the wildcard via Issuer.ValidFor.
+	var issuerSource func(bool) (op.IssuerFromRequest, error)
+	if strings.HasPrefix(issuer, "https://*.") {
+		issuerSource = op.IssuerFromHost("/oidc")
+	} else {
+		issuerSource = op.StaticIssuer(issuer)
+	}
```
andrewpmartinez: Hmmmm, this is problematic. It allows any client to request a token issued for any hostname. Not just the hostnames we are allowed. If someone shares keys for their servers and another JWT-inspecting application uses the same signer, this could be problematic.

dovholuknf: "Not just the hostnames we are allowed" - where and how are those hostnames identified? I don't know of anywhere that we configure it other than maybe the advertised address in the config? It'd be easy enough to bump a wildcard against the config somewhere to ensure it's an expected address assuming the config is available in that way? That what you're thinking?

andrewpmartinez: Ok, so if someone wants to use a wildcard DNS SAN, we need to know what issues they want to use, and just using "any" is not a great move.

My suggestion:

  - Add an `allowedHostnames` option to the `edge-oidc` binding, parsed next to `redirectURIs` in `NewOidcApiHandler` (`controller/webapis/oidc-api.go:148`) and threaded into `getPossibleIssuers`. Exact hostnames only, no `*`/patterns.
  - In `getPossibleIssuers` (`controller/webapis/oidc-api.go:199`): leave concrete DNS/IP SAN expansion as-is; expand a wildcard SAN only to `allowedHostnames` entries that match it via `x509.VerifyHostname`. Never emit a literal-`*` issuer.
  - Drop the dynamic path in `newOidcProvider` (`controller/oidc_auth/provider.go:161-168`). Every issuer is now a concrete host → `op.StaticIssuer`, so no `Forwarded`-header trust and a closed set of valid `iss` values.
  - Validate each `allowedHostnames` entry is covered by an active server cert SAN; if a wildcard SAN exists but `allowedHostnames` is unset, ignore the wildcard and warn.
  - Remove the dangling `isWildcardIssuer` references (`controller/webapis/oidc-api.go:238`,
  `tests/testdata/configs/wildcard-oidc-server/ctrl.yml:10`).

andrewpmartinez (changes requested):
See comments.

controller/oidc_auth/provider.go
dovholuknf: i believe it's not implemented in a way that's acceptable but you'll tell me! :)

ekoby (approved):

andrewpmartinez (dismissed):
The ctrl example files in /etc need to be updated w/ the new configuration options.

dovholuknf:
> The ctrl example files in /etc need to be updated w/ the new configuration options.

Oh good catch thanks! I'd still like to see us normalize on the same thing the ziti config generator/quickstart uses but i'll do it in both places

etc/ctrl.with.edge.yml
```diff
@@ -389,6 +389,9 @@ web:
       - binding: edge-client
       - binding: edge-oidc
         options:
+          # allowedHostnames (optional): a wildcard server-cert SAN serves OIDC only for the exact hostnames listed here.
+          # allowedHostnames:
+          #   - ctrl.example.com
```
dovholuknf, line 394: @andrewpmartinez - just this one is adequate?

andrewpmartinez (approved):

PR #3917, opened 2026-06-01
Convert router/forwarder to slog

### LINKED ISSUE #3916: Convert router/forwarder to slog
First chunk of the post-foundation slog conversion described in
[doc/design/slog-conversion-plan.md](doc/design/slog-conversion-plan.md).
Stacks on the foundation work in #3910.

## Scope

Convert `router/forwarder/` from pfxlog/logrus to `common/logging`. Three
files, three channels (justified by the central-square log evidence):

- `router.forwarder.route` — forwarder.go: circuit routing, payload /
  acknowledgement / control forwarding. Per-payload hot path with an
  explicit `Enabled` gate around the Debug emit.
- `router.forwarder.scan` — scanner.go: idle-circuit scanner. Dominant
  volume contributor in real traffic (~260k records per central-square
  dump); separable channel so operators can silence it without losing
  the data-plane channels.
- `router.forwarder.fault` — faulter.go: circuit and link fault
  reporting. Diagnostic surface between data-plane and controllers.

## Rationale

The central-square router log dumps (~1.35M records each) show
`router/forwarder` is the single highest-volume contributor — Unroute +
Scanner.scan + Faulter.run together account for ~41% of all records.
Converting it first delivers the most operator-visible benefit per
unit of conversion work.

The three-channel split is justified per the conversion plan's
"Sub-channels appear only when there's a specific operational reason
(high volume, separable failure modes)" rule: scan volume dominates,
fault diagnostics are independent of forwarding, and operators
benefit from being able to mute one without losing the others.

## Acceptance

- All `pfxlog` and direct `logrus` call sites in `router/forwarder/`
  are removed.
- Each file declares a package-scoped logger via `logging.For(name)`
  with godoc explaining the channel choice.
- Per-payload hot-path Debug calls are gated with `Enabled` so the
  cost when below Debug is a single level comparison.
- Levels preserved — conversion is mechanical, not a level audit.
- No new Warn/Error introduced at per-event rate.
- `go test ./router/forwarder/ -race` passes.

plorenz:
First chunk of the post-foundation slog conversion. Stacked on #3911.

Fixes #3916.

## What this PR does

Converts `router/forwarder/` (3 files, ~50 pfxlog/logrus call sites)
to \`common/logging\`, splitting into three channels:

| File | Channel | Why a dedicated channel |
|---|---|---|
| forwarder.go | \`router.forwarder.route\` | Per-payload data-plane hot path; gated \`Enabled\` check before Debug emit |
| scanner.go | \`router.forwarder.scan\` | Dominant log-volume contributor (~260k records/dump in real traffic); separable so operators can silence scan without losing forward/fault |
| faulter.go | \`router.forwarder.fault\` | Diagnostic surface between data-plane and controllers; isolated so a noisy fault loop doesn't drown out the others |

The three-channel split is justified per the conversion plan's
\"sub-channels only when there's a specific operational reason
(high volume, separable failure modes)\" rule.

## Patterns demonstrated

- \`var log = logging.For(\"name\")\` at package scope. Safe pre-Configure
  because the foundation's default Registry is initialized at package
  init with a stderr bootstrap root (per #3911 follow-on).
- WithField chains -> variadic key/value pairs.
- Scoped \`log.With(...)\` for loops where multiple emits share attrs.
- Hot-path Debug guarded by \`log.Enabled(...)\` so the cost when below
  Debug is one level comparison plus an atomic load.
- Channel-name godoc on each \`var log\` declaration so operators and
  contributors can find the toggle without grep.

## Operator usage

Once this PR (and the foundation stack) lands, operators get:

\`\`\`
ziti agent set-channel-log-level router.forwarder.scan warn   # quiet the scan firehose
ziti agent set-channel-log-level router.forwarder.route debug # surgical per-packet debug
ziti agent clear-channel-log-level router.forwarder.fault     # back to global
\`\`\`

### DIFF
--- a/router/forwarder/forwarder.go
+++ b/router/forwarder/forwarder.go
@@ -17,22 +17,35 @@
 package forwarder
 
 import (
+	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"sync/atomic"
 	"time"
 
-	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/foundation/v2/info"
+	"github.com/openziti/foundation/v2/uuidz"
 	"github.com/openziti/metrics"
 	"github.com/openziti/sdk-golang/xgress"
 	"github.com/openziti/ziti/v2/common/inspect"
+	"github.com/openziti/ziti/v2/common/logging"
 	"github.com/openziti/ziti/v2/common/pb/ctrl_pb"
 	"github.com/openziti/ziti/v2/common/trace"
 	"github.com/openziti/ziti/v2/router/env"
 	"github.com/openziti/ziti/v2/router/xlink"
-	"github.com/sirupsen/logrus"
 )
+
+// routeLog is the logger for the forwarder's circuit-routing and
+// payload/control-forwarding paths. Its channel name is
+// "router.forwarder.route"; operators adjust route visibility with
+// `ziti agent set-channel-log-level router.forwarder.route <level>`.
+//
+// The forward path is per-payload hot: every record on this channel runs
+// inside the data-plane critical section. The bridged-logrus and direct
+// slog channels both flow through the same AsyncHandler queue, so the
+// per-record cost is the same; level-gating is what keeps it cheap.
+var routeLog = logging.For("router.forwarder.route")
 
 // InvalidLinkDestinationError indicates a route referenced a link destination that does not exist.
 type InvalidLinkDestinationError struct {
@@ -90,7 +103,7 @@ func (forwarder *Forwarder) StartScanner(ctrls env.NetworkControllers) {
 	if scanner.interval > 0 {
 		go scanner.run()
 	} else {
-		logrus.Warnf("scanner disabled")
+		routeLog.Warn("scanner disabled")
 	}
 }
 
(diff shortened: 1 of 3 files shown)


router/forwarder/forwarder.go
```diff
@@ -207,7 +217,17 @@ func (forwarder *Forwarder) forwardPayload(srcAddr xgress.Address, payload *xgre
...
+				// the level check is the only cost when this channel is below Debug.
+				if routeLog.Enabled(context.Background(), slog.LevelDebug) {
+					routeLog.Debug("forwarded payload",
+						"srcAddr", string(srcAddr),
+						"dstAddr", string(dstAddr),
+						"circuitId", circuitId,
+						"seq", payload.Sequence,
+						"origin", payload.GetOriginator(),
+					)
+				}
```
andrewpmartinez, line 237: Drops the conditional UUID field GetLoggerFields() used to provide. Intentional?

router/forwarder/forwarder.go
```diff
@@ -248,12 +272,13 @@ func (forwarder *Forwarder) ForwardAcknowledgement(srcAddr xgress.Address, ackno
 
 func (forwarder *Forwarder) ForwardControl(srcAddr xgress.Address, control *xgress.Control) error {
 	circuitId := control.CircuitId
-	log := pfxlog.ContextLogger(string(srcAddr)).WithField("circuitId", circuitId)
+	log := routeLog.With("srcAddr", string(srcAddr), "circuitId", circuitId)
 
 	var err error
+	var dstAddr xgress.Address
```
andrewpmartinez: Odd hoist choice here. Has no effect, but if the hoist isn't referenced outside the blocks it is used in, it doesn't matter and only potentially causes bugs in the future.

router/forwarder/scanner.go
```diff
@@ -50,8 +58,8 @@ func (self *Scanner) setCircuitTable(circuits *circuitTable) {
 }
 
 func (self *Scanner) run() {
-	logrus.Info("started")
-	defer logrus.Warn("exited")
+	scanLog.Info("started")
+	defer scanLog.Warn("exited")
```
andrewpmartinez, line 62: If this isn't a singleton, should these have some kind of identifier just in case they interleave logs?

andrewpmartinez (approved):
Comments left for you to consider, lgtm as is but confirm/deny fix as needed.

router/forwarder/forwarder.go
plorenz: fixed

router/forwarder/forwarder.go
plorenz: fixed

router/forwarder/scanner.go
plorenz: it's a singleton in practice

PR #3918, opened 2026-06-02
Add role-attribute usage queries. Fixes #1593

### LINKED ISSUE #1593: Expanded attribute query support in management API; add policy attribute support and usage count
The current API supports retrieval of a list of distinct role attributes which have been declared on an Identity, Edge Router, or Service entity.  For example, a `GET` request to `/identity-role-attributes` returns an array of all attribute values (strings) declared across identities (a distinct set.)

The results of the existing API endpoints do not include attribute values that appear only on a policy.

I'd like to be able to retrieve attribute values that appear on a policy, and it'd be nice if each resulting attribute included a count of its usage.  That is, if `foo` is a role attribute on 3 Identities, then the usage count is `3`.  If querying role attributes on a policy, then the usage count would be the number of policies with that role attribute value.

Having a usage count will eliminate N followup API requests to get the set of entities with a given attribute value (once per N distinct attributes.)  While these followup requests can be individually limited to 1 result (by reading the total count from the pagination data), it still results in a potentially large number of API calls just to get the count of usage for each distinct attribute value.

While the count is needed, having an array of entity ids would be even better.  But I certainly understand the possible performance impact that might create.

Having a usage count enables an "Attribute Explorer" interface, where distinct attributes are listed along a left filter bar along with a count of entity matching that attribute.

plorenz:
- adds management API endpoints listing role-attribute usage for identities, edge routers, services, and posture checks, reporting per-source counts (and optionally ids via withIds) across home-entity collections and the policies that reference each attribute
- adds RoleAttributeUsage model with QueryRoleAttributeUsage, mapping each RoleAttributeKind to its contributing sources, with all reads in a single transaction so counts stay consistent with the attribute list; returns an empty result (not a panic) for kinds with no attributes
- adds boltz.SetIndexValueTransform and AddSetIndexWithTransform, letting a SetIndex filter and rewrite symbol values without persisting a derived field
- adds derived role-attribute set indexes to the service policy, edge router policy, and service edge router policy stores, built from existing role fields via a role-attribute-only transform that excludes the #all wildcard
- adds v45 migration that backfills the new indexes via each index's own CheckIntegrity(fix=true), scoped to the role-attribute indexes only, with a per-index summary log
- bumps edge-api to v0.31.0 for the new API operations

### DIFF
--- a/tests/role_attribute_usage_test.go
+++ b/tests/role_attribute_usage_test.go
@@ -1,0 +1,387 @@
+//go:build apitests
+
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package tests
+
+import (
+	"fmt"
+	"net/http"
+	"sort"
+	"testing"
+
+	"github.com/openziti/edge-api/rest_management_api_client/role_attributes"
+	"github.com/openziti/edge-api/rest_model"
+	"github.com/openziti/foundation/v2/util"
+	"github.com/openziti/ziti/v2/common/eid"
+)
+
+// usageFilter returns a filter restricting role-attribute usage results to
+// attributes carrying the given unique test prefix, sorted by id so result
+// order is deterministic.
+func usageFilter(prefix string) *string {
+	return util.Ptr(fmt.Sprintf(`id contains "%s" sort by id`, prefix))
+}
+
+// usageByAttr returns the usage detail for the given role attribute, or nil
+// if the list doesn't contain it.
+func usageByAttr(items rest_model.RoleAttributeUsageList, attr string) *rest_model.RoleAttributeUsageDetail {
+	for _, item := range items {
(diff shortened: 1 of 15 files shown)


tests/style.md
```diff
@@ -0,0 +1,197 @@
+# Test Style Guide for `tests/`
```
andrewpmartinez: Did you mean to include this?

plorenz: Yes. I figured since we're trying to move our tests to a better convention, this would be helpful. Let me know if you agree with the contents

tests/role_attribute_usage_test.go
```diff
@@ -0,0 +1,386 @@
...
+// usageFor returns the usage entry for one source on a detail, failing the
+// test with a diagnostic if the detail, source entry, or required count is
+// missing, so callers can dereference Count without nil checks.
+func (ctx *TestContext) usageFor(detail *rest_model.RoleAttributeUsageDetail, source string) rest_model.RoleAttributeSourceUsage {
+	ctx.Req.NotNil(detail, "expected a usage detail for source %s", source)
+	entry, ok := detail.Usage[source]
+	ctx.Req.True(ok, "usage map should contain source %s", source)
+	ctx.Req.NotNil(entry.Count, "count is required on source %s", source)
+	return entry
+}
```
andrewpmartinez, line 61: If you are going to write helpers that assert you should use `ctx.T().Helper()` to allow frame skipping in test output.

andrewpmartinez (approved):
Small comment on test helpers.

tests/style.md
andrewpmartinez: I'll just update it in another PR with the other things I have.

tests/role_attribute_usage_test.go
plorenz: fixed

tests/style.md
plorenz: removed since you posted a consolidated doc in another PR :+1:

Issue #3929, opened 2026-06-03
[Backport-2.0] Router does not enforce api-session or identity revocations on live connections


plorenz:
Backport to `release-v2.0.x` of #3927.

`CheckConnections` enforces only JWT expiry, so a revoked OIDC api-session or a disabled/deleted identity keeps its live circuits and hosted terminators until the access token expires. The router should also enforce the RouterDataModel `Revocations`:

- close connections whose api-session id is revoked (`API_SESSION`);
- close connections whose identity is revoked (`IDENTITY`), bounded by an `IssuedBefore` cutoff so a session re-authenticated after the cutoff survives;
- disabling/deleting an identity must produce an identity revocation (self-contained OIDC JWTs aren't stored, so cascade-delete doesn't reach them).

plorenz:
Merged to release-v2.0.x

PR #3922, opened 2026-06-02
Revalidate dial circuits and bind terminators on posture changes

plorenz:
## Summary

Router-side posture revalidation was incomplete: when access should be revoked due to posture, several active connections were left open.

1. **Posture-data changes covered only mux-sink dial conns.** `onPostureDataUpdate` iterated only the connId-keyed mux sinks (reading `ServiceSessionToken.ServiceId`), so it missed **SDK-hosted xgress dial circuits** (`xgCircuits`) and **hosted bind terminators**. A device that fell out of posture compliance kept those circuits/terminators alive.
2. **Posture-requirement changes weren't reacted to at all.** Tightening a service policy's posture requirements — adding/removing a posture check, or editing a check definition — updated the model but never re-evaluated active circuits, since the service-membership diff doesn't surface posture changes (`IdentityService` carries no posture info). They lingered until the next posture-data report, a fresh dial, or token expiry.

## Changes

- `onPostureDataUpdate` now re-evaluates **both** dial circuits (including SDK-hosted xgress in `xgCircuits`) **and** hosted bind terminators, keyed on `GetServiceId` (which also fixes a latent nil `ServiceSessionToken` deref), closing denied access via `CloseForDialAccessLoss` / `CloseForBindAccessLoss`.
- `edgeClientConn` now handles `IdentityPostureChecksUpdatedEvent`, re-evaluating its circuits and terminators when the identity's posture requirements change. This rides the existing identity-subscription marker/scan pipeline, so it dedups and retries if the work queue is full.
- Adds the supporting seams: `state.DialCircuit` / `state.BindTerminator` views and `ConnProvider.IterateDialCircuits` / `IterateBindTerminators`, `getTerminatorsForConn`, and `edgeTerminator.CloseForBindAccessLoss`; refactors `handleBindAccessLost` onto the shared close path.

## Backport

This is the main-line fix and closes #3908. The targeted backport to `release-v2.0.x` is tracked separately by #3928 (both gaps are pre-existing — verified at `v2.0.0`).

PR #3930, opened 2026-06-03
Enforce api-session and identity revocations on the router

### LINKED ISSUE #3927: Router does not enforce api-session or identity revocations on live connections; revoked OIDC sessions keep dialing/hosting until access-token expiry
`ManagerImpl.CheckConnections()` enforces only JWT *expiry*, so a revoked OIDC api-session, or a disabled/deleted identity, keeps its live circuits and hosted terminators until the access token expires (up to the access-token TTL). The router should also enforce the RouterDataModel `Revocations`:

- close connections whose api-session id is revoked (`API_SESSION`);
- close connections whose identity is revoked (`IDENTITY`), bounded by an issue-time cutoff (`IssuedBefore`) so a session re-authenticated after the cutoff survives the still-lingering revocation;
- disabling or deleting an identity must produce an identity revocation: self-contained OIDC JWTs aren't stored, so the existing cascade-delete of legacy api-sessions doesn't reach them.

This tightens access-loss propagation to the ~5s `CheckConnections` interval instead of the access-token TTL.

Split from the (mis-framed, now closed) #3912, whose closing note deferred this work to the connect-v2 effort. Related: #2255 (controller exchange/refresh boundary respecting revocations and existing identities), #3908 (posture-driven revocation of SDK-hosted xgress/terminators).

plorenz:
## Summary

`ManagerImpl.CheckConnections()` (the ~5s router reaper) enforced only JWT *expiry*, so a revoked OIDC api-session, or a disabled/deleted identity, kept its live circuits and hosted terminators until the access token expired. This makes the router enforce the RouterDataModel `Revocations` directly, tightening access-loss propagation to the `CheckConnections` interval.

## Changes

**Revocation `Type` taxonomy.** Adds a `Type` field to `DataState_Revocation` (and the raft `Revocation` command proto), mirroring `rest_model.RevocationTypeEnum` (`API_SESSION` / `IDENTITY` / `JTI`) so the management API, OIDC producers, sync, and router enforcement share one vocabulary. `common.RevocationType*` constants are compile-time bound to the REST enum to prevent drift.

**Api-session revocation (e.g. OIDC end-session).**
- `TerminateSessionFromRequest` revokes the specific api-session named by the `z_asid` claim in the end-session request, so revocation is precise to that session (a re-authentication gets a fresh `z_asid` and is unaffected); falls back to an identity-scoped revocation when the hint carries none.
- `RouterDataModel.IsApiSessionRevoked`, enforced in `CheckConnections`.

**Identity revocation (disable / delete / identity-scoped).**
- Adds an `IssuedBefore` cutoff to the revocation (proto + db + model + raft marshalling): an `IDENTITY` revocation invalidates only sessions issued before the cutoff, so a session re-authenticated after a re-enable survives the still-lingering revocation.
- `RouterDataModel.IsIdentityRevoked(identityId, issuedAt)`, enforced in `CheckConnections`.
- Disabling or deleting an identity now produces an identity revocation via an `EntityConstraint` on the identity store, so the revocation is written **in the same transaction** as the identity change and can't be skipped from any code path. Self-contained OIDC JWTs aren't stored, so the existing cascade-delete of legacy api-sessions doesn't reach them.
- The management revocation API and OIDC identity fallback set the cutoff; `RevocationManager.CreateOrReplace` lets a repeat logout/termination refresh the cutoff rather than collide on the reused id.

Also carries the previously-dropped `Type` through the batched raft revocation marshalling.

## Stacking / dependency

Based on `posture-revalidation` (#3922) — please merge that first; this PR's diff will reduce to its own two commits once it does.

Fixes #3927. Backport to `release-v2.0.x` tracked by #3929.

### DIFF
--- a/controller/internal/routes/revocation_router.go
+++ b/controller/internal/routes/revocation_router.go
@@ -25,12 +25,12 @@ import (
 	"github.com/openziti/edge-api/rest_management_api_server/operations/revocation"
 	"github.com/openziti/edge-api/rest_model"
 	"github.com/openziti/foundation/v2/errorz"
-	"github.com/openziti/ziti/v2/controller/storage/boltz"
 	"github.com/openziti/ziti/v2/controller/env"
 	"github.com/openziti/ziti/v2/controller/model"
 	"github.com/openziti/ziti/v2/controller/models"
 	"github.com/openziti/ziti/v2/controller/permissions"
 	"github.com/openziti/ziti/v2/controller/response"
+	"github.com/openziti/ziti/v2/controller/storage/boltz"
 )
 
 func init() {
@@ -76,7 +76,7 @@ func (r *RevocationRouter) Detail(ae *env.AppEnv, rc *response.RequestContext) {
 }
 
 // Create handles POST /revocations. It validates the submitted id against the
-// revocation type, computes the expiry from the configured refresh token duration,
+// revocation type, computes the expiry from the maximum configured token duration,
 // and persists the revocation entry.
 func (r *RevocationRouter) Create(ae *env.AppEnv, rc *response.RequestContext, params revocation.CreateRevocationParams) {
 	Create(rc, rc, RevocationLinkFactory, func() (string, error) {
@@ -84,7 +84,7 @@ func (r *RevocationRouter) Create(ae *env.AppEnv, rc *response.RequestContext, p
 		if err != nil {
 			return "", err
 		}
-		if err = ae.Managers.Revocation.Create(entity, rc.NewChangeContext()); err != nil {
+		if err = ae.Managers.Revocation.CreateOrReplace(entity, rc.NewChangeContext()); err != nil {
 			return "", err
 		}
 		return entity.Id, nil
@@ -122,7 +122,8 @@ func mapCreateRevocationToModel(ae *env.AppEnv, r *http.Request, create *rest_mo
 		}
 	}
 
-	expiresAt := time.Now().Add(ae.GetConfig().Edge.Oidc.RefreshTokenDuration)
+	now := time.Now()
+	expiresAt := now.Add(ae.GetConfig().Edge.Oidc.MaxTokenDuration())
 
 	return &model.Revocation{
 		BaseEntity: models.BaseEntity{
@@ -131,5 +132,8 @@ func mapCreateRevocationToModel(ae *env.AppEnv, r *http.Request, create *rest_mo
 		},
 		ExpiresAt: expiresAt,
 		Type:      string(revocationType),
+		// IssuedBefore is the cutoff for IDENTITY revocations: sessions issued
+		// before now are revoked, while a later re-authentication survives.
+		IssuedBefore: now,
 	}, nil
 }
--- a/controller/model/revocation_manager.go
+++ b/controller/model/revocation_manager.go
@@ -21,13 +21,13 @@ import (
 	"fmt"
 	"time"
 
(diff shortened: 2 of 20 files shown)


controller/model/revocation_manager.go
```diff
@@ -56,6 +56,20 @@ func (self *RevocationManager) Create(entity *Revocation, ctx *change.Context) e
 	return DispatchCreate[*Revocation](self, entity, ctx)
 }
 
+// CreateOrReplace creates the revocation, first removing any existing revocation
+// with the same id. Identity- and api-session-scoped revocations are keyed by
+// the identity/session id, so a repeat (e.g. re-disabling an identity, or a
+// second logout) would otherwise collide with the lingering prior entry;
+// replacing it refreshes the ExpiresAt and IssuedBefore cutoff.
+func (self *RevocationManager) CreateOrReplace(entity *Revocation, ctx *change.Context) error {
```
andrewpmartinez, line 64: This seems like something that would be good to call from the Management API handlers as well. I think this probably still uses a raw "create".

controller/oidc_auth/storage.go
```diff
@@ -898,7 +898,42 @@ func (s *HybridStorage) TokenRequestByRefreshToken(_ context.Context, refreshTok
...
+	return s.saveOrReplaceRevocation(revocation)
+}
+
+// TerminateSessionFromRequest implements op.CanTerminateSessionFromRequest. The
+// op framework prefers it over TerminateSession when present. It revokes the
+// specific api-session being terminated, identified by the z_asid claim in the
+// end-session request's id_token_hint, so revocation is precise to this session
+// (a re-authentication gets a fresh z_asid and is unaffected). When the hint
+// carries no api-session id, it falls back to the identity-scoped revocation.
+func (s *HybridStorage) TerminateSessionFromRequest(_ context.Context, endSessionRequest *op.EndSessionRequest) (string, error) {
```
andrewpmartinez, line 915: Should add `var _ op.CanTerminateSessionFromRequest = (*HybridStorage)(nil)`

controller/internal/routes/revocation_router.go
```diff
@@ -122,7 +122,8 @@ func mapCreateRevocationToModel(ae *env.AppEnv, r *http.Request, create *rest_mo
 		}
 	}
 
-	expiresAt := time.Now().Add(ae.GetConfig().Edge.Oidc.RefreshTokenDuration)
+	now := time.Now()
+	expiresAt := now.Add(ae.GetConfig().Edge.Oidc.RefreshTokenDuration)
```
andrewpmartinez: There is a function on the `oidc_auth.Config` struct named `MaxTokenDuration()` that the oidc code uses to set default revocation times.

I don't know how the module dependencies would work out, but we should use similar logic, if not the same logic. I don't know if the function can be moved to `config.Config.Edge.Oidc` and have the `oidc_auth.Config` just refer to `config.Config.Edge.Oidc` directly.

andrewpmartinez (approved):
Some smaller comments for the unification of logic, type assertion on the implemented interface, etc.

Private comment sent as well for investigation.

controller/model/revocation_manager.go
plorenz: fixed

controller/oidc_auth/storage.go
plorenz: fixed

controller/internal/routes/revocation_router.go
plorenz: fixed

Issue #3233, opened 2025-08-20
JWT Token Not Properly Verified - SAST High Severity Finding


cx-constantino-antunes:
Hi, during a recent SAST scan I encountered a HIGH severity finding related to the router state manager. The [ParseJWT function](https://github.com/openziti/ziti/blob/cca364d8b7f03c3c96266f198fc65253413a4a49/router/state/manager.go#L454) is not checking the token.Valid field which could potentially allow the use of manipulated or invalid tokens.

Looks like this could enable the use of expired tokens, but could you clarify what functionality would be impacted by this vulnerability?

Thanks in advance for your help!

andrewpmartinez:
Thanks for the report, we will address this shorty.

cx-duarte-carneiro:
Hi @andrewpmartinez. Is this a security issue or a false positive?

andrewpmartinez:
Ah this issue didn't get tagged in the code changes. In newer version of the `manager.go` file, ParseJWT was replaced with two different functions `ParseApiSessionJwt` and `ParseServiceSessionJwt` that both check `token.IsValid`.

See: https://github.com/openziti/ziti/blob/95fe961c957f8b5eacf9baa976d1d1646c7a2929/router/state/manager.go#L791
See: https://github.com/openziti/ziti/blob/95fe961c957f8b5eacf9baa976d1d1646c7a2929/router/state/manager.go#L842

I am marking this issue resolved.

Thank you for your report!

cx-constantino-antunes:
> Ah this issue didn't get tagged in the code changes. In newer version of the `manager.go` file, ParseJWT was replaced with two different functions `ParseApiSessionJwt` and `ParseServiceSessionJwt` that both check `token.IsValid`.

Hi @andrewpmartinez I see it was fixed but the changes never made to v1.6. Since that's now the maintenance LTS, could you apply the fix there too?

andrewpmartinez:
> > Ah this issue didn't get tagged in the code changes. In newer version of the `manager.go` file, ParseJWT was replaced with two different functions `ParseApiSessionJwt` and `ParseServiceSessionJwt` that both check `token.IsValid`.
> 
> Hi @andrewpmartinez I see it was fixed but the changes never made to v1.6. Since that's now the maintenance LTS, could you apply the fix there too?

Yes we can. Adding to my list of things to do this week.

PR #3932, opened 2026-06-03
backport openziti/ziti#3233 to release-v1.6.x explicit token.Valid ch…

andrewpmartinez:
…eck in router ParseJwt

- adds an explicit jwtToken.Valid check to ParseJwt in the router state manager, mirroring the ParseApiSessionJwt/ParseServiceSessionJwt split on main; invalid tokens are already rejected by jwt.ParseWithClaims, so this is defense in depth that closes the SAST finding
- adds Test_Router_ParseJwt_TokenValidation, an apitests integration test that boots a controller and edge router and drives the router's ParseJwt: accepts a properly signed, unexpired access token and rejects tokens signed by an unknown key or carrying an expired claim

### DIFF
--- a/router/state/manager.go
+++ b/router/state/manager.go
@@ -464,6 +464,9 @@ func (sm *ManagerImpl) ParseJwt(jwtStr string) (*jwt.Token, *common.AccessClaims
 	}
 
 	if accessClaims.Type == common.TokenTypeAccess {
+		if !jwtToken.Valid {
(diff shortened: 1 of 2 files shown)


andrewpmartinez:
I couldn't use the same implementation from 2.x.x as its was nearly a complete rewrite of that area and the lift was too big. This change does the same thing.

plorenz (approved):

PR #3962, opened 2026-06-10
Add main build and LTS line selectors to setup-cli action

CLBRITTON2:
- setup-cli now takes a `ref` input to build the CLI from source at any branch, tag, or commit, and `version: active-lts` / `maint-lts` to fetch the latest release on a given LTS line. Those lines are read from a new `lts-versions.json` at the repo root.
- This lets any repo using setup-cli build ziti from any ref or pull a specific LTS release instead of hand-rolling that fetch in the repo's CI.

### DIFF
--- a/setup-cli/action.yml
+++ b/setup-cli/action.yml
@@ -5,13 +5,18 @@ inputs:
     required: false
     default: stable
     description: |
-      version of ziti CLI to download,
-      use tag regex (like v1.6) or special values:
-      'stable' (default) means latest release (GitHub marked `latest`),
-      'active-lts' / 'maint-lts' resolve the latest release on the LTS line named in the repo's lts-versions.json,
-      'main' builds the CLI from the openziti/ziti main branch instead of downloading a release,
-      regex can be used to pick latest release matching the pattern,
-      for example, `2.0.[0-9]+$` will match latest 2.0.x release ignoring any 2.0.x-pre* releases
+      which released ziti CLI to download, one of:
+        - 'stable' (default): the release GitHub marks `latest`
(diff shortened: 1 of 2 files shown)


setup-cli/action.yml
```diff
@@ -7,25 +7,38 @@ inputs:
...
            ver=$(gh release view --repo openziti/ziti | awk '/^tag/{ print $2 }')
+        elif [[ ${{ inputs.version }} == "active-lts" || ${{ inputs.version }} == "maint-lts" ]]; then
+           key=active; [[ ${{ inputs.version }} == "maint-lts" ]] && key=maintenance
+           echo "resolving ${{ inputs.version }} from lts-versions.json"
+           minor=$(gh api repos/openziti/ziti/contents/lts-versions.json -H "Accept: application/vnd.github.raw" | jq -r ".${key} // empty")
+           if [[ -z "$minor" ]]; then
+              echo "::error ::No ${key} entry in lts-versions.json"
+              exit 1
+           fi
+           echo "${{ inputs.version }} is ${minor}, fetching its latest release"
```
plorenz, line 45: Note: Drafts aren't filtered (minor but confusing failure). select(.isPrerelease==false) skips prereleases but not drafts. With a repo-write token (common in-org), gh release list includees drafts; a draft on the line gets selected, and since a draft's tag doesn't exist in git yet, the later gh release download <tag> fails cryptically. Could add and (.isDraft|not).

However, in practice we don't currently use drafts for ziti releases. So this shouldn't be a problem. Noting in case that ever changes

plorenz (approved):

PR #3975, opened 2026-06-11
Add router.link.v1 config type. Fixes #3974

plorenz:
## Summary

- Defines the built-in `router.link.v1` config type (target=router) with a JSON schema for listeners, dialers, heartbeats, payload/ack queue sizes, and a top-level `gcMode`.
- Registered for fresh databases via `initialize` and for existing databases via a migration step that bumps the schema version from 46 to 47.
- This is the first concrete router-targeted config type for the controller-managed router configuration work (see `doc/design/ctrl-managed-router-config.md`). The RDM integration that actually distributes it follows in a separate PR.

PR #3977, opened 2026-06-12
Distribute routers and router-target configs through the RDM. Fixes #3976

### LINKED ISSUE #3976: Distribute routers and router-target configs through the RDM
Add routers and router-target configs to the Router Data Model so the controller can ship per-router configuration to every router through the same event/replay infrastructure that already handles services and identities.

1. Add a `Router` message (id, name, fingerprint, configs) to the `DataState` event stream alongside Identity/Service/Config/etc.
2. Load all routers into the `RouterDataModelSender` at startup; register entity-change listeners so create/update/delete on the router store produces RDM events.
3. Per-router filtering at `RouterSender`: every router sees the full `Router` set (needed for link-peer fingerprint validation) but only the `Config` entities associated with itself.
4. Receiver side: parse and store `Router` entities. Router-target Configs that fall off the router's `Configs` list are GC'd locally so the cache stays aligned with assignment without synthetic remove events.

Depends on #3743 (config type target), #3780 (router configs field), and #3974 (`router.link.v1` config type). The fablab `router-data-model-test` is extended to exercise router-config distribution end-to-end including chaos scenarios.

Design doc: `doc/design/ctrl-managed-router-config.md` (sections "Propagation" and "Per-Router Filtering").

plorenz:
Stacked on top of #3975.

## Summary

- Adds `DataState.Router` to the RDM event stream; loads routers into the sender at startup; wires entity-change listeners on the router store.
- `RouterSender` filters `Config` events per-router: each router gets the full Router set (for link-peer fingerprint validation) but only its own router-target Configs.
- Receiver side: parses + stores Router entities; GCs router-target Configs whose IDs leave the router's `Configs` list, so the cache stays aligned without synthetic remove events.
- Extends the `router-data-model-test` fablab model with router-config distribution scenarios (assignment, reassignment, controller restart, RDM cache miss).
- Refreshes the design doc with the per-router filtering and change-notification details.

## Notes

The base PR (#3975) defines the `router.link.v1` config type; this PR makes that config actually reach a router. The follow-up PR on top will plug the receiver side into a router-side handler.

### DIFF
--- a/doc/design/ctrl-managed-router-config.md
+++ b/doc/design/ctrl-managed-router-config.md
@@ -247,20 +247,24 @@ There are two distinct triggers that cause a `Config` event to flow to a router:
 
 1. **A router's `Configs` list changes.** When an operator associates or disassociates a config:
    - Emit the `Router` event with the new `Configs` IDs.
-   - For any IDs newly added to that router's list, also emit the full `Config` entity to that
-     router (it may not have it yet).
-   - For any IDs newly removed from that router's list, do nothing extra. The receiving router
-     reads its own `Router` entity in the RDM, sees the config is no longer in its list, and stops
-     applying it. The orphaned `Config` entity stays cached on the receiver until the next full
-     sync prunes it. Skipping the explicit-remove event keeps the protocol simpler at the cost of
-     a small amount of dead weight in the receiver's cache, which we consider acceptable.
+   - Whenever that router's own `Router` event flows (on any create or update, not just the initial
+     association), also emit the full `Config` entity for *every* config currently in its `Configs`
+     list. These are sent as synthetic `Config` `Create` events; the receiver caches and dedupes, so
+     re-sending the whole list is simpler than tracking which IDs are newly added and costs nothing
+     the receiver can't absorb. (Full syncs skip this synthesis: the snapshot already carries every
+     relevant `Config` globally.)
+   - For any IDs newly removed from that router's list, do nothing extra on the wire. When the
+     receiving router processes the updated `Router` event for itself, it GCs router-target
+     `Config` entries that are no longer in its `Configs` list. Service-target configs are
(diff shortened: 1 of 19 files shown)


doc/design/ctrl-managed-router-config.md
```diff
@@ -273,11 +273,12 @@ update produces an RDM event, so routers already tolerate index gaps. We may nee
 the gap-handling logic into the `RouterSender` so it can track the previous index per router and
 set it correctly on filtered change sets, rather than relying on the receiver to reconcile.
```
andrewpmartinez: Isn't this resolved w/ this impl?

doc/design/ctrl-managed-router-config.md
andrewpmartinez: I don't recall seeing `cost` anywhere.

doc/design/ctrl-managed-router-config.md
```diff
@@ -249,11 +249,11 @@ There are two distinct triggers that cause a `Config` event to flow to a router:
    - Emit the `Router` event with the new `Configs` IDs.
    - For any IDs newly added to that router's list, also emit the full `Config` entity to that
      router (it may not have it yet).
```
andrewpmartinez: Don't we do this on all updates not just create?

andrewpmartinez (approved):
Mostly looks like the design doc needs to be brought inline w/ imple.

doc/design/ctrl-managed-router-config.md
plorenz: yes, updated doc

doc/design/ctrl-managed-router-config.md
plorenz: removed from doc. ended up not including it

doc/design/ctrl-managed-router-config.md
plorenz: correct. Updated doc

PR #3981, opened 2026-06-12
Issue 3980 posture check changes

### LINKED ISSUE #3980: process-related posture checks don't work with 2.0+
When [looking into/debugging a discourse issue](https://openziti.discourse.group/t/windows-ziti-edge-tunnel-returns-posture-state-was-nil-for-process-checks-os-check-works-fine/5881), it seems that process-related posture checks are not working properly with 2.0+

dovholuknf:
fixes #3980 

**NOTE** please review change to `internalError` wrt 'busy|retry' vs 'error'

* forwards process checks to the controller to properly clear posture checks
* adds tests to verify bad/fixed functionality (red|green tested)
* verified using ziti-edge-tunnel 1.7 sdk live in ZDEW + process posture check + os posture check
* fixes ziti list posture check issue where incomplete listing would occur

### DIFF
--- a/controller/handler_edge_ctrl/errors.go
+++ b/controller/handler_edge_ctrl/errors.go
@@ -19,8 +19,10 @@ package handler_edge_ctrl
 import (
 	"errors"
 
+	"github.com/openziti/foundation/v2/errorz"
 	"github.com/openziti/sdk-golang/ziti/edge"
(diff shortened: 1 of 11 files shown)


andrewpmartinez (changes requested):
The controller posture system is available only for legacy sessions. Making this HA aware is not the move.

dovholuknf:
cancelling - total airball on understanding that there are two systems at play not just one

Issue #3928, opened 2026-06-03
[Backport-2.0] Router posture-data updates don't revoke SDK-hosted xgress circuits or hosted terminators


plorenz:
Backport to `release-v2.0.x` of #3908 (main-line fix: PR #3922).

`onPostureDataUpdate` covered only connId mux-sink dial conns: it missed SDK-hosted xgress circuits and hosted bind terminators, and did not react at all when a service policy's posture requirements changed. The router should re-evaluate both dial circuits and hosted terminators on a posture-data change, and react to posture-requirement changes via the identity subscription.

Pre-existing, verified at `v2.0.0`. Backport once #3908 / PR #3922 merges to main.

plorenz:
Merged to release-v2.0.x

PR #3985, opened 2026-06-13
[Backport-2.0] Enforce api-session and identity revocations on the router

### LINKED ISSUE #3929: [Backport-2.0] Router does not enforce api-session or identity revocations on live connections

plorenz:
## Summary

Backport to `release-v2.0.x` of #3927 (main-line fix: PR #3930).

On the 2.0.x router, `CheckConnections` (the ~5s reaper) enforced only JWT expiry, so a revoked OIDC api-session or a disabled/deleted identity kept its live circuits and hosted terminators until the access token expired. This makes the router enforce the RouterDataModel `Revocations` directly:

- close connections whose api-session id is revoked (`API_SESSION`);
- close connections whose identity is revoked (`IDENTITY`), bounded by an `IssuedBefore` cutoff so a session re-authenticated after the cutoff survives;
- disabling/deleting an identity now produces an identity revocation in the same transaction as the identity change (self-contained OIDC JWTs aren't stored, so cascade-delete doesn't reach them).

## Backport notes

- Single-commit cherry-pick of #3930. `router/state/manager.go` (the only file shared with the #3928 posture backport) merged cleanly, so this PR is independent of #3984.
- The only conflicts were generated-proto drift (`main` is on 2.1.0). The `.proto` sources merged cleanly; the `.pb.go` files were regenerated against the 2.0.x base, each with its existing generator version (`edge_cmd.pb.go` v1.31.0, `edge_ctrl.pb.go` v1.36.11), so the generated diff is limited to the new `type`/`issuedBefore` fields.
- `go build ./common/... ./controller/... ./router/...` and `go vet ./tests/` pass on the 2.0.x base.

Fixes #3929.


controller/model/identity_manager.go
```diff
@@ -81,6 +82,16 @@ func NewIdentityManager(env Env) *IdentityManager {
...
 	RegisterCommand(env, &UpdateServiceConfigsCmd{}, &edge_cmd_pb.UpdateServiceConfigsCmd{})
 
+	// Revoke a deleted or disabled identity's live OIDC sessions atomically with
+	// the identity change itself, so the revocation cannot be skipped. The db-level
+	// constraint is config-agnostic, so the revocation type and lifetime are
+	// supplied here.
+	env.GetStores().Identity.AddEntityConstraint(db.NewIdentityRevocationConstraint(
+		env.GetStores().Revocation,
+		common.RevocationTypeIdentity,
+		func() time.Duration { return env.GetConfig().Edge.Oidc.RefreshTokenDuration },
```
andrewpmartinez: This should be `MaxTokenDuration()` which is the greatest value from all token durations. The refresh token should be the biggest, but we need to be sure.

andrewpmartinez (approved):
One small bug for consistency and security.

controller/model/identity_manager.go
plorenz: Fixed and ported the change back to main in another PR

Issue #3988, opened 2026-06-15
Support multiple LAN interfaces for tproxy mode(resolver)


emoscardini:
The resolver option in xgress_edge_tunnel tproxy binding currently accepts only a single udp://ip:port
  address for the internal DNS server that maps intercepted service hostnames to virtual IPs. In
  deployments with multiple LAN-facing interfaces on different subnets, DNS clients on all but one subnet
  cannot reach the resolver without binding it to 0.0.0.0 — which is not optimal in a security product.

  Proposed Change

  Allow resolver to accept either a single string (backwards-compatible) or a YAML list of udp://
  addresses. Internally, a single shared resolver instance handles all listeners so hostname→IP mappings
  are consistent regardless of which address a DNS query arrives on.

  Single address (unchanged behavior):
  - binding: tunnel
    options:
      mode: tproxy
      resolver: udp://172.18.102.70:53

  Multiple addresses (new):
  - binding: tunnel
    options:
      mode: tproxy
      resolver:
        - udp://172.18.102.70:53
        - udp://192.168.10.1:53

  Notes

  - All entries must use the udp:// scheme. Mixing schemes or using multiple file:// entries is an error.
  - The file:// resolver scheme continues to work as before and only accepts a single entry.
  - This change is complementary to issue #3972 (multi-interface lanIf support) and is intended to  support the same multi-LAN-interface deployment scenario.

emoscardini:
https://github.com/openziti/ziti/pull/3989

PR #3989, opened 2026-06-15
Feature/resolver multi interface

emoscardini:
https://github.com/openziti/ziti/issues/3988 

  resolver previously accepted only a single udp:// address. This change makes it accept either a bare
  string (backwards-compatible) or a YAML list of addresses. A single shared resolver instance handles
  all listeners so hostname mappings remain consistent across interfaces.

### DIFF
--- a/tunnel/dns/dns_linux.go
+++ b/tunnel/dns/dns_linux.go
@@ -70,33 +70,33 @@ func NewDnsServer(addrs []string, upstreams []string, unanswered unansweredDispo
 		log.Infof("configured upstream DNS server: %s over %s", upstreamURL.Host, upstreamURL.Scheme)
 	}
 
+	startedCh := make(chan struct{}, len(addrs))
 	errChan := make(chan error, len(addrs))
+
 	for _, addr := range addrs {
 		s := &dns.Server{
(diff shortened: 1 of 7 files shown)


tunnel/dns/dns_linux.go
```diff
@@ -74,22 +69,35 @@ func NewDnsServer(addr string, upstreams []string, unanswered unansweredDisposit
...
+			errChan <- s.ListenAndServe()
+		}()
+	}
 
-	select {
-	case err := <-errChan:
-		if err != nil {
-			return nil, fmt.Errorf("dns server failed to start: %w", err)
-		} else {
+	deadline := time.After(2 * time.Second)
```
scareything: Why is this a timeout? Is there a real concern that any of the LitenAndServe calls won't complete? Or is the timeout just allowing enough time for all of them to complete? If so this is probably better done with a waitgroup.

If a timeout is used, it seems better to error out if any of the addresses aren't serving.

emoscardini: I'll push an update to error out.

scareything (approved):

PR #3991, opened 2026-06-15
Harden and clarify firewalled controller cluster join

### LINKED ISSUE #3841: Controller Cluster - new controllers must be able to be dialed by the leader in order to join successfully
Desired Setup:
 * Two node cluster, where one node is public and the second node is behind a network firewall with zero inbound ports open

To reproduce:
  * Configure the public node and initialize a ha cluster
  * Configure the private node and attempt to join the cluster by sending an add command to the public node
 
Observed behavior:
* The private node establishes a raft connection with the leader and sends an add member command
* The leader attempts to dial the private node which times out after one minute
* The leader drops the new node from the cluster and terminates the raft connection

plorenz:
## What

Closes #3841.

A controller behind a firewall with no inbound ports could fail to join an HA cluster: the leader would try to dial the joining node's advertise address, time out after ~1 minute, and drop the new member.

The core functional fix for this (the mesh reusing the joining node's already-established inbound connection instead of dialing out) already landed on `main` in #3684, but `#3841` was still open and unverified. This PR closes it out by:

- **Regression tests** in the mesh package asserting that `GetPeerInfo`, `WaitForPeer`, and `GetOrConnectPeer` reuse an existing inbound peer connection rather than dialing the peer's (possibly unreachable) address. These guard the three points in the add-member / replication / forwarding paths where the leader could otherwise dial out.
- **A clearer operator error** on the leader-side add path. Previously, adding an unreachable node from the leader produced an opaque dial timeout. It now explains that a node with no inbound ports must initiate the join itself.

## Note

Adding from the leader fundamentally cannot reach a zero-inbound node; the supported workflow for a firewalled controller is to initiate the join from the joining node so the leader can reuse the inbound connection. The new error message points operators toward that.

### DIFF
--- a/controller/raft/member.go
+++ b/controller/raft/member.go
@@ -121,7 +121,14 @@ func (self *Controller) HandleAddPeerAsLeader(req *cmd_pb.AddPeerRequest) error 
 
 	peerId, peerAddr, err := self.Mesh.GetPeerInfo(req.Addr, 15*time.Second)
 	if err != nil {
-		return err
+		// GetPeerInfo reuses an already-established connection if one exists, so reaching
+		// this point means the joining node is neither connected nor dialable from here.
+		// A node with no inbound ports open (e.g. behind a firewall) cannot be dialed by
+		// the leader, and must instead initiate the join itself so the leader can reuse the
+		// inbound connection. Surface that guidance rather than an opaque dial timeout.
+		return errors.Wrapf(err, "unable to reach peer at '%s' to add it to the cluster; "+
+			"if this node has no inbound ports open (e.g. behind a firewall), run the join from the "+
+			"joining node instead ('ziti agent cluster add -i <joining-node> <joining-node-advertise-addr>')", req.Addr)
 	}
 
 	r := self.GetRaft()
--- a/controller/raft/mesh/mesh_test.go
+++ b/controller/raft/mesh/mesh_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hashicorp/raft"
 	"github.com/openziti/foundation/v2/versions"
 	"github.com/openziti/ziti/v2/controller/event"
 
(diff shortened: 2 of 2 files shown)


controller/raft/member.go
```diff
@@ -121,7 +121,14 @@ func (self *Controller) HandleAddPeerAsLeader(req *cmd_pb.AddPeerRequest) error
...
 	if err != nil {
-		return err
+		// GetPeerInfo reuses an already-established connection if one exists, so reaching
+		// this point means the joining node is neither connected nor dialable from here.
+		// A node with no inbound ports open (e.g. behind a firewall) cannot be dialed by
+		// the leader, and must instead initiate the join itself so the leader can reuse the
+		// inbound connection. Surface that guidance rather than an opaque dial timeout.
+		return errors.Wrapf(err, "unable to reach peer at '%s' to add it to the cluster; "+
+			"if this node has no inbound ports open (e.g. behind a firewall), run the join from the "+
+			"joining node instead ('ziti agent cluster add <existing-member-addr>' on the joining node)", req.Addr)
```
andrewpmartinez: Not: `ziti agent cluster add -i <joining-node> <joining-node-advertise-addr>` ?

controller/raft/mesh/mesh_test.go
```diff
@@ -121,6 +122,72 @@ func Test_RemovePeer_RemovesReadonlyWhenDeletingLastPeer(t *testing.T) {
 	assert.Equal(t, false, m.readonly.Load(), "Expected readonly to be false, got ", m.readonly.Load())
 }
 
+// firewalledPeerAddr is a well-formed transport address that is not reachable.
+// The reuse tests below register a peer at this address, then assert the mesh
+// returns the existing connection without attempting to dial it.
+const firewalledPeerAddr = "tls:firewalled-node.example:6262"
+
+// Test_GetPeerInfo_ReusesExistingConnectionWithoutDialing is a regression test
+// for #3841. When the leader adds a member that is behind a firewall (no inbound
```
andrewpmartinez: I don't like referencing GitHub commits in godoc/comments. Best to just describe what we are doing and why rather than assuming the external reference will always exist.

controller/raft/mesh/mesh_test.go
```diff
@@ -121,6 +122,72 @@ func Test_RemovePeer_RemovesReadonlyWhenDeletingLastPeer(t *testing.T) {
...
+	elapsed := time.Since(start)
+
+	assert.NoError(t, err)
+	assert.Equal(t, raft.ServerID("joining-node-id"), id)
+	assert.Equal(t, raft.ServerAddress(firewalledPeerAddr), addr)
+	assert.Less(t, elapsed, 200*time.Millisecond, "should reuse existing connection, not dial out")
+}
+
+// Test_WaitForPeer_ReturnsExistingPeerWithoutDialing is a regression test for
+// #3841. The raft StreamLayer Dial path (used when the leader replicates to a new
```
andrewpmartinez: Same. Github issue reference.

andrewpmartinez (approved):

controller/raft/mesh/mesh_test.go
plorenz: fixed

controller/raft/mesh/mesh_test.go
plorenz: fixed

controller/raft/member.go
plorenz: You are correct. Fixed

Issue #3974, opened 2026-06-11
Add router.link.v1 config type


plorenz:
Define a built-in `router.link.v1` config type that describes a router's link subsystem.

1. Target is `router` (per #3743).
2. Schema covers listeners, dialers, heartbeats, payload/ack sender queue sizes, and a top-level `gcMode` for auto-GC of stale links.
3. Registered as a built-in via the standard config-type migration; new and existing databases pick it up automatically.

This is the first concrete config type for the controller-managed router configuration work; the design lives in `doc/design/ctrl-managed-router-config.md`.

PR #3999, opened 2026-06-17
Add router-side managed configuration registry. Fixes #3998

### LINKED ISSUE #3998: Add router-side managed configuration registry
Part of #3743. Following #3974 (the `router.link.v1` config type) and #3976 (distributing router configs through the RDM), the router needs a framework to receive controller-managed config and make it available to subsystems. This covers the framework only -- wiring an actual subsystem (link management) to consume it is a follow-up.

Scope:
- A managed-config registry tracking available config per base type by source (controller vs. local) and version.
- Reconciliation to an effective config with local-config-wins precedence at the base-type level, selecting the highest supported version.
- A local config-type allow-list so an operator bounds which types the controller may enable (the security boundary stays on the router).
- Loading the `managedConfig` section from the router config file.
- Wiring the RDM subscriber to feed router-target config events into the registry.
- Inspect support for the registry.

Out of scope (follow-ups): consuming the registry for link management (routing `router.link.v1` through it), per-type YAML -> JSON translation, and stale-link detection/GC.

Design: `doc/design/ctrl-managed-router-config.md`.

plorenz:
Adds the router-side framework for controller-managed configuration: a registry that receives managed config (via the RDM) and makes an effective config available to subsystems, with source tracking, version selection, and local-config-wins precedence.

This is framework only. The registry is wired up and fed by the RDM subscriber, but no subsystem consumes it yet; routing link management (`router.link.v1`) through it is the next PR.

Highlights:
- Managed-config registry tracking available config per base type by source (controller/local) and version, reconciled to an effective config with local-wins precedence at the base-type level.
- Local config-type allow-list so the operator bounds which types the controller may enable; the security boundary stays on the router.
- Loads the `managedConfig` section from the router config file.
- RDM subscriber feeds router-target config events into the registry.
- Inspect support for the registry.

Part of #3743. Builds on #3974 and #3976. Design: `doc/design/ctrl-managed-router-config.md`.

### DIFF
--- a/router/managedconfig/registry.go
+++ b/router/managedconfig/registry.go
@@ -1,0 +1,623 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package managedconfig
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/michaelquigley/pfxlog"
+	"github.com/openziti/foundation/v2/goroutines"
+	"github.com/openziti/ziti/v2/common/inspect"
+)
+
+const (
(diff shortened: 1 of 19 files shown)


router/managedconfig/registry.go
```diff
@@ -0,0 +1,574 @@
...
+
+func (self *Registry) spawnReconcile(entry *handlerEntry) {
+	self.wgMu.Lock()
+	if self.closed.Load() {
+		self.wgMu.Unlock()
+		return
+	}
+	self.wg.Add(1)
+	self.wgMu.Unlock()
+	go self.reconcileAsync(entry)
```
andrewpmartinez: No bounds on this?

andrewpmartinez (approved):
No integration tests?

router/managedconfig/registry.go
plorenz: Added a coalesce mechanism, so only one reconcile and one queued reconcile will be active at any time.

plorenz:
> No integration tests?

Integration tests come in the next PR when there's a user of the registry that we can test.

PR #4006, opened 2026-06-18
  fixes #3992 stop controller panic when a web server listener fails

### LINKED ISSUE #3992: Overlay edge-oidc listener panics when its redirect_uri is not in the redirectURIs allow-list
**Experienced**

The controller emits a nil pointer dereference panic when an overlay-exposed edge-oidc listener authenticates and its redirect_uri is not in the edge-oidc redirectURIs allow-list. The OIDC request returns a 400 invalid_request, and the controller then panics instead of reporting the misconfiguration.

**Expected**

When the redirect_uri is not in the allow-list, the controller should log a clear configuration error, not panic.

**Reproduction**

1. Configure an edge-oidc API on a web server exposed over the overlay.
2. Set redirectURIs that omit the callback the OIDC client uses (default `http://localhost:8080/auth/callback`):
   ```yaml
   - binding: edge-oidc
     options:
       redirectURIs:
         - "https://ctrl.example.com:*/auth/callback"
         - "https://node.example.com:*/auth/callback"
   ```
3. Start the controller.
4. The overlay listener authenticates, and the OIDC request returns a 400 because the redirect_uri is not in the allow-list.
5. The controller panics with a nil pointer dereference.

Workaround: add the requested callback to redirectURIs, for example `http://localhost:8080/auth/callback`.

**Notes**

Observed with xweb v3.0.4 and sdk-golang v1.7.0.

Error logged just before the panic:
```
error listening on overlay: failed to listen: no apiSession, authentication attempt failed:
... the expected redirect URL (http://localhost:8080/auth/callback) is not allowed:
{"error":"invalid_request","error_description":"The requested redirect_uri is missing in the
client configuration."}
```

Panic stack (trimmed):
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28]
net/http.(*Server).Serve(..., {0x0, 0x0})
github.com/openziti/xweb/v3.(*Server).Start.func1()
        github.com/openziti/xweb/v3@v3.0.4/server.go:196
```

andrewpmartinez:
- updates openziti/xweb/v3 for actual fix
  - adds Test_OidcListener_BindFailure_DoesNotPanic and the oidc-listener-bind-failure config set, asserting the controller keeps the other web servers running when an edge-oidc listener cannot bind

Requires: https://github.com/openziti/xweb/pull/35

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -75,7 +75,7 @@ require (
 	github.com/openziti/secretstream v0.1.51
 	github.com/openziti/transport/v2 v2.0.216
 	github.com/openziti/x509-claims v1.0.3
-	github.com/openziti/xweb/v3 v3.0.4
+	github.com/openziti/xweb/v3 v3.0.5-0.20260618140905-54cc19903f1c
 	github.com/orcaman/concurrent-map/v2 v2.0.1
 	github.com/pkg/errors v0.9.1
 	github.com/rabbitmq/amqp091-go v1.11.0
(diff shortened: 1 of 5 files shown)


plorenz (approved):
Assuming you'll update the deb before merging

PR #4012, opened 2026-06-19
Accept SDK inspect replies regardless of content type

### LINKED ISSUE #4011: Router deletes terminators ~12m after creation when host SDK replies with wrong inspect content type
## Summary

Routers (v2.0.0, v2.0.0-pre10) delete healthy terminators roughly 12 minutes after they are created when the hosting application is built with an affected sdk-golang version. The terminators are fully functional; they are removed because the router's post-create inspect never receives a usable response.

## Root cause

After a terminator is established, the router runs a `post-create inspect` cycle: it sends a `ConnInspectRequest` (content type 60798) to the SDK and expects a `ConnInspectResponse` (60799). It registers a typed receive handler for 60799 and correlates the reply by sequence.

sdk-golang **v1.1.0 through v1.5.3** reply to that request with content type **InspectResponse (60805)** instead of **ConnInspectResponse (60799)**. This is a constant-aliasing regression in the SDK's `ziti/edge/messages.go` (`ContentTypeConnInspectResponse` pointed at the wrong protobuf enum); it was fixed in **v1.5.4**.

Because the router fire-and-forgets the request (`TrySend`) and has no typed handler for 60805, the channel drops the reply (`dropped message. type [60805]`). The post-create inspect never completes, and after the send-count/timeout threshold the router deletes the terminator (`post-create inspect: timed out waiting for response, closing terminator`).

The same content-type check in `edge.UnmarshalInspectResult` also makes controller-driven terminator validation unable to read status from these hosts (it falls back to assuming the terminator is ok).

## Affected SDK versions

sdk-golang v1.1.0 - v1.5.3 (acting as a terminator host). Fixed in sdk-golang >= v1.5.4. Apps in the wild built on the affected range (e.g. anything embedding v1.2.x) trigger this against a v2.0.0 router.

## Fix

Router-side mitigation that keeps the existing validation behavior intact:
- Correlate the post-create inspect reply by reply sequence (via the channel waiter map) rather than by content type, so the reply is accepted regardless of whether it is stamped 60799 or 60805.
- Parse the reply leniently (read headers/body without the content-type guard) on the post-create and validate-terminators paths.

The fix is self-contained and removable once the affected SDK versions are out of support. It does not loosen the timeout-delete validation, which exists to catch genuinely invalid terminators.

Reported on the community forum: https://openziti.discourse.group/t/router-v2-0-0-pre10-and-v2-0-0-stable-deletes-all-terminators-after-12-minutes/5901

plorenz:
## What

Fixes #4011.

Routers were deleting healthy terminators ~12 minutes after creation when the hosting application used an affected sdk-golang version. The router's post-create inspect sends a `ConnInspectRequest` and waits for a `ConnInspectResponse` (content type 60799), but sdk-golang v1.1.0 - v1.5.3 reply with the wrong content type (`InspectResponse`, 60805). The router had no handler for that type, so the reply was dropped, the inspect timed out, and the terminator was removed.

## Changes

- Correlates the post-create inspect reply by reply sequence (via the channel waiter map) instead of by message content type, so the reply is accepted whether stamped 60799 or 60805.
- Adds a lenient inspect-result parser (reads headers/body without the content-type guard) and uses it on the post-create and controller-driven terminator-validation paths.

The timeout-delete validation behavior is unchanged; this only ensures a valid reply actually reaches the validator. The workaround is self-contained and tagged for removal once sdk-golang < v1.5.4 is out of support.

## Backport

Marked for backport once merged to main.

### DIFF
--- a/router/xgress_edge/hosted.go
+++ b/router/xgress_edge/hosted.go
@@ -591,6 +591,56 @@ func (self *inspectResponseEvent) handle(registry *hostedServiceRegistry) {
 	}
 }
 
+// inspectRequestSendable wraps a ConnInspectRequest so its reply is correlated by reply
+// sequence (via the channel waiter map) instead of by message content type.
+//
+// Workaround for sdk-golang v1.1.0 - v1.5.3: those versions reply to a ConnInspectRequest
+// (60798) with content type InspectResponse (60805) instead of ConnInspectResponse (60799).
+// Such a reply has no typed receive handler on the router, so it is dropped, the post-create
+// inspect never completes, and the terminator is deleted as unresponsive. Registering a reply
+// receiver keyed on the request sequence lets us accept the reply regardless of content type.
+//
+// To remove once sdk-golang < v1.5.4 is out of support: send a plain *channel.Message again
+// and let the ContentTypeConnInspectResponse handler registered in accept.go process the reply.
+type inspectRequestSendable struct {
+	*channel.Message
+	registry *hostedServiceRegistry
+	conn     *edgeClientConn
+}
+
+var _ channel.Sendable = (*inspectRequestSendable)(nil)
+var _ channel.ReplyReceiver = (*inspectRequestSendable)(nil)
+
+// ReplyReceiver returns the sendable itself so the channel delivers the reply to AcceptReply.
+func (self *inspectRequestSendable) ReplyReceiver() channel.ReplyReceiver {
+	return self
+}
+
+// AcceptReply is invoked on the channel rx goroutine when the inspect reply arrives. It parses
+// the result leniently (ignoring content type) and hands off to the registry event loop.
+func (self *inspectRequestSendable) AcceptReply(reply *channel.Message) {
+	self.registry.queue(&inspectResponseEvent{
+		conn:     self.conn,
+		replyFor: reply.ReplyFor(),
+		result:   unmarshalInspectResultLenient(reply),
+	})
+}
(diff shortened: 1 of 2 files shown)


router/xgress_edge/hosted.go
```diff
@@ -591,6 +591,53 @@ func (self *inspectResponseEvent) handle(registry *hostedServiceRegistry) {
...
+//
+// Workaround for sdk-golang v1.1.0 - v1.5.3: those versions reply to a ConnInspectRequest
+// (60798) with content type InspectResponse (60805) instead of ConnInspectResponse (60799).
+// Such a reply has no typed receive handler on the router, so it is dropped, the post-create
+// inspect never completes, and the terminator is deleted as unresponsive. Registering a reply
+// receiver keyed on the request sequence lets us accept the reply regardless of content type.
+//
+// To remove once sdk-golang < v1.5.4 is out of support: send a plain *channel.Message again
+// and let the ContentTypeConnInspectResponse handler registered in accept.go process the reply.
+type inspectRequestSendable struct {
```
andrewpmartinez, line 605: Consider interface type guard:

`var _ channel.ReplyReceiver = (*inspectRequestSendable)(nil)`

andrewpmartinez (approved):

router/xgress_edge/hosted.go
plorenz: done

PR #4013, opened 2026-06-19
Prevent enrollment-based privilege escalation to admin identities. Fixes #4010

### LINKED ISSUE #4010: Non-admin identity with `enrollment` entity permission can create enrollments for any identity including admins, achieving full privilege escalation to admin
## Summary

OpenZiti v2.0.0 introduced a new entity-level permissions model that allows an admin to grant fine-grained permissions (such as `enrollment`, `identity`, `service`, etc.) to non-admin identities. A non-admin identity granted the `enrollment` permission can reach the `POST /edge/management/v1/enrollments` endpoint via the `DefaultManagementAccess` gate, which does not verify that the caller is authorized for the target identity. An attacker with only `enrollment` entity permission can create an OTT (one-time-token) enrollment for the default admin identity, then use the unauthenticated enrollment endpoint (`POST /edge/client/v1/enroll/ott`) to enroll an attacker-controlled certificate for the admin identity. The resulting certificate authenticates as the admin identity (isAdmin=true, isDefaultAdmin=true) with full management API access.

### Root cause

`enrollment_router.go` registers the `Create` handler with entity type `"enrollment"` and `DefaultManagementAccess()`:

```go
// controller/internal/routes/enrollment_router.go, line 70-75
ae.POST("", func(ae *env.AppEnv, rc *response.RequestContext) {
    rc.SetEntityType("enrollment")
    rc.SetEntityAction(permissions.Create)
    permissions.DefaultManagementAccess().IsAllowed(rc) // gate
    Create(ae, rc)
})
```

`DefaultManagementAccess()` allows any authenticated identity that holds the `enrollment` entity permission:

```go
// controller/permissions/management.go
func DefaultManagementAccess() RequestChecker {
    return NewRequireAllOf(
        IsManagementApi(),
        HasOneOf(IsAdmin(), HasAdminReadOnlyAccess(), HasEntityAccess(), HasEntityActionAccess()),
    )
}
// HasEntityAccess passes when ctx.HasPermission("enrollment") == true
```

The `Create` handler calls `enrollment_manager.ApplyCreate` with the caller-supplied `identityId`:

```go
// controller/internal/routes/enrollment_router.go, line 115-120
func Create(ae *env.AppEnv, rc *response.RequestContext) {
    MapCreate(ae.Managers.Enrollment.Create,
        MapCreateEnrollmentToModel(params.Enrollment), rc)
}

// MapCreateEnrollmentToModel
func MapCreateEnrollmentToModel(create *rest_model.EnrollmentCreate) *model.Enrollment {
    return &model.Enrollment{
        IdentityId: create.IdentityID,  // attacker-supplied
        Method:     string(*create.Method),
        ...
    }
}
```

`ApplyCreate` in `enrollment_manager.go` validates only that the identity exists and the enrollment method is valid -- it never checks that the session caller is authorized to create enrollments on behalf of the target identity:

```go
// controller/model/enrollment_manager.go, ApplyCreate()
// Checks: identity exists, method valid, CA exists (for OTTCA), username not duplicate (for UPDB)
// MISSING: check that cmd.Enrollment.IdentityId == caller's identity, or caller is admin
```

Once the attacker holds an OTT token for the admin identity, they call the publicly-accessible enrollment endpoint:

```go
// POST /edge/client/v1/enroll/ott  (permissions.Always() -- no auth required)
// enrollment_mod_ott.go Process():
//   1. reads enrollment record by token
//   2. finds target identity from record (admin's identity)
//   3. signs attacker's CSR as that identity
//   4. creates new cert authenticator bound to admin's identityId
```

No ownership check exists at any point in this chain.

### Attack conditions

1. The attacker must have an authenticated session on the Management API (`POST /edge/management/v1/authenticate`).
2. An admin must have granted the attacker's identity the `enrollment` entity permission (or the broader `enrollment.create` action permission). This is a legitimate administrative delegation scenario intended for provisioning automation.

### Impact

A non-admin identity with `enrollment` entity permission achieves complete privilege escalation to the default admin (or any other admin) identity. The attacker gains:
- Persistent certificate-based authentication as the admin identity
- Full Management API access (create/delete/modify all identities, services, service-policies, edge-router-policies, auth-policies, CAs, etc.)
- Ability to create additional admin identities or escalate any other non-admin identity

## Proof of Concept

(available upon request)

plorenz:
## What

Closes a privilege-escalation hole in the entity-level permission system. A non-admin granted the `enrollment` (or `enrollment.create`) permission could create a one-time-token enrollment for an admin identity and then enroll as that admin, gaining full admin access. Because an enrollment also carries the live token/JWT, reading or refreshing an admin identity's enrollment is an equivalent leak.

## Why

The `identity` and `authenticator` routers already prevent non-admins from acting on admin identities, but the enrollment router had no such guard. This brings enrollment in line with those existing protections.

## Changes

- Non-admins can no longer create, read, refresh, or delete an enrollment that belongs to an admin identity.
- Admin identity enrollments are filtered out of enrollment lists for non-admin callers.
- Full admins are unaffected; the guards only trigger on admin-target enrollments, so existing non-admin enrollment flows against non-admin identities continue to work.
- Adds a permissions test covering each escalation path plus the admin-allowed behavior.

### DIFF
--- a/controller/internal/routes/enrollment_router.go
+++ b/controller/internal/routes/enrollment_router.go
@@ -17,16 +17,20 @@
 package routes
 
 import (
+	"errors"
 	"time"
 
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/openziti/edge-api/rest_management_api_server/operations/enrollment"
 	"github.com/openziti/edge-api/rest_model"
 	"github.com/openziti/foundation/v2/errorz"
+	"github.com/openziti/foundation/v2/stringz"
 	"github.com/openziti/ziti/v2/controller/env"
 	"github.com/openziti/ziti/v2/controller/model"
+	"github.com/openziti/ziti/v2/controller/models"
 	"github.com/openziti/ziti/v2/controller/permissions"
 	"github.com/openziti/ziti/v2/controller/response"
+	"github.com/openziti/ziti/v2/controller/storage/ast"
 )
 
 func init() {
@@ -76,18 +80,40 @@ func (r *EnrollmentRouter) Register(ae *env.AppEnv) {
 }
 
 func (r *EnrollmentRouter) List(ae *env.AppEnv, rc *response.RequestContext) {
-	ListWithHandler[*model.Enrollment](ae, rc, ae.Managers.Enrollment, MapEnrollmentToRestEntity)
+	lister := ae.Managers.Enrollment
+	ListWithQueryF[*model.Enrollment](ae, rc, lister, MapEnrollmentToRestEntity, func(query ast.Query) (*models.EntityListResult[*model.Enrollment], error) {
+		// Enrollments expose the one-time-token used to enroll as their target identity, so a
+		// non-admin must not be able to see enrollments belonging to admin identities.
+		if !rc.HasPermission(permissions.AdminPermission) {
+			scopeQuery, err := ast.Parse(lister.GetSymbolTypes(), "not (identity.isAdmin = true)")
+			if err != nil {
+				return nil, err
+			}
+			query.SetPredicate(ast.NewAndExprNode(query.GetPredicate(), scopeQuery.GetPredicate()))
+		}
+		return lister.BasePreparedList(query)
+	})
 }
 
 func (r *EnrollmentRouter) Detail(ae *env.AppEnv, rc *response.RequestContext) {
+	if !r.allowNonAdminAccessToEnrollment(ae, rc) {
+		return
+	}
 	DetailWithHandler[*model.Enrollment](ae, rc, ae.Managers.Enrollment, MapEnrollmentToRestEntity)
 }
 
 func (r *EnrollmentRouter) Delete(ae *env.AppEnv, rc *response.RequestContext) {
+	if !r.allowNonAdminAccessToEnrollment(ae, rc) {
+		return
+	}
 	DeleteWithHandler(rc, ae.Managers.Enrollment)
 }
 
 func (r *EnrollmentRouter) Refresh(ae *env.AppEnv, rc *response.RequestContext, params enrollment.RefreshEnrollmentParams) {
+	if !r.allowNonAdminAccessToEnrollment(ae, rc) {
+		return
+	}
+
 	id, err := rc.GetEntityId()
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


controller/internal/routes/enrollment_router.go
```diff
@@ -113,12 +139,60 @@ func (r *EnrollmentRouter) Refresh(ae *env.AppEnv, rc *response.RequestContext,
...
+func (r *EnrollmentRouter) allowNonAdminEnrollmentForIdentity(ae *env.AppEnv, rc *response.RequestContext, identityId string) bool {
+	if rc.HasPermission(permissions.AdminPermission) {
+		return true
+	}
+
+	if identityId != "" {
+		if identity, _ := ae.Managers.Identity.Read(identityId); identity != nil && identity.IsAdmin {
+			rc.RespondWithError(nonAdminNotAllowedError(errors.New("non-admins may not manage enrollments for admin identities")))
+			return false
+		}
```
andrewpmartinez: If an error occurs, we return `true`. If we error we should bail and say "no"  by returning `false`.

andrewpmartinez (changes requested):
Fix truth error return.

controller/internal/routes/enrollment_router.go
plorenz: fixed

controller/internal/routes/enrollment_router.go
```diff
@@ -113,12 +139,65 @@ func (r *EnrollmentRouter) Refresh(ae *env.AppEnv, rc *response.RequestContext,
...
+
+	return true
+}
+
+// allowNonAdminAccessToEnrollment ensures that a non-admin caller is not reading, refreshing, or
+// deleting an enrollment that belongs to an admin identity. The enrollment carries the live token
+// and JWT used to enroll as its target identity, so exposing or refreshing an admin identity's
+// enrollment would permit privilege escalation. It responds with an unauthorized error and returns
+// false when the operation should be denied. When the enrollment cannot be loaded, it allows the
+// request to proceed so the downstream handler can produce the appropriate not-found response.
```
andrewpmartinez: Not the biggest fan of this type of security handling. I'd rather helpers return a direct, concrete, and verifiable state. In security contexts, "letting the handler downstream do the right thing implicitly" is just a recipe for future code changes causing security lapses.

andrewpmartinez (approved):
Its..."fine".

controller/internal/routes/enrollment_router.go
plorenz: expanded handling and changed to continuation passing style

PR #4027, opened 2026-06-23
Update deps and changelog for 1.5.16 release

plorenz:


zititest/go.mod
```diff
@@ -27,8 +27,8 @@ require (
 	github.com/sirupsen/logrus v1.9.4
 	github.com/spf13/cobra v1.10.2
 	github.com/stretchr/testify v1.11.1
-	go.etcd.io/bbolt v1.4.3
```
andrewpmartinez, line 30: Confirm this is backwards compat w/ the db files.

andrewpmartinez (approved):

zititest/go.mod
plorenz: rolled back to 1.4.3, not b/c I'm worried about incompatibility, but because it was only released 3 days ago.

PR #4043, opened 2026-06-26
Add ziti ops log-pipe command

plorenz:
Adds a small log-capture command so a component's stdout/stderr can be written to a file with a chosen retention strategy, rather than relying on external tooling.

- adds `ziti ops log-pipe`, which reads stdin and writes it to a log file
- registers the command under both the ops and tools command groups

Carved out of the gossip link-state work as an independent base change.

### DIFF
--- a/tests/testdata/configs/oidc-listener-bind-failure/ctrl.yml
+++ b/tests/testdata/configs/oidc-listener-bind-failure/ctrl.yml
@@ -3,12 +3,13 @@ v: 3
 db: ${ZITI_TEST_DB}
 
 identity:
-  cert: testdata/ca/intermediate/certs/ctrl-client.cert.pem
-  server_cert: testdata/ca/intermediate/certs/ctrl-server.cert.pem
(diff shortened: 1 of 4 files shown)


dovholuknf (approved):

andrewpmartinez:
Where is the:

> supports truncate, append-across-restarts, and rotate-with-retention strategies

?

andrewpmartinez:
Also w/o a related issue, this won't show up in the changelog automatically.

plorenz:
> Also w/o a related issue, this won't show up in the changelog automatically.

Since this is intended for use in fablab tests, I think excluding it from the changelog makes sense. I'll check and see if the 'hidden' flag is set, and if not, I'll set it

plorenz:
> Where is the:
> 
> > supports truncate, append-across-restarts, and rotate-with-retention strategies
> 
> ?

Fixed. These are supported in fablab, not all by log-pipe directly. Truncate is supported by not using the new log-pipe tool. If you use log-pipe, you get append-across-restarts + rotate-with-retention.

PR #4044, opened 2026-06-26
Add fablab log strategy and host-reachability-aware chaos helpers

plorenz:
Fablab test-harness improvements used for long-running soak tests.

- adds a configurable LogStrategy (truncate/append/rotate) to zitilab components; rotate pipes component output through `ziti ops log-pipe`
- adds host-reachability checks and a convergence deadline that excludes time a host is unreachable, so a validation loop can distinguish a dead process from unreachable infrastructure
- extends the chaos toolkit used to disrupt components during runs

Carved out of the gossip link-state work as an independent base change. The rotate strategy pairs with the `ziti ops log-pipe` PR.

### DIFF
--- a/zititest/zitilab/chaos/chaos.go
+++ b/zititest/zitilab/chaos/chaos.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/michaelquigley/pfxlog"
+	"github.com/openziti/fablab/kernel/lib/parallel"
 	"github.com/openziti/fablab/kernel/model"
 	"github.com/openziti/ziti/v2/zitirest"
 )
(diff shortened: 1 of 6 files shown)


zititest/zitilab/chaos/chaos.go
```diff
@@ -148,6 +159,106 @@ func RestartSelected(run model.Run, concurrency int, list ...*model.Component) e
...
+			for {
+				isRunning, err := c.IsRunning(run)
+				if err != nil {
+					return err
+				}
+				if !isRunning {
+					break
+				}
+				time.Sleep(250 * time.Millisecond)
+			}
```
andrewpmartinez, line 250: I assume the lack of a deadline here is expected.

andrewpmartinez (approved):

zititest/zitilab/chaos/chaos.go
plorenz: Added a deadline

PR #4048, opened 2026-06-26
Add VPC dependency-aware teardown to the AWS housekeeper

plorenz:
Add VPC dependency-aware teardown to the AWS housekeeper

Makes it easier to clean up orphaned AWS resources left behind by smoke tests.

- adds delete_vpc_with_deps, which tears down a VPC's dependencies (instances, leftover network interfaces, internet gateways, custom route tables, and security groups) before deleting the VPC itself
- treats mutating calls as best-effort, logging a warning and continuing so one stuck resource does not abort the whole cleanup run

### DIFF
--- a/zititest/scripts/housekeeper-aws.bash
+++ b/zititest/scripts/housekeeper-aws.bash
@@ -235,13 +235,151 @@ function terminate_instances(){
   done
 }
 
+# delete_vpc_with_deps deletes a single VPC after removing the dependencies that
+# would otherwise block delete-vpc. The order mirrors how fablab provisions a VPC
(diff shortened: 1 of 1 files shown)


andrewpmartinez (approved):
There are a number of other types that could also cause this to fail, but I assume since we don't use them in Fablab VPCs we don't care right now.

plorenz:
> There are a number of other types that could also cause this to fail, but I assume since we don't use them in Fablab VPCs we don't care right now.

Yes, if we start adding additional resource types to fablab setup, we'll need to expand this. For this pass, I was trying to make it a single step operation. Before it was scan, generate files and then do a second pass pointing at those files.

Issue #3990, opened 2026-06-15
Expose service change subscriptions to external SDKs over protobuf


andrewpmartinez:
Add edge router protobuf messages that let external SDKs subscribe to service change events because SDKs currently have no push-based way to learn about these changes and must poll or reconnect to detect them.

The Edge Router Tunneler (ERT) embedded in routers already consumes an internal subscription model for these events. This feature exposes that same model to external SDKs over the protobuf protocol.

- Expose the existing router subscription model to external SDKs over the protobuf protocol.
- Push service access changes (access gained and access lost) to subscribed SDKs.
- Push service definition changes to subscribed SDKs.
- Push service posture check changes to subscribed SDKs.

PR #4057, opened 2026-06-29
fixes openziti/ziti#3990 push service and posture changes to subscrib…

andrewpmartinez:
…ed SDKs

- routers push service change (ServiceChangeSet) and posture state (PostureStateChange) notifications directly to subscribed SDK connections, removing the controller from this path
- adds RouterServiceSubscriptions and RouterDataModel capability bits and advertises them in the control-channel capabilities mask
- persists each edge router's reported capabilities mask and version with a write-on-mismatch raft update, and renders them on the edge router APIs as a capability string list and version
- normalizes a single PROCESS posture check to a one-element PROCESS_MULTI on the wire
- closes accepted edge connections on router shutdown so SDKs detect a disconnect promptly rather than waiting for a heartbeat timeout
- adds integration tests for capability reporting, service subscriptions, push/poll reconciliation, and posture-state push
- resolves each subscribing identity's effective config set (service base configs merged with the identity's per-service overrides) and includes the config bodies on pushed service defs
- adds GetIdentityServiceConfigs and refactors loadServiceConfigs to delegate to it
- config edits ride the existing service-change path: a config change pushes a service-changed to every service that references it, carrying the new body
- adds an integration test covering snapshot bodies, adding a config to a service, and editing a config shared by two services
- asserts the resolved config body is returned by GetService, not just present on the service event, in the snapshot, config-added, and config-edited cases

Depends on: 
- https://github.com/openziti/sdk-golang/pull/969
- https://github.com/openziti/edge-api/pull/202
- https://github.com/openziti/edge-api/pull/204


common/capabilities/consts.go
```diff
@@ -52,6 +52,16 @@ const (
...
 	RouterMultiChannel int = 1
+
+	// RouterServiceSubscriptions indicates the router can push service change and posture
+	// state notifications to subscribed SDK clients over the edge protocol
+	RouterServiceSubscriptions int = 2
+
+	// RouterDataModel indicates the router supports the router data model. This is also
+	// negotiated per-connection via the Header_RouterDataModel header (which additionally
+	// carries the current index); the capability bit provides a static, persistable
+	// advertisement of the same support.
```
plorenz: Is the intended audience for this the SDK? If so, what is the SDK expected to do differently based on that flag?

andrewpmartinez: Removed from SDK

ekoby (approved):

PR #4061, opened 2026-06-30
add support for using ext-jwt-signers with verify traffic

### LINKED ISSUE #4060: support ext-jwt-signer in ziti verify traffic
`ops verify ext-jwt-signer oidc` is helpful to diagnose an ext-jwt-signer that can't authenticate, but it doesn't pass data using the ephemeral certs returned by the controller for ext-jwt-signer-based identities.

Add support to `verify traffic` for ext-jwt-signer-based auth/ephemeral certificates on both the server (bind) and client (dial) sides, so the ephemeral-cert data-plane path is verified end to end.

dovholuknf:
closes #4060 

* Adds ext-jwt-signer (OIDC) support to `ziti ops verify traffic`, so the certless ephemeral-cert path can be traffic-tested.
* adds `--ext-jwt-signer <name>` to the command to pick a signer to use to auth/generate a cert with
* adds `--ext-jwt-redirect-url` (default `http://localhost:20314/auth/callback`) should you want to change that
* extracts the OIDC flow into reusable `AuthenticateWithSigner` + `TokenForSigner` shared by both commands and returns errors instead of `log.Fatal` on client/cert setup failures.

### DIFF
--- a/ziti/cmd/ops/verify/ext-jwt-signer/oidc/oidc.go
+++ b/ziti/cmd/ops/verify/ext-jwt-signer/oidc/oidc.go
@@ -397,7 +397,7 @@ func NewOidcVerificationCmd(out io.Writer, errOut io.Writer, initialContext cont
 // AuthenticateWithSigner runs the OIDC auth-code/PKCE flow for the named ext-jwt-signer
 // using the configured login options, returning the obtained tokens and the signer
 // record. Shared by the `oidc` verify command and `ops verify traffic`'s
-// --client-ext-jwt-signer path so both drive the exact same OIDC flow.
+// --ext-jwt-signer path so both drive the exact same OIDC flow.
 func (opts *OidcVerificationConfig) AuthenticateWithSigner(ctx context.Context, signerName string) (*OIDCResponse, *rest_model.ClientExternalJWTSignerDetail, error) {
 	log := pfxlog.Logger()
 
--- a/ziti/cmd/ops/verify/ext-jwt-signer/oidc/oidc.go
+++ b/ziti/cmd/ops/verify/ext-jwt-signer/oidc/oidc.go
@@ -298,90 +298,12 @@ func NewOidcVerificationCmd(out io.Writer, errOut io.Writer, initialContext cont
 			}
 			internal.ConfigureLogFormat(logLvl)
 
-			config, _, cfgErr := util.LoadRestClientConfig()
-			if cfgErr == nil {
-				// any error indicates there are probably no saved credentials.
-				id := config.GetIdentity()
-				if defaultId := config.EdgeIdentities[id]; defaultId != nil && !opts.IgnoreConfig {
-					opts.Token = defaultId.Token
-				}
-			}
-
-			if opts.ControllerUrl != "" && !strings.HasPrefix(opts.ControllerUrl, "http") {
-				opts.ControllerUrl = "https://" + opts.ControllerUrl
-			}
-			ctrlUrl, urlParseErr := url.Parse(opts.ControllerUrl)
-			if urlParseErr != nil {
-				return fmt.Errorf("invalid controller URL: %w", urlParseErr)
-			}
-
-			if err := opts.ConfigureCerts(opts.ControllerUrl, ctrlUrl); err != nil {
-				log.WithError(err).Warn("failed to configure certificates")
-				return err
-			}
-
-			m, merr := opts.NewClientApiClient()
-			if merr != nil {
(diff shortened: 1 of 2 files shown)


ziti/cmd/ops/verify/ops_verify_traffic.go
```diff
@@ -530,3 +565,195 @@ func (t *traffic) doClient(cancel context.CancelFunc) {
...
 }
+
+// doBothExtJwt runs the server (bind) and client (dial) in one process as the SAME certless
+// ext-jwt identity. It performs a single OIDC login and reuses the token for both sides,
+// which avoids a second browser flow colliding on the OIDC redirect port and the
+// bind-completes-after-the-client-already-gave-up-waiting race.
+func (t *traffic) doBothExtJwt() {
+	log.Infof("--ext-jwt-signer %q will be used for BOTH the server (bind) and the client (dial)", t.extJwtSigner)
+	t.configureService()
+	defer t.cleanupServer()
```
andrewpmartinez, line 603: Does cleanup know about the `.extjwt` files?

Also, I recall helpers in this codebase calling logging `.Fatal*()` functions, which do not trigger defer as they call `os.Exit()`.

ziti/cmd/ops/verify/ops_verify_traffic.go
```diff
@@ -530,3 +565,195 @@ func (t *traffic) doClient(cancel context.CancelFunc) {
...
+	parts := strings.Split(token, ".")
+	if len(parts) < 2 {
+		return nil
+	}
+	raw, err := base64.RawURLEncoding.DecodeString(parts[1])
+	if err != nil {
+		if raw, err = base64.StdEncoding.DecodeString(addJwtPadding(parts[1])); err != nil {
+			return nil
+		}
+	}
```
andrewpmartinez: This looks weird. If something fails `base64.RawURLEncoding.DecodeString` I am pretty sure it will also always fail `base64.StdEncoding.DecodeString`. If true, this means this recover code never recovers.

ziti/cmd/ops/verify/ext-jwt-signer/oidc/oidc.go
```diff
@@ -472,6 +394,105 @@ func NewOidcVerificationCmd(out io.Writer, errOut io.Writer, initialContext cont
 	return cmd
 }
 
+// AuthenticateWithSigner runs the OIDC auth-code/PKCE flow for the named ext-jwt-signer
+// using the configured login options, returning the obtained tokens and the signer
+// record. Shared by the `oidc` verify command and `ops verify traffic`'s
+// --client-ext-jwt-signer path so both drive the exact same OIDC flow.
```
andrewpmartinez: Does the flag '--client-ext-jwt-signer ' still exist?

ziti/cmd/ops/verify/ops_verify_traffic.go
```diff
@@ -530,3 +565,195 @@ func (t *traffic) doClient(cancel context.CancelFunc) {
...
+	_ = createServicePolicy(t.client, dialPolicyName, rest_model.DialBindDial,
+		rest_model.Roles{"@" + identityID}, rest_model.Roles{"#" + t.svcAttr()})
+	defer func() {
+		sp := mgmt.ServicePolicyFromFilter(t.client, mgmt.NameFilter(dialPolicyName))
+		deleteServicePolicy(t.client, sp)
+	}()
+
+	log.Infof("dialing %s as certless ext-jwt identity (signer %q)", t.svcName, t.extJwtSigner)
+	if err := t.startClient(t.client, t.svcName, t.extJwtConfig(token)); err != nil {
+		log.Fatalf("ext-jwt client failed: %v", err)
```
andrewpmartinez: Stylistically, `Fatal` in helpers instead of returning an error makes the helper reusable. There is always the case where no one will ever use it.

Additionally Fatals call `os.Exit()` which kills other controls like `defer()`.

andrewpmartinez (approved):
Fine, but not great.

ziti/cmd/ops/verify/ops_verify_traffic.go
dovholuknf: indeed. i see lots of fatals in this older code. i'll refactor those away to use RunE instead or just handle it better overally.

ziti/cmd/ops/verify/ext-jwt-signer/oidc/oidc.go
dovholuknf: no, missed this in a refactor. will clean

Issue #4071, opened 2026-07-01
Unify router capabilities into a single shared bitmask across control and edge channels


andrewpmartinez:
Router capabilities are advertised using a single shared capability namespace across the control channel and the edge (SDK) channel, rather than two independently numbered bitmasks that can drift apart, because a capability should mean the same thing everywhere and be defined in one place.

- Each router capability has one bit value, used identically on both the control channel and the edge channel.
- A router advertises the full capability set to both controllers and SDKs; each consumer acts on the bits it recognizes and ignores the rest.
- Capability advertisement stays wire-compatible with already-released ziti versions.
- The edge router capabilities exposed on the client edge-routers API are unchanged.

PR #4073, opened 2026-07-01
fixes openziti/ziti#4071 unify router capabilities into one shared na…

andrewpmartinez:
…mespace

- keys router capability bits off the sdk-golang RouterCapability enum
- adds GetRouterCapabilitiesMask advertising the full set on both channels
- advertises the shared mask on the control and edge channels
- keeps MultiChannel at bit 1 for control-channel back-compat with released routers

Depends on: https://github.com/openziti/sdk-golang/pull/972


zititest/go.mod
```diff
@@ -190,7 +190,7 @@ require (
 	github.com/openziti/runzmd v1.0.90 // indirect
 	github.com/openziti/secretstream v0.1.51 // indirect
 	github.com/openziti/x509-claims v1.0.3 // indirect
-	github.com/openziti/xweb/v3 v3.0.4 // indirect
+	github.com/openziti/xweb/v3 v3.0.5-0.20260618140905-54cc19903f1c // indirect
```
andrewpmartinez, line 193: Temporary hash reference to be removed after the dependent PR is merged.

plorenz (approved):
Looks good. There are some linter errors, but I think those are unrelated to your commit.

PR #4077, opened 2026-07-02
Fix flaky iat comparison in Test_API_Session_TOTP_Tokens

plorenz:
`Test_API_Session_TOTP_Tokens` intermittently fails at `tests/api_session_totp_token_test.go:195`, asserting the TOTP token's `iat` and the API session token's `iat` are within 2ms. Both are truncated to one-second precision by jwt's default `TimePrecision`, so the delta is either 0 (same wall-clock second) or >= 1s (second-boundary straddle) — the 2ms bound flakes whenever the two issuances straddle a second, which is load-dependent.

Widens the tolerance to one second, which reflects the achievable precision and the intent ("issued at approximately the same time") while still catching tokens issued far apart. Unrelated to any feature work; surfaced on an unrelated PR's CI run.

### DIFF
--- a/tests/api_session_totp_token_test.go
+++ b/tests/api_session_totp_token_test.go
@@ -191,8 +191,12 @@ func Test_API_Session_TOTP_Tokens(t *testing.T) {
 				ctx.NoError(err)
 				ctx.NotNil(issuedAt)
 
+				// The two iat claims are truncated to one-second precision (jwt's default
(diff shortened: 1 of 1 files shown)


dovholuknf (approved):
no issue to link to?

plorenz:
> no issue to link to?

I usually don't make issues for test fixes, since these changes are usually only of interest to internal devs

PR #4080, opened 2026-07-02
Mark router disconnected before rerouting its links

### LINKED ISSUE #4079: DisconnectRouter reroutes circuits before marking the router disconnected
## Summary

`Network.DisconnectRouter` (`controller/network/network.go`) runs the `RerouteLink` cascade
*before* calling `Router.MarkDisconnected(r)`. Because the router is still marked connected during
the cascade, reroute can compute a replacement path *through the very router that is being
disconnected*.

## Details

Current ordering:

```go
func (network *Network) DisconnectRouter(r *model.Router) {
    // 1: remove Links for Router
    for _, l := range r.GetLinks() {
        ...
        if wasConnected {
            network.RerouteLink(l)   // reroute runs here...
        }
    }
    // 2: remove Router
    network.Router.MarkDisconnected(r)  // ...but the router is only marked disconnected here
}
```

`RerouteLink` -> `handleRerouteLink` -> `rerouteCircuit` -> `UpdatePath` -> `shortestPath`, which
iterates `Router.AllConnected()`. Since `MarkDisconnected` (which removes the router from the
connected map) hasn't run yet, `shortestPath` and other connected-map reads still treat the dying
router as a viable node, so a rerouted circuit can be placed back onto the router that is going
away.

This is latent today but becomes load-bearing for correct reroute decisions.

## Fix

Mark the router disconnected before the reroute cascade so global router state is truthful during
reroute. Note `MarkDisconnected` also calls `routerLinks.Clear()`, so the router's links must be
snapshotted before the call, otherwise the link-removal/reroute loop would iterate an empty set.

plorenz:
Fixes #4079

`Network.DisconnectRouter` ran the reroute cascade before marking the router disconnected, so reroute could still see the disconnecting router as connected and compute a replacement path through the very router being removed. This reorders the two steps so global router state is truthful during reroute.

Because `MarkDisconnected` also clears the router's link set, the router's links are snapshotted before the call so the link-removal/reroute loop still runs.

### DIFF
--- a/controller/network/network.go
+++ b/controller/network/network.go
@@ -462,8 +462,19 @@ func (n *Network) ValidateRouterErtTerminators(filter string, cb ErtTerminatorVa
 }
 
 func (network *Network) DisconnectRouter(r *model.Router) {
-	// 1: remove Links for Router
-	for _, l := range r.GetLinks() {
+	// Snapshot the router's links before marking it disconnected: MarkDisconnected clears the
+	// router's link set (routerLinks.Clear()), so a later r.GetLinks() would return nothing and
+	// the link-removal/reroute cascade below would be skipped entirely.
+	links := r.GetLinks()
+
+	// Mark the router disconnected before the RerouteLink cascade, so reroute and everything it
+	// calls (shortestPath, connected-map reads) sees the dying router as gone. Otherwise reroute
+	// runs while the router still appears connected and can compute a replacement path through the
+	// very router that is being removed.
+	network.Router.MarkDisconnected(r)
+
+	// remove Links for Router, rerouting circuits off any that were connected
+	for _, l := range links {
 		wasConnected := l.CurrentState().Mode == model.Connected
 		if l.Src.Id == r.Id {
 			network.Link.Remove(l)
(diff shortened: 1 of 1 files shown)


controller/network/network.go
```diff
@@ -462,8 +462,19 @@ func (n *Network) ValidateRouterErtTerminators(filter string, cb ErtTerminatorVa
 }
 
 func (network *Network) DisconnectRouter(r *model.Router) {
-	// 1: remove Links for Router
-	for _, l := range r.GetLinks() {
+	// Snapshot the router's links before marking it disconnected: MarkDisconnected clears the
+	// router's link set (routerLinks.Clear()), so a later r.GetLinks() would return nothing and
+	// the link-removal/reroute cascade below would be skipped entirely.
+	links := r.GetLinks()
```
dovholuknf, line 468: there's no semaphore to capture before doing this work? what if a link (or many) sneak in between the GetLinks and MarkDisconnected?

plorenz: if we're marking disconnected then the control channel is closed, meaning there's no mechanism for links to be reported from the router. 

If links are reported once the router is marked disconnected, they won't be added and the chances of a new link being used in the timeframe is very small. 

There may still be a window, but this closes the window most of the way. Additionally, I'm reworking this logic elsewhere, so further refinement will be done there.

dovholuknf (approved):

PR #4082, opened 2026-07-02
Make circuit endpoint cleanup endpoint-scoped

### LINKED ISSUE #4081: cleanupXgressCircuit tears down the whole circuit, killing a co-located terminator on single-router circuits
## Summary

`edgeClientConn.cleanupXgressCircuit` (`router/xgress_edge/listener.go`) tears down *every*
forwarding destination for a circuit, even though it is invoked for a single closing endpoint.
On a single-router circuit (ingress and terminator on the same router, hosted by two different
edge connections), closing one side destroys the co-located other side.

## Details

`cleanupXgressCircuit` calls `self.forwarder.EndCircuit(circuitId)`, which is
`UnregisterDestinations(circuitId)` — it removes *all* destinations linked to that circuit id and
fires `Unrouted()` on each (which sends `StateClosed` to that endpoint's SDK).

For a multi-router circuit this is fine: the router hosts only one endpoint of the circuit. But
for a single-router circuit where a dialer SDK and a hosting SDK are both connected to the same
router `R` (`Path.Nodes[0] == Path.Nodes[len-1]`, common in single-ER deployments), both the
dialer-side and terminator-side `xgEdgeForwarder`s register destinations against the same circuit
id in the router-global forwarder. When the dialer's edge channel closes, `cleanupXgressCircuit`
runs `EndCircuit` and removes the terminator's destination too, firing its `Unrouted()` and
sending `StateClosed` to the hosting SDK — tearing down a perfectly healthy terminator.

## Fix

Make cleanup endpoint-scoped: unregister only the closing endpoint's destination (by its
`address`), preserving any other endpoint of the same circuit still registered on this router.
The forwarder already keys destinations by address, so both ends already coexist; only the
cleanup path was over-broad. Behavior is identical for multi-router circuits (they only ever have
one endpoint on a given router).

plorenz:
Fixes #4081

`cleanupXgressCircuit` called `EndCircuit`, which removes *every* forwarding destination for a circuit even though it runs for a single closing endpoint. On a single-router circuit (ingress and terminator on the same router, hosted by two different edge connections), closing one side also tore down the co-located other side.

This adds an endpoint-scoped `Forwarder.UnregisterDestination(circuitId, address)` and switches `cleanupXgressCircuit` to it, so only the closing endpoint is retired. Behavior is identical for multi-router circuits (a router only ever hosts one endpoint of a given circuit); for single-router circuits the co-located terminator is preserved instead of being torn down locally when one side's channel closes.

### DIFF
--- a/router/forwarder/endpoint_cleanup_test.go
+++ b/router/forwarder/endpoint_cleanup_test.go
@@ -1,0 +1,254 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package forwarder
+
+import (
+	"fmt"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/openziti/sdk-golang/v2/xgress"
+)
+
+// fakeXgressDestination is a minimal XgressDestination that records whether Unrouted was called.
+type fakeXgressDestination struct {
+	unrouted atomic.Bool
+}
+
+func (f *fakeXgressDestination) SendPayload(*xgress.Payload, time.Duration, xgress.PayloadType) error {
+	return nil
+}
+func (f *fakeXgressDestination) SendAcknowledgement(*xgress.Acknowledgement) error { return nil }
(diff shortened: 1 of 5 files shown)


router/forwarder/endpoint_cleanup_test.go
```diff
@@ -0,0 +1,254 @@
...
+		t.Fatal("expected circuit entry to be removed once its last address was unlinked")
+	}
+
+	// unlinking from an unknown circuit is a no-op that reports no remaining addresses
+	if remaining := dt.unlinkDestinationFromCircuit("missing", "x"); remaining {
+		t.Fatal("expected false when unlinking from an unknown circuit")
+	}
+}
+
+func TestUnregisterDestinationIsEndpointScoped(t *testing.T) {
```
andrewpmartinez, line 82: Odd that some test cases have godoc and others don't.

andrewpmartinez (approved):
- sent a finding about a bind'er that is also dialing itself. Not sure if that is a real issue.
- ednpoint removal isn't atomic. Unsure if it is an issue.

PR #4084, opened 2026-07-02
Manage router link configuration via the controller

### LINKED ISSUE #4004: Manage router link configuration via the controller
Part of #3743. Builds on #3998 (the router-side managed-config registry). Wire the router's link subsystem to the registry so its listeners and dialers are driven by controller-managed `router.link.v1` config, and propagate listener changes to peers at runtime.

Scope:
- A link `ConfigHandler` (`router/link` FactoryRegistry) implementing the registry's handler contract. Apply rebuilds the listener/dialer set wholesale; established Xlinks survive because `Listener.Close()` only closes the accept loop.
- Local `link:` YAML translated to `router.link.v1` JSON and pushed through the registry as local config at startup.
- An `UpdateLinkListeners` ctrl message so the router republishes its listener set to the controller on change; the controller re-fans via the existing PeerStateChange path. Dialer changes trigger `RescanForDialOpportunities` locally.

Out of scope (follow-up): stale-link detection and auto-GC.

plorenz:
Part of the controller-managed router configuration work (#3743), building on the router-side config registry merged in #3999.

Lets router link configuration be managed centrally via the controller instead of only local YAML. A router receives its `router.link.v1` config through the router data model and applies it to the link subsystem:

- A link ConfigHandler applies `router.link.v1`: it rebuilds the listener/dialer set wholesale, while established links survive (closing a listener only stops its accept loop).
- Local `link:` YAML is translated into `router.link.v1` and pushed through the managed-config registry at startup, so local and controller-sourced config share one path (local wins).
- A router republishes its listener set to the controller on change (new `UpdateLinkListeners` control message), and dialers are re-evaluated on link-group/listener changes.

Fixes #4004.

Draft: first of the router-link series; `router-link-gc` and `router-link-runtime-config` stack on top.


controller/model/router_model.go
```diff
@@ -32,17 +33,28 @@ import (
...
-	Listeners         []*ctrl_pb.Listener
+	Name        string
+	Fingerprint *string
+	// Listeners is the router's currently-advertised link listener set.
+	// Hello carries the initial snapshot; UpdateLinkListeners pushes
+	// mid-session changes from the router. Protected by mu — callers
+	// should use SetLinkListeners / GetLinkListeners rather than direct
+	// field access. The field is left exported for backward compat
+	// with code that hasn't been migrated.
+	Listeners []*ctrl_pb.Listener
```
andrewpmartinez: If we have accessors for this, why is it public?

controller/model/testing.go
andrewpmartinez, line 10: What an odd file that I assume is compiled into the release binaries and then pulls all the testing libs in with it.

andrewpmartinez (approved):
- gc potential leak sent in dm
- router.Listener export used in tests instead of using accessor w/ lock

Small. GC issue is unconfirmed, I'l let you decide.

controller/model/testing.go
plorenz: This is a longer standing issue that needs to be addressed. We do have test utilities that we want to share, but there are better ways to package it.

controller/model/router_model.go
plorenz: fixed

controller/model/testing.go
plorenz: fixed, replaced test imports with `github.com/stretchr/testify/require`, the require.TestingT type, which covers the testing interface but doesn't bring it in. Unfortunately the hashicorp raft library brings testing in as well, so even though it's cleaned up here, it's still being imported.

PR #4085, opened 2026-07-02
Deliver the router data model to transit routers

plorenz:
Part of the controller-managed router configuration work (#3743). Independent of the router-link series (touches only the controller sync strategy), so it can land in parallel.

Today only edge routers are tracked by the data-model sync strategy, so transit (fabric-only) routers never receive their controller-managed configuration. This delivers the router data model to transit routers too:

- Creates a router sender on demand when an untracked router subscribes to the data model, so non-edge routers get their config without going through the edge connect/sync flow.
- Adds a race-safe `GetOrCreate` keyed on the control channel so concurrent subscribes and reconnects can't produce duplicate senders. The connect path adopts a sender that a subscribe already created for the same channel instead of treating it as a duplicate connect, which would leave the router with no hello and no initial sync.
- Keeps the legacy api-session/session broadcasts going to edge routers only, since routers register those handlers universally and would otherwise hold session state they never use. A router counts as edge when an edge router record exists for its id, which is now the lookup used on connect, on disconnect and on subscribe, so the three paths cannot disagree. Classification is fixed for the life of a connection.
- Removes the now-unused `RouterSender.EdgeRouter` field and generalizes connect/disconnect logging that assumed edge routers.
- Documents that the api-session-added events and the legacy session model are deprecated for removal in OpenZiti 3.0.

For #3743.


andrewpmartinez (changes requested):
The entire concept of this broker is somewhat dubious now in regards of it only handling edge router connections. 

Now that the fabric/edge split is gone, we should unify entire connection process. Having to hotpath in fabric routers here after the fact is a code smell.

I don't think we should address it here as it will have major backwards compatability problems. It is a good v3.0 candidate.


For this PR:

Things to address:

-  A subscribe during a reconnect makes the controller skip an edge router's hello and initial sync
- Edge-vs-transit classification uses a different lookup than the connect path
- The strategy's flow comment describes the removed EdgeRouter field and an edge-only connect path
- RouterSender's godoc still says it represents an edge-router connection


Details sent in DMs.

plorenz:
> The entire concept of this broker is somewhat dubious now in regards of it only handling edge router connections.
> 
> Now that the fabric/edge split is gone, we should unify entire connection process. Having to hotpath in fabric routers here after the fact is a code smell.
> 
> I don't think we should address it here as it will have major backwards compatability problems. It is a good v3.0 candidate.
> 
> For this PR:
> 
> Things to address:
> 
>     * A subscribe during a reconnect makes the controller skip an edge router's hello and initial sync
> 
>     * Edge-vs-transit classification uses a different lookup than the connect path
> 
>     * The strategy's flow comment describes the removed EdgeRouter field and an edge-only connect path
> 
>     * RouterSender's godoc still says it represents an edge-router connection
> 
> 
> Details sent in DMs.

Issues addressed. Agree we'll have to revisit the notion of the broken in the future.

controller/sync_strats/sync_instant.go
```diff
@@ -600,20 +615,57 @@ func (strategy *InstantStrategy) ReceiveResync(routerId string, _ *edge_ctrl_pb.
...
+	rtx, created := strategy.rtxMap.GetOrCreate(routerId, router.Control, func() *RouterSender {
+		newRtx := newRouterSender(strategy.ae, router, strategy.RouterTxBufferSize, strategy.GetRouterDataModel())
+		newRtx.SetIsOnline(true)
+		// A subscribe request is proof the router wants the data model, so mark it supported.
+		// Set before the sender is added to rtxMap (inside GetOrCreate) so the model-event
+		// fanout can't observe SupportsRouterModel mid-write.
+		newRtx.SupportsRouterModel = true
+		newRtx.setEdge(isEdge)
+		return newRtx
+	})
```
andrewpmartinez, line 659: Should a transit router receive identities, services, policies and posture checks at all?

controller/sync_strats/sync_instant.go
```diff
@@ -429,7 +444,7 @@ func (strategy *InstantStrategy) ApiSessionAdded(apiSession *db.ApiSession) {
 		Sequence: 0,
 	}
 
-	strategy.rtxMap.Range(func(rtx *RouterSender) {
+	strategy.rtxMap.RangeEdge(func(rtx *RouterSender) {
 		_ = strategy.sendApiSessionAdded(rtx, false, state, []*edge_ctrl_pb.ApiSession{apiSessionProto})
 	})
 }
```
andrewpmartinez, line 450: The PR/commit descriptions don't state how we decide which routers continue to receive legacy sessions.

controller/sync_strats/router_tx_map_test.go
```diff
@@ -0,0 +1,189 @@
...
+func isStopped(rtx *RouterSender) bool {
+	select {
+	case <-rtx.closeNotify:
+		return true
+	default:
+		return false
+	}
+}
+
+func TestRouterTxMap_GetOrCreate(t *testing.T) {
```
andrewpmartinez: These test names don't match the rest of the package.

The existing tests name the function under test, like Test_filterEventsForRouter_dropsUnrelatedConfigs.

controller/sync_strats/sync_instant.go
andrewpmartinez: I suppose it doesn't matter. Adding another layer to the RMD to slim the model would add complexity. Hmm.

andrewpmartinez (approved):
The comment about transit routers getting full RDMs is interesting but I don't know if it is meaninful at this time.

controller/sync_strats/sync_instant.go
plorenz: Also if we unify the model, we'd have to resync as things moved to/from edge

controller/sync_strats/sync_instant.go
plorenz: fixed

controller/sync_strats/router_tx_map_test.go
plorenz: fixed

Issue #4067, opened 2026-07-01
JWKS peer signer kid is undecodable raw bytes (should be hex) -- %s vs %x on sha1.Sum in oidc_auth/storage.go KeySet()


bjornmage:
## Summary

`HybridStorage.KeySet()` in `controller/oidc_auth/storage.go` renders a peer
controller's signing-cert key id as raw binary bytes instead of a hex string,
producing an undecodable `kid` in the `/oidc/keys` (JWKS) response for every
entry sourced from `GetPeerSigners()`.

## Root cause

```go
// controller/oidc_auth/storage.go, KeySet()
for _, cert := range signers {
    kid := fmt.Sprintf("%s", sha1.Sum(cert.Raw))   // <-- bug: %s on a [20]byte
    ...
}
```

`sha1.Sum()` returns `[20]byte`. `fmt.Sprintf("%s", ...)` on a byte array
formats it as a raw (often non-UTF8, non-printable) string of those 20 bytes,
not a hex-encoded string. Notably, the error-path logging four lines below
this same function already does it correctly:

```go
pfxlog.Logger().
    WithField("issuer", cert.Issuer).
    WithField("subject", cert.Subject).
    WithField("kid", fmt.Sprintf("%x", sha1.Sum(cert.Raw))).   // <-- correct: %x
    ...
```

so the fix is a one-character change: `%s` -> `%x` on line ~992.

## Impact

- `KeySet()`'s "this controller" entry (`s.signingKey`, via
  `jwtsigner.TlsJwtSigner.KeyId()`) is unaffected -- self key ids are already
  hex-encoded correctly through a different code path.
- Every **peer** signer entry (`GetPeerSigners()`) gets this corrupt,
  non-hex `kid`.
- This affects OIDC verification wherever more than one controller can serve
  JWKS / verify each other's tokens (e.g. a raft-based HA controller
  cluster): a JWKS lookup landing on a controller other than the one that
  signed the token in hand sees the actual signer's key advertised under
  this corrupt `kid` and cannot match it against the token's `kid` header,
  causing signature/key-lookup verification to fail.

## Versions affected

- Confirmed present in `v2.0.0` (GA, released 2026-05-20).
- Confirmed still present in `main` at commit `9e5fa89` (2026-07-01).

## Suggested fix

```diff
- kid := fmt.Sprintf("%s", sha1.Sum(cert.Raw))
+ kid := fmt.Sprintf("%x", sha1.Sum(cert.Raw))
```

[...]

PR #4068, opened 2026-07-01
fix(oidc_auth): hex-encode peer signer kid in KeySet()

bjornmage:
## Summary

One-character fix for #4067: `KeySet()` in `controller/oidc_auth/storage.go`
formatted a peer signer's SHA1 cert fingerprint with `%s` instead of `%x`,
producing an undecodable raw-byte `kid` in the `/oidc/keys` JWKS response for
every entry sourced from `GetPeerSigners()`. The error-path logging four lines
below this same function already used `%x` correctly, which is how this was
spotted.

```diff
- kid := fmt.Sprintf("%s", sha1.Sum(cert.Raw))
+ kid := fmt.Sprintf("%x", sha1.Sum(cert.Raw))
```

## Impact

This affects OIDC verification wherever more than one controller can serve
JWKS / verify each other's tokens (e.g. a raft-based HA controller cluster).
A JWKS lookup landing on a controller other than the one that signed the
token in hand sees the real signer's key advertised under this corrupt peer
`kid` and cannot match it against the token's `kid` header, failing
verification. Each controller correctly renders its own key as a clean
40-char hex `kid`, but renders each peer's key as raw undecodable bytes.

## Test plan

- `gofmt -l` clean on the changed file.
- Manually confirmed the JWKS output: peer entries show corrupt (non-hex,
  non-printable) `kid` values while self entries are clean hex, consistent
  with the `%s`/`%x` mismatch.

Fixes #4067

PR #4099, opened 2026-07-11
backport: fix peer-signer JWKS kid format (#4068) to release-v2.0.x

bjornmage:
fix openziti/ziti#4067

Backports commit 4a8b63d9 (#4068, merged to `main` 2026-07-08) to `release-v2.0.x`, which is not yet on the fix and has no `v2.0.1` tag.

## What it fixes

`KeySet()` in `controller/oidc_auth/storage.go` formatted a peer controller's SHA1 cert fingerprint with `%s` instead of `%x` when building JWKS entries sourced from `GetPeerSigners()`, producing an undecodable raw-byte `kid`. The error-path logging four lines below already used `%x` correctly.

In a multi-controller raft HA deployment, each controller signs OIDC/OTT JWTs with its own root ctrl-plane-identity key (`GetRootTlsJwtSigner()`, independent of raft role). A JWKS lookup landing on a controller other than the signer sees the true signer's key advertised under this corrupt `kid` in `/oidc/keys` and cannot match it against the token's `kid` header, failing verification.

## Cherry-pick provenance

Straight cherry-pick (`-x`) of 4a8b63d9 onto `release-v2.0.x` (tip `35c78c834`) — applied with no conflicts, no adaptation needed.

## Test evidence

- `go build ./controller/...` — clean
- `go test ./controller/oidc_auth/...` — `ok`

## Production context

We run this controller in a 3-node raft HA cluster and hit this exact bug: it forces our client-facing Service to pin to a single controller instead of load-balancing across the cluster, since only same-node JWKS lookups succeed. We'll report deployment verification (JWKS kids resolving correctly across all 3 nodes) on #4067 once we've rolled a patched build.

### DIFF
--- a/controller/oidc_auth/storage.go
+++ b/controller/oidc_auth/storage.go
@@ -989,7 +989,7 @@ func (s *HybridStorage) KeySet(_ context.Context) ([]op.Key, error) {
 	signers := s.env.GetPeerSigners()
 
 	for _, cert := range signers {
-		kid := fmt.Sprintf("%s", sha1.Sum(cert.Raw))
+		kid := fmt.Sprintf("%x", sha1.Sum(cert.Raw))
 
 		if _, found := s.keys.Get(kid); found {
 			continue


bjornmage:
Deployment verification, as promised above and on #4067.

## Environment

3-node raft HA controller cluster (v2.0.0 base), one voter per node. We carried this backport as a patched image (`v2.0.0` + cherry-picked `4a8b63d9`) ahead of a tagged `release-v2.0.x` release, and rolled it one node at a time (staggered, single-voter-at-a-time recreate) to keep quorum intact through the version-mixed window.

## BEFORE (stock v2.0.0, all 3 nodes)

Pulled `/oidc/keys` from all 3 controllers and matched entries by RSA modulus (`n`). Each of the 3 nodes' own signing key rendered a clean hex `kid` on exactly one node (itself) and a corrupted/undecodable `kid` on the other two — 3 keys × 3 nodes = 3 clean + 6 mangled, stable across repeated calls. This lines up exactly with the bug mechanism described above: a JWKS lookup landing on a controller other than the signer sees the signer's key advertised under the corrupt `kid`.

Operationally this forced us to pin our client-facing Service to a single controller instead of load-balancing across the cluster, since only same-node JWKS lookups succeeded.

## AFTER (patched image on all 3 nodes)

Repeated the same 3×3 check once all 3 controllers were running the patched image. All 9 entries now render identical clean-hex `kid`s, matched by modulus, from every node. No mangled entries anywhere in the matrix.

## Mixed-version window

During the staged rollout (patched nodes and unpatched nodes coexisting under raft), we saw no corrupted `kid`s in a patched node's own JWKS output at any point, including while peer nodes were still unpatched. Worth noting for anyone doing the same node-by-node rollout under HA: the fix appears to be entirely local to the serving node's own encode/decode path and doesn't depend on peer nodes also being patched.

Cluster-wide raft state (quorum, leadership) stayed healthy through all three node recreates.

kiegsgroot:
Independent production confirmation of the underlying bug: on our 3-node v2.0.0 HA cluster, `/oidc/keys` on each controller serves its own kid correctly but both peers' kids as raw bytes — byte-for-byte the hex-decode of the true kids, matching #4067 exactly. We hit it while diagnosing a cluster-wide `public key not found` outage triggered by the default-enabled cert-renewal timer (that turned out to be a distinct staleness bug — filed as #4116 with a full code walk; packaging fix in #4117). Worth noting for reviewers: since the router-side RDM computes kids correctly, this backport alone doesn't prevent the renewal outage — but it's still needed for every JWKS consumer, and v2.0.1 shipped without it. Would be good to see this land in the next v2.0.x.

bjornmage:
Checking in on this — it's been open a couple weeks with no maintainer response.

For context on why this matters beyond our own report: v2.0.2 shipped yesterday (2026-07-27) with two unrelated security fixes (GHSA-mrpr-756c-xm47, GHSA-cc5m-7mhm-xh9f), and it still doesn't include this JWKS `kid` fix — so the `release-v2.0.x` line now has no path to it short of this backport landing. v2.1.0-pre1 also just went out (2026-07-28) as a pre-release, so anyone running 2.0.x in production HA who needs this can't yet move to a GA 2.1.0 either.

Between our own deployment verification above and @kiegsgroot's independent confirmation of the same corrupted-`kid` symptom on a separate 3-node cluster, this looks solidly validated. Is there anything blocking review, or additional evidence that would help move it along?

dovholuknf:
sorry, it's easy for us to miss github notifications - we get many. I happened to see this one, thanks for the poke... i see this is identical to https://github.com/openziti/ziti/pull/4068/changes.

dovholuknf (approved):

plorenz:
Hi @bjornmage , I think it's ready to merge, we just require signed commits. As soon as that is fixed, we can merge.

frdbonif:
We've independently reproduced this on a 3-node Raft HA deployment running v2.0.3.

Each controller exposes the same three RSA signing keys via `/oidc/keys`, but each node reports a valid hex `kid` only for its own signing key; the two peer keys have the undecodable raw-byte `kid` values described in #4067.

This results in intermittent cross-controller JWT verification failures:

`token is unverifiable: error while executing keyfunc: public key not found`

Pinning clients to a single controller avoids the problem, but defeats the HA/load-balancing behaviour.

Given that #4099 already has the backport applied cleanly, passes the relevant tests, has maintainer approval, and has independent 3-node HA deployment verification, I thought another real-world confirmation might be useful.

Hopefully the signed-commit requirement can be sorted and this can make it into the next 2.0.x release. This is a fairly significant operational issue for HA deployments.

Issue #4088, opened 2026-07-03
Identity revocation timestamps diverge across controllers (non-deterministic raft apply)


plorenz:
## Summary

In an HA (multi-controller) cluster, deleting or disabling an identity produces a `DataState_Revocation` whose `ExpiresAt` and `IssuedBefore` timestamps **differ on each controller**. The RouterDataModel validation surfaces this as persistent `revocations ... ExpiresAt/IssuedBefore mismatch` diffs between the controller and router copies, and the router-data-model validation suite fails.

## Root cause

The `IdentityRevocationConstraint` (added with the identity/api-session revocation enforcement work) runs as a store pre-commit constraint, i.e. *inside* the bolt transaction of the identity delete/disable command. Ziti replicates **commands**, not database pages: `BoltDbFsm.Apply` re-runs `cmd.Apply(ctx)` on every raft node. The constraint sets the revocation's `ExpiresAt`/`IssuedBefore` from `time.Now()`, so each controller independently derives its own timestamp during apply. The values diverge by the raft apply latency (sub-second to ~1s), which is exactly what the RDM diff reports.

This is a raft state-machine determinism violation, not just a test artifact: the three controllers genuinely enforce three different revocation cutoffs for the same identity.

## Impact

- router-data-model validation fails on multi-controller clusters.
- The identity revocation cutoff (`IssuedBefore`) and expiry are inconsistent across controllers and the routers that sync from them.

## Note

The OIDC/management revocation paths (`RevocationManager.CreateOrReplace`) are unaffected: they compute the timestamp once on the leader and carry it in the replicated command, which is already deterministic. Only the store-constraint path recomputes per node.

PR #4103, opened 2026-07-14
Load revocations when rebuilding the router data model

### LINKED ISSUE #4102: Controller RouterDataModel rebuild (BuildAll) drops all existing revocations
## Summary

`InstantStrategy.BuildAll` (`controller/sync_strats/sync_instant.go`) rebuilds a controller's in-memory `RouterDataModelSender` from the bolt store, but it does **not** load revocations. It builds config-types, configs, identities, services, posture-checks, service-policies, and public-keys; there is no `BuildRevocations` and none is called. Revocations only ever enter the in-memory model via live `RevocationCreate` store events.

## Impact

Any time a controller (re)builds its RDM (startup, becoming leader, snapshot install), it loses every revocation that already existed in the store. Revocations created afterward (live) are unaffected.

- **Revocation enforcement gap (primary concern):** a router that syncs from a freshly-rebuilt controller never receives the pre-existing revocations, so a revoked/disabled identity's sessions are not enforced on those routers after a controller restart or leadership change.
- **Stale revocations on routers:** routers that already hold such revocations keep them, and the rebuilt controller cannot tell them to drop them (it does not know they exist). They clear only when the revocation eventually expires and the store delete fires a delete event, so the window is the full revocation TTL.

## Fix

Add a `BuildRevocations` that iterates the revocation store and applies each entry into the RDM (mirroring `BuildIdentities` etc.), and call it from `BuildAll`. Skip already-expired revocations at build time so a rebuild does not reintroduce entries the enforcer would immediately purge.

Present on `main` and `release-v2.0.x`. Related to #4088.

plorenz:
Fixes #4102

## Problem

`InstantStrategy.BuildAll` rebuilds a controller's in-memory router data model from the store but never loaded revocations, so any revocation already in the store was dropped whenever a controller rebuilt its model (startup, becoming leader, snapshot install). A router syncing from a rebuilt controller then stopped enforcing those revocations, and stale revocations lingered on routers that already had them until expiry.

Found via the router-data-model validation suite: routers held revocations the controller's data state lacked, persistently, and revocations were the only entity type affected because they were the only type `BuildAll` omitted.

## Change

- Add `BuildRevocations` and call it from `BuildAll`, loading each stored revocation into the data model (same pattern as the other entity types).
- Skip already-expired revocations at load time so a rebuild doesn't reintroduce entries the enforcer would immediately purge.
- Add a test covering the load and the expiry skip.

Related to #4088.

### DIFF
--- a/controller/sync_strats/sync_instant.go
+++ b/controller/sync_strats/sync_instant.go
@@ -991,6 +991,55 @@ func (strategy *InstantStrategy) BuildPublicKeys(tx *bbolt.Tx, rdm *common.Route
 	return nil
 }
 
+// BuildRevocations loads every unexpired revocation from the store into the
+// router data model. Without it, a data model rebuilt from the store (on startup,
+// on becoming leader, or after a snapshot install) would silently omit every
+// revocation already persisted, so routers syncing from that controller would
+// stop enforcing them.
+func (strategy *InstantStrategy) BuildRevocations(tx *bbolt.Tx, rdm *common.RouterDataModelSender) error {
+	return loadRevocations(strategy.ae.GetStores().Revocation, tx, rdm, time.Now())
+}
+
+// loadRevocations applies each revocation in the store into rdm, skipping any
+// that have already expired. Expired revocations are left out because the
+// revocation enforcer purges them, so re-adding them here would only strand them
+// on routers until the next purge.
+func loadRevocations(store db.RevocationStore, tx *bbolt.Tx, rdm *common.RouterDataModelSender, now time.Time) error {
+	for cursor := store.IterateIds(tx, ast.BoolNodeTrue); cursor.IsValid(); cursor.Next() {
+		id := string(cursor.Current())
+
+		revocation, err := store.LoadById(tx, id)
+		if err != nil {
+			return err
+		}
+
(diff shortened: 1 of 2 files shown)


controller/sync_strats/sync_instant_revocations_test.go
```diff
@@ -0,0 +1,76 @@
+/*
```
andrewpmartinez, line 1: Tests don't cover BuildAll() to ensure the revocations are actually used.

controller/sync_strats/sync_instant.go
```diff
@@ -991,6 +991,55 @@ func (strategy *InstantStrategy) BuildPublicKeys(tx *bbolt.Tx, rdm *common.Route
 	return nil
 }
 
+// BuildRevocations loads every unexpired revocation from the store into the
+// router data model. Without it, a data model rebuilt from the store (on startup,
+// on becoming leader, or after a snapshot install) would silently omit every
```
andrewpmartinez, line 1089: Is "on becoming the leader" true? This should only be done at startup.

andrewpmartinez (approved):
Comments on small things and additionally should `ValidateAll` be updated for this? It validates the rest of the router data model.

PR #4115, opened 2026-07-16
Guard tunnel dialer against a nil hosted-service registry

### LINKED ISSUE #4114: Router panics on ERT-terminator inspect when it hosts no tunnel services (1.6.x)
## Summary

An edge router that does not host tunnel services (no tunnel listener configured, e.g. a pure
initiator router) panics with a nil-pointer dereference when it receives an ERT-terminators inspect
request. This is reachable via terminator validation (`ziti fabric validate terminators` / the mgmt
`ValidateRouterErtTerminators` flow), which inspects every connected router, so a single validate call
can crash all non-hosting routers in a network.

## Stack trace

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20]

xgress_edge_tunnel_v2.(*HostedServiceRegistry).queueWithTimeout(0x0, ...)  hosted.go:394
xgress_edge_tunnel_v2.(*HostedServiceRegistry).Inspect(0x0, ...)           hosted.go:628
xgress_edge_tunnel_v2.(*tunneler).Inspect(...)                             dialer.go
handler_ctrl.(*inspectRequestContext).inspectXgressDialer("tunnel", ...)   inspect.go:101
handler_ctrl.(*inspectHandler).HandleReceive(...)
```

## Root cause

The tunnel factory wires `tunneler.hostedServices` in `CreateListener`, which is only called on routers
that have a tunnel listener configured. On a router without one, `tunneler.hostedServices` stays nil.
`tunneler.Inspect` then calls `self.hostedServices.Inspect(...)` on the nil registry and the router
crashes.

## Affected versions

1.6.x (and any release where the tunnel `HostedServiceRegistry` is wired lazily in `CreateListener`).
2.0 / main are not affected: the registry is created and wired eagerly in `NewFactory`
(`router/xgress_edge_tunnel/factory.go`), so it is never nil.

## Fix

Guard `tunneler.Inspect` so it returns no ERT terminators when the router hosts none (nil registry),
instead of dereferencing it.

plorenz:
Fixes #4114

A router that hosts no tunnel services (no tunnel listener, e.g. an initiator router) has a nil `tunneler.hostedServices`, because it's only wired in `CreateListener`. Control plane requests for tunnel terminators reach every router, not just hosting ones, so several dialer entry points can dereference the nil registry.

The clearest case is an ERT-terminator inspect: `ziti fabric validate router-ert-terminators` inspects every connected router, and a non-hosting router panics (SIGSEGV). Terminator validation and circuit dials hit the same registry, so they're exposed too if the controller still holds a `tunnel` binding terminator for a router that no longer hosts.

This makes all of those paths tolerate a missing registry:

- `Inspect` returns an empty ERT terminator result, so validation reports zero terminators for the router rather than failing it. That also lets the controller flag its own stale terminator records for that router, which a failed result skips.
- `IsTerminatorValid` returns false and `Dial` returns an invalid terminator error.
- create tunnel terminator responses are ignored. These were also registered as a method value on a nil receiver at bind time.

The registry is now stored atomically, since it's set while the listener is being created, which can race with control channel requests.

2.0/main aren't affected: the registry is wired eagerly in `NewFactory` there, with `Start()` split out of the constructor. 1.6.x starts the registry inside its constructor, so wiring it eagerly would put an event loop on every router, hence the guards instead.

Also includes an unrelated CI fix: codespell started flagging the adopter name Actieve in `ADOPTERS.md` as a misspelling of "active", failing the check on every PR. The name is added to the ignore list. `main` and the other release branches need the same change.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -14,5 +14,5 @@ jobs:
       - name: Run code spelling check
         uses: codespell-project/actions-codespell@v2
         with:
-          ignore_words_list: allos,ans,dne,noe,referr,ssudo,te,tranfer,ue
+          ignore_words_list: actieve,allos,ans,dne,noe,referr,ssudo,te,tranfer,ue
           skip: go.*,zititest/go.*,./controller/storage/zitiql/zitiql_parser.go
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# Release 1.6.20
+
+## What's New
+
+* Fixes a router panic on routers which aren't hosting tunnel services
+
+## Component Updates and Bug Fixes
+
+* github.com/openziti/ziti: [v1.6.19 -> v1.6.20](https://github.com/openziti/ziti/compare/v1.6.19...v1.6.20)
+    * [Issue #4114](https://github.com/openziti/ziti/issues/4114) - Router panics on ERT-terminator inspect when it hosts no tunnel services (1.6.x)
+
 # Release 1.6.19
 
 ## What's New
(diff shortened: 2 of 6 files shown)


andrewpmartinez (approved):
The PR is small and clean here. 

I sent over details of potential readside and related issues that are outside the PR however.

Issue #4126, opened 2026-07-22
[Backport-2.0] Legacy create-session signs service JWT with a mismatched session id after dedup


plorenz:
Backport the fix for #4125 to `release-v2.0.x`.

The legacy create-session JWT/session-id mismatch affects the 2.0.x line, where legacy-authenticated clients surviving a single-node -> multi-node HA cluster expansion hit a permanent `invalid session` retry storm. The fix (order `Session.Create` before `CreateJwt` so the token is signed with the durable session id) applies cleanly to `release-v2.0.x`.

Tracks the 2.0.x backport of #4125.

PR #4127, opened 2026-07-22
Sign legacy session JWT with the durable session id

### LINKED ISSUE #4125: Legacy create-session signs service JWT with a mismatched session id after dedup
## Summary
In the edge client `POST /sessions` (legacy) flow, the service-access JWT is signed with the session id **before** the durable session is created. When the create dedups to a pre-existing session for the same `(api-session, type, service)`, the JWT keeps the freshly generated id while the stored record uses the existing id. The client is left holding a token whose `jti` matches no stored session, so every subsequent `create-circuit` / `create-terminator` fails with `invalid session`.

## Impact
Legacy-authenticated SDK clients (authenticated before OIDC was available and still on their legacy api-session) can enter a permanent `invalid session` retry storm. Reproduced when expanding a single-node controller into a multi-node HA cluster: the cluster churn triggers a session re-create while the old durable record still exists, the dedup fires, and the client's dial/bind traffic stops working. Affects both dial (`create.circuit`) and bind (`create.terminator`).

## Root cause
`controller/internal/routes/session_router.go` `Create`:
1. `MapCreateSessionToModel` assigns a fresh cuid as `entity.Id`.
2. `CreateJwt` signs the JWT with `jti = entity.Id`.
3. `Session.Create` runs `getExistingSessionEntity`; on a dedup hit it overwrites `entity.Id` with the existing session's id and returns without persisting a record under the new id.

Result: `jti` (new id) != persisted session id (existing id). `loadFromBolt` -> `Session.Read(jti)` -> not found -> `invalid session`.

## Fix
Run `Session.Create` (which resolves the id via dedup or persists) before `CreateJwt`, so the token is always signed with the id of the durable record. Preserves delete-as-revoke semantics (still a real durable session; no bypass).

## Affected
- main and release-v2.0.x (2.0.x line).

plorenz:
Fixes #4125

The legacy edge client `POST /sessions` flow signed the service-access JWT with a freshly generated session id **before** creating the durable session. When the create deduped to an existing session for the same `(api-session, type, service)`, the stored record kept the existing id while the token kept the new one, so the client held a token whose id backed no session. Every subsequent `create-circuit`/`create-terminator` then failed with `invalid session`.

This bit legacy-authenticated clients after a single-node controller was expanded into a multi-node HA cluster: the churn triggered a session re-create while the old durable record still existed, the dedup fired, and dial/bind traffic stopped with a permanent retry storm.

The fix reorders the flow so `Session.Create` resolves the id (dedup or persist) before `CreateJwt` signs the token. Delete-as-revoke semantics are preserved, the session stays a durable record loaded via `Session.Read`.

Includes an apitests regression test asserting the returned token id matches the persisted session id, including on a deduped create.

Backport to release-v2.0.x tracked in #4126.

### DIFF
--- a/controller/internal/routes/session_router.go
+++ b/controller/internal/routes/session_router.go
@@ -188,6 +188,19 @@ func (r *SessionRouter) Create(ae *env.AppEnv, rc *response.RequestContext, para
 	}
 
 	entity := MapCreateSessionToModel(identity.Id, apiSession.Id, params.Session)
+
+	// Legacy sessions are backed by a durable record, and the JWT must be signed with that
+	// record's id. Create must run before CreateJwt: its dedup can resolve entity.Id to a
+	// pre-existing session for the same (api-session, type, service) rather than the freshly
+	// generated id. Signing first would mint a token whose id has no backing session, so every
+	// subsequent create-circuit/create-terminator would fail to load the session.
+	if rc.HasLegacySecurityToken() {
+		if _, err = ae.Managers.Session.Create(entity, rc.NewChangeContext()); err != nil {
+			rc.RespondWithError(err)
+			return
+		}
+	}
+
 	jwtStr, err := ae.Managers.Session.CreateJwt(entity, rc.HasLegacySecurityToken())
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


controller/internal/routes/session_router.go
andrewpmartinez, line 207: We don't clean up the durable record if we err here?

andrewpmartinez (approved):
One comment about record clean up. Otherwise fine.

Less worried about it as this shouldn't happen often.

controller/internal/routes/session_router.go
plorenz: I thought about it, and since edge sessions are idempotent, it's probably better to just leave it. They're likely going to retry and this way we don't delete and then immediately recreate. It also ensure we don't delete and edge session that may be used by another request. Either way, it'll get cleaned up with the api session is deleted, so it's not long lived.

PR #4128, opened 2026-07-22
Sign legacy session JWT with the durable session id (backport to 2.0.x)

### LINKED ISSUE #4126: [Backport-2.0] Legacy create-session signs service JWT with a mismatched session id after dedup

plorenz:
Backport of #4127 to `release-v2.0.x`. Fixes #4126.

Cherry-pick of the fix for #4125. The legacy create-session flow signed the service JWT with a freshly generated session id before the durable Create ran; on a dedup to an existing session the token id no longer matched the stored record, so legacy clients hit a permanent `invalid session` retry storm after a single-node -> multi-node HA cluster expansion.

Reorders `Session.Create` before `CreateJwt` so the token is signed with the durable session id. Delete-as-revoke semantics preserved. Includes the apitests regression test.

### DIFF
--- a/controller/internal/routes/session_router.go
+++ b/controller/internal/routes/session_router.go
@@ -188,6 +188,19 @@ func (r *SessionRouter) Create(ae *env.AppEnv, rc *response.RequestContext, para
 	}
 
 	entity := MapCreateSessionToModel(identity.Id, apiSession.Id, params.Session)
+
+	// Legacy sessions are backed by a durable record, and the JWT must be signed with that
+	// record's id. Create must run before CreateJwt: its dedup can resolve entity.Id to a
+	// pre-existing session for the same (api-session, type, service) rather than the freshly
+	// generated id. Signing first would mint a token whose id has no backing session, so every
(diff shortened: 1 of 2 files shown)


andrewpmartinez (commented):
See main PR #4127 for cleanup comment.

plorenz:
> See main PR #4127 for cleanup comment.

From main PR:

I thought about it, and since edge sessions are idempotent, it's probably better to just leave it. They're likely going to retry and this way we don't delete and then immediately recreate. It also ensure we don't delete and edge session that may be used by another request. Either way, it'll get cleaned up with the api session is deleted, so it's not long lived.

andrewpmartinez (approved):

Issue #4146, opened 2026-07-23
[Backport-2.0] Upgraded controller rejects legacy clients' existing sessions and gives no recovery signal for invalid service tokens


plorenz:
Backport the fix for #4145 to `release-v2.0.x`.

The legacy-session handling and error-classification gap in `loadFromBolt` is present on `release-v2.0.x`, so legacy clients hitting a 2.0 controller with a pre-upgrade session token are affected there too.

Tracks the 2.0.x backport of #4145.

PR #4147, opened 2026-07-23
Accept legacy sessions and signal recovery for invalid service tokens

### LINKED ISSUE #4145: Upgraded controller rejects legacy clients' existing sessions and gives no recovery signal for invalid service tokens
## Summary
On the router-facing edge control path, `loadFromBolt` (the legacy/durable api-session path) only handled **JWT** service-access tokens: it fed the session token straight to `ValidateServiceAccessToken`, which parses it as a JWT. Two problems follow when a legacy client's session predates a controller upgrade:

1. A legacy (1.6.12-era, opaque/UUID) session token can't be parsed as a JWT, so `create.circuit`/`create.terminator` fails with a generic `internalError` (`token is malformed: token contains an invalid number of segments`). The client's existing, still-valid durable session is rejected.
2. Because the failure is a generic `internalError` (retry hint `RetryDefault`), the SDK gets no signal to discard and re-create the session, so the client retries the dead token and never recovers.

Reproduced in the upgrade test: after the controller is upgraded to 2.0 with routers still on 1.6.12, a subset of legacy clients wedge dialing with their pre-upgrade session tokens, stalling traffic recovery.

## Fix
1. **Accept legacy sessions.** In `loadFromBolt`, when the session token is not JWT-prefixed, look up the durable session by token (`Session.ReadByToken`) and verify it belongs to the presented api-session, rather than parsing it as a JWT. A legacy client's existing session keeps working across the upgrade (no forced re-create).
2. **Recoverable vs infrastructure errors.** Classify token-validation failures (malformed/expired/bad claims/mismatched api-session/revoked) as `InvalidSessionError` (`RetryStartOver` -> client re-creates), while keeping revocation-store/datastore read failures as `internalError`, so a transient controller fault does not make clients discard valid sessions (no reauthentication storm). Introduces a typed `common.InvalidTokenError` to distinguish the two in `ValidateServiceAccessToken`.

## Affected
main and release-v2.0.x.

plorenz:
Fixes #4145

The router-facing `loadFromBolt` only handled JWT service-access tokens, so a legacy client whose session predates a controller upgrade failed `create.circuit`/`create.terminator` with a generic `internalError` (`token is malformed`) — its existing durable session was rejected, and the generic error gave the SDK no signal to re-create, so the client wedged.

- **Accept legacy sessions:** a non-JWT session token is now looked up via `Session.ReadByToken` (with an api-session ownership check) instead of being parsed as a JWT, so a legacy client's existing session keeps working across the upgrade.
- **Recoverable vs infrastructure errors:** token-validation failures (malformed/expired/bad claims/mismatched/revoked) map to `InvalidSessionError` (`RetryStartOver` -> re-create); revocation-store/datastore failures stay `internalError` so a transient fault doesn't trigger a client reauth storm. New typed `common.InvalidTokenError` distinguishes them.

Authorization is unchanged — the accepted legacy session still flows through `verifyEdgeRouterAccess`/policy checks. Backport to release-v2.0.x tracked in #4146.

### DIFF
--- a/common/oidc_tokens.go
+++ b/common/oidc_tokens.go
@@ -209,6 +209,26 @@ func (c *ServiceAccessClaims) HasAudience(targetAud string) bool {
 	return false
 }
 
+// InvalidTokenError indicates a token was rejected because of the token itself (bad signature or
+// parse, expiry, unexpected claims/audience/type, a mismatched api session, or revocation) rather
+// than an infrastructure failure (e.g. a datastore read) encountered while validating it. Callers
+// use it to distinguish a client that should discard and re-create its session from a transient
+// controller failure that must not invalidate otherwise-valid client state.
+type InvalidTokenError struct {
+	Err error
+}
+
+func (e *InvalidTokenError) Error() string {
+	if e.Err == nil {
+		return "invalid token"
+	}
+	return e.Err.Error()
+}
+
+func (e *InvalidTokenError) Unwrap() error {
+	return e.Err
+}
+
 type AccessClaims struct {
 	oidc.AccessTokenClaims
 	CustomClaims
--- a/controller/env/appenv.go
+++ b/controller/env/appenv.go
@@ -254,19 +254,19 @@ func (ae *AppEnv) ValidateServiceAccessToken(token string, apiSessionId *string)
 	parsedToken, err := jwt.ParseWithClaims(token, serviceAccessClaims, ae.JwtSignerKeyFunc)
 
 	if err != nil {
-		return nil, err
+		return nil, &common.InvalidTokenError{Err: err}
 	}
 
 	if !parsedToken.Valid {
-		return nil, errors.New("service access token is invalid")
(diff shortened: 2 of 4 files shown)


andrewpmartinez (approved):
This happens to work for the CSDK, but the errors being sent to it are not updated to return structured errors. So, the CSDK will remain doing `strcmp` on error messages. Specifically it keys of `invalid session` not `Invalid Session`.

It will work eventually when the clauses fail through, but we are not helping the SDK side out at all with this.

plorenz:
> This happens to work for the CSDK, but the errors being sent to it are not updated to return structured errors. So, the CSDK will remain doing `strcmp` on error messages. Specifically it keys of `invalid session` not `Invalid Session`.
> 
> It will work eventually when the clauses fail through, but we are not helping the SDK side out at all with this.

I'll put together a PR to convert this to structured errors as a follow-up, rather than conflating the two, especially given timing.

PR #4158, opened 2026-07-24
Signal congestion on terminator establish/remove timeouts

### LINKED ISSUE #4155: Router mis-signals congestion on terminator establish/remove timeouts
When the controller is slow to respond, the router resolves the rate-limit control for a terminator establish (and for the synchronous remove) as a neutral failure or a success on timeout, instead of a backoff. The adaptive window therefore doesn't shrink under congestion, so a transient controller stall can tip into a self-reinforcing stall that won't recover on its own.

Fix: measure establishment latency from the last attempt and report Backoff at or over the establishment timeout (Success below it); resolve a superseded attempt's control with Backoff rather than orphaning it; treat a synchronous remove reply timeout as congestion. Router-side; suitable for backport to release-v2.0.x.

plorenz:
Under controller slowness the router mis-signals congestion on terminator establish/remove timeouts (neutral/success instead of backoff), so the adaptive window doesn't shrink and a transient stall can become self-reinforcing. Resolves the rate-limit control with Backoff on timeout for the establish and synchronous remove paths, and stops orphaning a superseded attempt's control.

Router-side; intended for backport to release-v2.0.x.

Stacked on the sdk-hosting-test PR.

Fixes #4155

### DIFF
--- a/router/xgress_edge/fabric.go
+++ b/router/xgress_edge/fabric.go
@@ -235,10 +235,33 @@ func (self *edgeTerminator) newConnection(connId uint32) (*edgeXgressConn, error
 	return result, nil
 }
 
-func (self *edgeTerminator) SetRateLimitCallback(control rate.RateLimitControl) {
+// replaceRateLimitCallback stores control as the rate-limit control for a new establishment attempt.
+// If a control from a prior attempt is still outstanding, that attempt exceeded establishmentTimeout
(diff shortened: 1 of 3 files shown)


andrewpmartinez (approved):
There is some difference in how the GO SKD and CSDK function. I sent the details in a messages. It is outside the scope of this PR and the SDKs differ on implemntation. 

Additionally I assume thet ER/T related behavior is out of scope.

PR #4159, opened 2026-07-24
Fix SDK terminator remove-path congestion collapse

### LINKED ISSUE #4156: SDK terminator removal can enter a self-sustaining retry storm under heavy churn
Terminator removal is synchronous and, on controller rejection, retried immediately with no pacing. Under heavy churn (e.g. mass SDK-host loss), routers retry removes against the raft leader faster than it can apply them; the leader rate-limits (SERVER_TOO_MANY_REQUESTS), routers retry again, and the loop sustains itself, leaving terminators stuck and hammering the leader indefinitely.

Fix, in three parts: (1) asynchronous RemoveTerminatorsV2 behind a controller capability, with fallback to synchronous removal for older controllers; (2) exponential, jittered retry backoff so a fast-failing operation can't busy-loop; (3) a leader fast-path that skips a remove for a terminator the router confirms was created but that no longer exists, keeping no-op retries off the command rate limiter. Protocol change, so main only. Validated by a 24h chaos run.

plorenz:
Under heavy terminator churn, synchronous removal retried without pacing becomes a self-sustaining retry storm against the raft leader (SERVER_TOO_MANY_REQUESTS). Adds asynchronous RemoveTerminatorsV2 behind a controller capability (falling back to synchronous removal for older controllers), exponential jittered retry backoff, and a leader fast-path that skips confirmed-already-removed terminators to keep no-op retries off the command rate limiter. Validated by a 24h chaos run.

Protocol change; main only. Stacked on #4155's PR.

Fixes #4156


controller/handler_ctrl/remove_terminators_v2.go
```diff
@@ -0,0 +1,103 @@
...
+		log.WithError(err).WithField("terminatorIds", toDelete).
+			Error("unable to remove terminators")
+		response.Msg = err.Error()
+	}
+
+	self.sendResponse(ch, response)
+}
+
+func (self *removeTerminatorsV2Handler) sendResponse(ch channel.Channel, response *ctrl_pb.RemoveTerminatorsV2Response) {
+	if err := protobufs.MarshalTyped(response).Send(ch); err != nil {
```
andrewpmartinez, line 99: An unbounded send?

controller/handler_ctrl/base.go
```diff
@@ -55,33 +55,69 @@ func (self *baseHandler) lookupTerminatorOwner(id string) (routerId string, pres
...
 			Warn("router attempted to remove terminators it does not own; rejected")
 	}
+
+	if skipped > 0 {
+		pfxlog.Logger().
+			WithField("routerId", self.router.Id).
+			WithField("skipped", skipped).
+			WithField("remaining", len(kept)).
+			Debug("leader fast-path skipped confirmed already-removed terminators")
+	}
```
andrewpmartinez, line 85: Should we log the terminator ids that were skipped/rejected? Or even just the ones actually deleted?

Not really sure how big these id lists get.

andrewpmartinez (approved):
Looks good. Sent over a potential gap in deletes on leaders that are behind on applying their log. Not sure if it is an issue. V1 used to order against the log somehow.

controller/handler_ctrl/base.go
plorenz: I think the Warn is enough to let us know that there's a problem. The individual terminators aren't that important. That a router is trying to do things with terminators it doesn't own, is. 
The id list can get pretty big during churn events

controller/handler_ctrl/remove_terminators_v2.go
plorenz: fixed

PR #4169, opened 2026-07-25
Preserve tags on external JWT signer and transit router create

### LINKED ISSUE #4124: Managed-auth External JWT Signer created without network-id/managed tags on OpenZiti v2.0.0 controllers
**Summary**                                                                                                                                                                                                                                
                                                                                                                                                                                                                                         
When core-management provisions a network's managed-auth External JWT Signer, the create request correctly includes tags ({any-tag}), but the resulting signer on the controller has empty tags (tags: {}). The signer is  otherwise created successfully.

| Operation | HTTP | Tags after (via GET by id) |
|---|---|---|
| CREATE with tags | `201` | **`{}`** — dropped |
| PATCH with tags | `200` | persisted ✅ |
| PUT with tags | `200` | persisted ✅ |

### 1. Create with tags → `201`
```bash
curl -k -sS -L -X POST "$API/external-jwt-signers" \
  -H "zt-session: $ZT_SESSION" -H 'Content-Type: application/json' \
  -d '{
        "name": "claude-tags-diagnostic",
        "enabled": true,
        "issuer": "https://example/issuer",
        "audience": "example-aud",
        "jwksEndpoint": "https://example/.well-known/jwks.json",
        "claimsProperty": "sub",
        "useExternalId": false,
        "tags": { "managed": "NetFoundry", "network-id": "00000000-0000-0000-0000-000000000000" }
      }'
# -> {"data":{...,"id":"<ID>"},"meta":{}}
```

### 2. Read by id (detail endpoint) → tags are EMPTY
```bash
curl -k -sS -L -H 'Accept: application/json' \
  -H "zt-session: $ZT_SESSION" \
  "$API/external-jwt-signers/<ID>"
# -> "...","tags":{},"..."     <-- dropped despite being in the create body
```

plorenz:
Fixes #4124

Creating an external JWT signer with tags returned 201, but the tags were silently discarded; PUT and PATCH on the same entity persisted them fine. The create mapper built its `BaseEntity` empty and never copied tags out of the request body.

The same bug exists in the transit router create mapper, so both are fixed here. This isn't a 2.0 regression: the ext-jwt-signer case dates back to 2022.

Adds API tests that assert tags round-trip through create for both entity types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -484,6 +484,7 @@ today and expand as packages are converted. The global `ziti agent set-log-level
 * github.com/openziti/transport/v2: [v2.0.215 -> v2.0.216](https://github.com/openziti/transport/compare/v2.0.215...v2.0.216)
 * github.com/openziti/xweb/v3: [v3.0.4 -> v3.0.5](https://github.com/openziti/xweb/compare/v3.0.4...v3.0.5)
 * github.com/openziti/ziti/v2: [v2.0.0 -> v2.1.0](https://github.com/openziti/ziti/compare/v2.0.0...v2.1.0)
+    * [Issue #4124](https://github.com/openziti/ziti/issues/4124) - Tags are dropped when creating external JWT signers, transit routers and authenticators
     * [Issue #4193](https://github.com/openziti/ziti/issues/4193) - ER/T terminator create reports "invalid edge router for session" with no router-side error and slow recovery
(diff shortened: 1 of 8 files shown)


dovholuknf (approved):
seems like some level of boundary/negative tests would be useful in here as well if we could

dovholuknf (approved):

PR #4194, opened 2026-07-29
Make ER/T terminator create failures diagnosable

### LINKED ISSUE #4193: ER/T terminator create reports "invalid edge router for session" with no router-side error and slow recovery
When an edge router tunneler gains bind access to a service before the edge router policy and
service edge router policy that also govern hosting are in place, the terminator create is
rejected and the failure is close to undiagnosable.

## Symptoms

The controller logs:

```
{"error":"invalid edge router for session","file":".../create_tunnel_terminator_v2.go:86",
 "level":"error","msg":"responded with error","routerId":"...","terminatorId":"..."}
```

No terminator appears for the service, and the router logs nothing at all.

## Problems

1. **The error message is wrong.** There is no session in the ER/T hosting path; it was removed
   when ER/T was migrated to the router data model. `InvalidEdgeRouterForSessionError` is reused
   from the session-based path, so the message misdirects anyone diagnosing it toward session
   state. The same message is used on the sessionless `create_circuit_v3` path.

2. **The error is not actionable.** `IsAccessToEdgeRouterAllowed` checks two independent policy
   links (identity to edge router via edge router policy, service to edge router via service
   edge router policy) and collapses them to a single bool, so the denial can't say which
   policy is missing.

3. **The router logs nothing.** `HandleCreateTerminatorResponse` discards `response.Msg` and
   `response.ErrorCode` on failure. From the router's side hosting silently doesn't work, and
   the only trace anywhere is the single controller log line.

4. **Recovery is slow.** On a non-busy failure the terminator stays in `Establishing` and is not
   requeued, so the next attempt waits on the retry scan, which requires `lastAttempt` older
   than 2 minutes against a 1 minute ticker. Config applied in quick succession (a script or a
   paste of several `ziti edge create` commands) can lose the race and then appear broken for
   2-3 minutes.

## Reproduction

Create a service, add a bind service policy, then grant the router the hosting role attribute
*before* creating the edge router policy and service edge router policy. The router begins
hosting as soon as the bind policy grants access and immediately fires the create.

## Notes

Requiring the service edge router policy on the hosting side is intentional. The underlying
gap is that the ER/T has no view of edge router or service edge router policies, so it can't
know it is not yet in a valid hosting situation. That is expected to be resolved separately by
bringing the data into the router data model. This issue covers making the failure diagnosable
and narrowing the race in the meantime.

## Also found

Controller error codes never reached SDK clients on the dial paths. The router read only the
message body from the controller's error reply and dropped `ErrorCodeHeader`, and
`sendStateClosedReply` looked for the code on the SDK's *request* rather than the controller's
response. As a result `edge.ConnRefusalError` always classified dial refusals as
`CauseUnknown`, making the access-denied, invalid-session, and invalid-service classifications
unreachable.

plorenz:
Fixes #4193

An ER/T that gains bind access to a service before the edge router policy and service edge
router policy are in place has its terminator create rejected, and the failure is close to
undiagnosable: the controller reports "invalid edge router for session" on a path that has no
session, doesn't say which policy is missing, and the router logs nothing at all.

Requiring the service edge router policy on the hosting side is intentional. The real gap is
that the ER/T has no view of those policies, so it can't tell it isn't in a valid hosting
situation yet. That's expected to be addressed separately by bringing the data into the router
data model. This makes the failure diagnosable and narrows the race in the meantime:

- denials now name the missing policy instead of reusing a session error
- the router logs the controller's rejection instead of discarding the code and message
- a new terminator waits 2s before its first create, so config applied in quick succession
  settles first rather than failing and waiting out the 2-3 minute retry scan

Investigating the error path turned up a second defect, fixed here as well: controller error
codes never reached SDK clients on the dial paths. The router read only the message body from
the controller's error reply and dropped the error code, so `edge.ConnRefusalError` classified
every dial refusal as unknown, making the access-denied, invalid-session, and
invalid-service classifications unreachable. Both dial relays now pass the code and retry hint
through as a structured error.

The changelog's Component Updates and Bug Fixes section was regenerated with `ziti-ci
update-release-notes`. Three existing entries the commit and pull request scan cannot
rediscover are now marked with `<!-- keep -->` so regeneration stops dropping them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -484,6 +484,7 @@ today and expand as packages are converted. The global `ziti agent set-log-level
 * github.com/openziti/transport/v2: [v2.0.215 -> v2.0.216](https://github.com/openziti/transport/compare/v2.0.215...v2.0.216)
 * github.com/openziti/xweb/v3: [v3.0.4 -> v3.0.5](https://github.com/openziti/xweb/compare/v3.0.4...v3.0.5)
 * github.com/openziti/ziti/v2: [v2.0.0 -> v2.1.0](https://github.com/openziti/ziti/compare/v2.0.0...v2.1.0)
+    * [Issue #4193](https://github.com/openziti/ziti/issues/4193) - ER/T terminator create reports "invalid edge router for session" with no router-side error and slow recovery
     * [Issue #4149](https://github.com/openziti/ziti/issues/4149) - upgrading a running 1.x controller/router to 2.x fails to create the service user
     * [Issue #3990](https://github.com/openziti/ziti/issues/3990) - Expose service change subscriptions to external SDKs over protobuf
     * [Issue #4108](https://github.com/openziti/ziti/issues/4108) - Controller retains bbolt-managed memory past transaction (create-circuit response SIGSEGV)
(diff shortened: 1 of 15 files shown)


dovholuknf (approved):
lgtm

Issue #4141, opened 2026-07-22
Service-policy enforcer deletes valid legacy sessions; type= queries use numeric id against the string-mapped symbol


plorenz:
## Summary
Commit 9ac9a4f17 ("Filter service policies by type=\"Dial\"/\"Bind\" in queries", #3818) changed the service-policy `type` query symbol to a string (`ast.NodeTypeString` + `policyTypeSymbolMapper` mapping the stored int32 to `"Dial"`/`"Bind"`). Three queries were not updated and still filter with the numeric `PolicyType.Id()` (`type = 1`/`type = 2`), which can never match the string-mapped value, so they return zero rows.

## Impact
- **`controller/internal/policy/service_policy_enforcer.go` (Run)** — the startup/periodic scan checks each session's identity for a covering service policy with `type = <int>`, always gets 0, and **deletes every non-admin session** it scans. Only legacy durable edge sessions are stored/scanned (OIDC sessions are self-contained), so the damage is scoped to legacy-auth clients, and it recurs on **every controller start** (the enforcer is primed to scan at construction). Surfaced in the upgrade test: legacy clients taken through a standalone->HA conversion lose their sessions (two controller starts) and hit `invalid session` + orphaned SDK terminators.
- **`controller/internal/routes/service_router.go` (listIdentities)** and **`controller/internal/routes/identity_router.go` (listServices)** — `?policyType=Dial|Bind` filters build `type = %d` and return empty results.

## Fix
Query the mapped string form, e.g. `type = "Dial"` via `PolicyType.String()` (quoted), in all three spots. Add an enforcer regression test (valid Dial+Bind sessions survive a `Run()`; a session with no covering policy is deleted); 9ac9a4f17's test only covered the store-level string query syntax.

## Affected
main and release-v2.0.x. **Not** release-v1.6.x (its `type` symbol is still int; the numeric query is correct there).

PR #4200, opened 2026-07-30
Add in-place upgrade smoke test

### LINKED ISSUE #4199: Add in-place upgrade smoke test
## Summary

We lack automated coverage for in-place upgrades of a running OpenZiti cluster. Regressions in upgrade/downgrade behavior, HA cluster recovery, and steady-state traffic during version transitions can slip through existing unit and integration tests.

## Proposal

Add a fablab-based smoke test (`zititest/models/upgrade-test`) that:

- stands up a multi-region HA controller cluster plus routers and loop4 sim traffic clients (SDK, ERT, ZET, tunneler)
- drives repeatable upgrade/downgrade iterations against a running environment, re-bootstrapping in place each iteration
- exercises HA cluster disruption and recovery (snapshot restore, node rejoin) between iterations
- validates steady-state traffic and that expected terminators are present after each disruption

This gives us a long-running, iterating test that surfaces upgrade-path and HA-recovery regressions that shorter tests miss.

plorenz:
## Summary

Adds a fablab-based in-place upgrade smoke test that exercises upgrading a running OpenZiti cluster across versions, so upgrade-path and HA-recovery regressions are caught before release.

The test (`zititest/models/upgrade-test`) stands up a multi-region HA controller cluster, routers, and loop4 sim traffic clients (SDK, ERT, ZET, tunneler), then drives repeatable upgrade/downgrade iterations. Each iteration re-bootstraps in place, disrupts and recovers the HA cluster (snapshot restore, node rejoin), and validates steady-state traffic and expected terminators.

It also hardens the loop4 sim-control harness so the validation gate tolerates the client reconnection churn that in-place upgrades produce: run-scenario requests are re-sent to sims that reconnect mid-scenario, each scenario tracks its expected client set explicitly, a sim runs one scenario at a time and answers duplicate or conflicting requests instead of going silent, and superseded/errored sim connections are closed instead of leaked.

## Product code change

Beyond the test harness, this changes controller behavior:

- makes a clustered controller honor `restartSelfOnSnapshot` on the replicated db-restore path, not just the raft FSM restore
- moves `RestartController` to a package function so both restore paths share it

The setting is false unless `cluster:` is configured, so this only affects HA deployments that opted into it. An operator running with `restartSelfOnSnapshot: true` gets a controller that restarts itself after a replicated db restore, where it previously exited and waited for something external to restart it.

## Running it

`.github/workflows/upgrade-test.yml` adds a `workflow_dispatch` entry point, with optional inputs for the ZET version and the from/to versions. Blank inputs fall through to the model's own defaults. There is no scheduled or PR-triggered run; it is manual only.

## Known limitation

`toVersionRef` defaults to the `release-v2.0.x` branch rather than a release tag, because the two fixes this test depends on (#4125 and #4141) are on that branch but not yet in a published 2.0.x release. Until v2.0.3 ships, a run builds whatever the branch points at, so runs are not reproducible from the version numbers alone. Once v2.0.3 is out, `toVersion: v2.0.3` with an empty `toVersionRef` removes the source build entirely.

Closes #4199

### DIFF
--- a/doc/design/sdk-hosting-recovery-fix.md
+++ b/doc/design/sdk-hosting-recovery-fix.md
@@ -1,0 +1,132 @@
+# SDK Hosting Recovery Bug: Fix Plan
+
+## Summary
+
+Go SDK hosted services do not recover after their bind (service) session is invalidated. This is
+triggered by the controller upgrade from 1.6 to 2.0: the 2.0 controller deletes all pre-2.0 service
+sessions (a known, correct compatibility step, pre-2.0 sessions are not understood by 2.0), and the
+routers restart during the upgrade. After that, a Go SDK host that held its connection through the
+event never re-creates its bind session. It keeps re-binding the stale, now-deleted session, so
+terminator creation fails with "invalid session" forever, and that host's traffic stays dead until the
+process is restarted.
+
+The C SDK (ziti-edge-tunnel) and edge-router-tunnelers recover correctly in the same scenario. A full
+restart of the Go app also fixes it (a fresh listener manager creates a valid session). So the controller
+behavior is correct; the gap is in the Go SDK's hosting-recovery path.
+
+## Evidence (from the upgrade-test live run)
+
+- Controller: `create.terminator` -> `loadFromBolt: invalid session` (create_terminator_v2.go). The
+  router presents the SDK's bind-session token; the controller cannot find it in bolt because it is the
+  stale/deleted session.
+- Go SDK host log: `establishing bind session` = 0, `successfully created` = 0, `failure creating` = 0.
+  So `createSession` is never called, the SDK is not re-creating the session at all.
+- Go SDK host log: ~21k `timed out sending listener start-over event`, ~36k `invalid session`.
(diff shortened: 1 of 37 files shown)


doc/design/sdk-hosting-recovery-fix.md
```diff
@@ -0,0 +1,132 @@
+# SDK Hosting Recovery Bug: Fix Plan
```
andrewpmartinez: Did you mean to include this? It reads more like a bug fix than a design meant for external consumers.

doc/design/upgrade-test.md
```diff
@@ -0,0 +1,236 @@
+# Upgrade Test Design
```
andrewpmartinez: Same with this.

dovholuknf (approved):

doc/design/sdk-hosting-recovery-fix.md
plorenz: fixed

doc/design/upgrade-test.md
plorenz: fixed

PR #4205, opened 2026-07-30
Honor ZITI_LOG_NO_JSON and restore terminal-aware log-format default

### LINKED ISSUE #4204: ziti run/tunnel ignore PFXLOG_NO_JSON and lost the terminal-aware log-format default
## Problem

The slog-based logging used by `ziti run` and `ziti tunnel` picks its output format solely from the `--log-formatter` flag, defaulting to `pfxlog` (pretty) when unset. Two behaviors regressed versus the pfxlog `GlobalInit` path:

- **`PFXLOG_NO_JSON` is ignored.** pfxlog honored `PFXLOG_NO_JSON` to force non-JSON output; the new path never reads it. Test harnesses (`tests/cli_tests/cluster_test.go`, `tests/testutil/overlay.go`) and deployments set it and silently get no effect on the new path.
- **The default lost terminal awareness.** pfxlog defaulted to JSON when stderr was piped/redirected (production) and pretty on a terminal. The new path defaults to pretty always, so redirected/production output is no longer machine-parseable JSON by default.

## Fix

Restore pfxlog's terminal-aware default and honor the env override on the new path:

- when `--log-formatter` is unset: `ZITI_LOG_NO_JSON` (or the deprecated `PFXLOG_NO_JSON`) forces `pfxlog`; otherwise a terminal stderr gets `pfxlog` and a redirected stderr gets `json`
- an explicit `--log-formatter` value always wins
- introduce `ZITI_LOG_NO_JSON` as the current name and keep `PFXLOG_NO_JSON` honored as a deprecated alias (it also still works on the legacy pfxlog path, so it stays portable during the migration)

plorenz:
The slog-based logging on `ziti run` / `ziti tunnel` chose its output format only from `--log-formatter` (defaulting to pretty), which dropped two behaviors the old pfxlog `GlobalInit` path had: it ignored `PFXLOG_NO_JSON`, and it lost the terminal-aware default that emitted JSON when stderr was redirected (production) and pretty on a terminal.

This adds `logging.ResolveFormat`, used by both `ziti run` and `ziti tunnel`, to restore that behavior when `--log-formatter` is unset:

- `ZITI_LOG_NO_JSON` (or the deprecated `PFXLOG_NO_JSON`) forces `pfxlog` output
- otherwise a terminal stderr gets `pfxlog`, a redirected stderr gets `json`
- an explicit `--log-formatter` value always wins

`ZITI_LOG_NO_JSON` is the new name; `PFXLOG_NO_JSON` stays honored as a deprecated alias. Keeping the old name matters during the pfxlog migration: it's also understood natively by the legacy pfxlog path (e.g. `ziti edge quickstart`), so it stays portable across both, while `ZITI_LOG_NO_JSON` only affects the new path for now. The existing test harnesses that set `PFXLOG_NO_JSON` are left as-is for that reason.

Unit tests cover the precedence rules, both env var names, and the pfxlog-compatible boolean parsing. `go test ./common/logging/`, `go vet`, and builds of the two callers pass.

Fixes #4204

### DIFF
--- a/common/logging/resolve.go
+++ b/common/logging/resolve.go
@@ -25,7 +25,7 @@ import (
 	"golang.org/x/term"
 )
 
-// Environment variables that force non-JSON (pfxlog) output when the
+// Environment variables that force non-JSON (pretty) output when the
 // --log-formatter flag is unset. EnvNoJson is the current name; EnvNoJsonLegacy
 // is the pfxlog-era name, still honored so existing deployments and test
 // harnesses keep working.
@@ -38,11 +38,11 @@ const (
 // value and output destination.
 //
 // An explicit, non-empty flagValue always wins and is returned unchanged. When
-// the flag is unset, the format is chosen the way pfxlog historically chose it:
-// ZITI_LOG_NO_JSON (or the deprecated PFXLOG_NO_JSON) forces pfxlog (pretty)
-// output; otherwise a terminal out gets pfxlog output and a non-terminal
-// (piped or redirected) out gets JSON, so production processes default to the
-// machine-parseable shape.
+// the flag is unset, the format is chosen by terminal detection, the way ziti
+// has historically chosen it: ZITI_LOG_NO_JSON (or the deprecated
+// PFXLOG_NO_JSON) forces pretty output; otherwise a terminal out gets pretty
+// output and a non-terminal (piped or redirected) out gets JSON, so production
+// processes default to the machine-parseable shape.
 //
 // out is the writer the handler will target (os.Stderr for ziti binaries). A
 // nil out, or one whose fd is not a terminal, counts as non-terminal.
--- a/common/logging/resolve.go
+++ b/common/logging/resolve.go
@@ -1,0 +1,93 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


ziti/run/root.go
```diff
@@ -42,7 +42,7 @@ type Options struct {
 
 func (self *Options) BindFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().BoolVarP(&self.Verbose, "verbose", "v", false, "Enable verbose logging")
-	cmd.PersistentFlags().StringVar(&self.LogFormatter, "log-formatter", "", "Specify log formatter [json|pfxlog|text]")
+	cmd.PersistentFlags().StringVar(&self.LogFormatter, "log-formatter", "", "Specify log formatter [json|pfxlog|text]; default is pfxlog on a terminal and json when redirected (ZITI_LOG_NO_JSON forces pfxlog)")
```
dovholuknf: what about just 'pretty' instead of 'pfxlog'?

dovholuknf (approved):
small request wrt pretty vs pfxlog also i added https://github.com/openziti/ziti/issues/4255 - we should do the full work of converting this imo

ziti/run/root.go
plorenz: done, with pfxlog left as a hidden alias for backwards compatibility

common/logging/resolve.go
```diff
@@ -0,0 +1,93 @@
...
+		return FormatPretty
+	}
+	return FormatJSON
+}
+
+// noJsonEnv reports whether either NO_JSON environment variable is set to a
+// truthy value, preferring the current name over the deprecated one.
+func noJsonEnv() bool {
+	return envBool(EnvNoJson) || envBool(EnvNoJsonLegacy)
+}
```
andrewpmartinez, line 72: The godoc here is a lie. It doesn't have a preference. Additionally, the deprecated value is always considered, and the new value can never overcome it.

It looks like no tests set both.

Issue #4208, opened 2026-07-31
[Backport-1.6] Lock order inversion in ConnectionTracker deadlocks the controller


plorenz:
Backport of #4206 to `release-v1.6.x`.

`ConnectionTracker` in `controller/model/identity_manager.go` acquires the cmap shard
lock and the per-identity `identityConnections` lock in opposite orders in
`ScanForDisconnectedRouters` versus `MarkConnected`/`MarkDisconnected`. Racing on the
same identity deadlocks permanently, and the held shard lock cascades into leaked bbolt
read transactions and a fully wedged controller. See #4206 for the full analysis.

The 1.6 branch needs the patch adapted: module path is `github.com/openziti/ziti`
(no `/v2`), the tracker takes `channel.Channel` rather than `ctrlchan.CtrlChannel`,
and router ids come from `ch.Id()` rather than `ch.PeerId()`.

Note that on 1.6 the connect events handler spawns a goroutine per message instead of
using a bounded worker pool, so a controller in this state accumulates goroutines
considerably faster than on 2.0 and main.

plorenz:
Merged. Closing

Issue #4207, opened 2026-07-31
[Backport-2.0] Lock order inversion in ConnectionTracker deadlocks the controller


plorenz:
Backport of #4206 to `release-v2.0.x`.

`ConnectionTracker` in `controller/model/identity_manager.go` acquires the cmap shard
lock and the per-identity `identityConnections` lock in opposite orders in
`ScanForDisconnectedRouters` versus `MarkConnected`/`MarkDisconnected`. Racing on the
same identity deadlocks permanently, and the held shard lock cascades into leaked bbolt
read transactions and a fully wedged controller. See #4206 for the full analysis.

The 2.0 code is identical to `main` here (same `/v2` module path, same
`ctrlchan.CtrlChannel` signatures), so the fix applies unchanged.

PR #4209, opened 2026-07-31
Fix lock order inversion in ConnectionTracker

### LINKED ISSUE #4206: Lock order inversion in ConnectionTracker deadlocks the controller
## Summary

`ConnectionTracker` acquires the cmap shard lock and the per-identity
`identityConnections` lock in opposite orders in different code paths. When the two
paths race on the same identity, they deadlock permanently, and the held shard lock
then wedges the entire controller.

## Lock order inversion

`controller/model/identity_manager.go`

**`ScanForDisconnectedRouters` (1194-1203)** takes the value lock, then the shard lock:

```go
entry.Val.Lock()                        // holds identityConnections lock
if len(entry.Val.routers) == 0 {
    self.connections.RemoveCb(...)      // wants cmap shard lock
}
entry.Val.Unlock()
```

**`MarkConnected` (1210-1221) and `MarkDisconnected` (1245-1252)** take the shard lock,
then the value lock:

```go
self.connections.Upsert(identityId, nil, func(...) {   // holds cmap shard lock
    ...
    valueInMap.Lock()                                   // wants identityConnections lock
```

(`concurrent-map/v2` runs both the `Upsert` and `RemoveCb` callbacks with the shard lock
held: `concurrent_map.go:87-89` and `:161-163`.)

If the scan loop is reaping an identity with no remaining router connections at the same
moment a connect-events message for that identity is being processed, the two block on
each other forever.

Both conditions are routine. `SyncAllFromRouter` calls `MarkDisconnected` for every
identity not reported by the syncing router, and `MarkDisconnected` inserts an empty
entry for identities it has never seen, which the scan loop then tries to reap. On a
network with many routers doing periodic full-state syncs, entries are constantly being
created and reaped, so the window is hit regularly.

Secondary defect on the same lines: the `RemoveCb` callback reads `len(v.routers)`
without holding the value lock, which is a data race regardless of the deadlock.

## Impact

The shard lock is never released, so the failure escalates well past the tracker:

1. `Identity.fillFrom` calls `IdentityManager.HasErConnection` ->
   `ConnectionTracker.GetIdentityOnlineState` -> `cmap.Get`, which needs an `RLock` on
   the stuck shard. Every identity read blocks.
2. Those reads run inside `bbolt` `db.View` transactions, so each blocked reader leaks an
   open read transaction. Edge API requests start returning 503.
3. Once a write transaction needs to grow the database, `bbolt.(*DB).mmap` requires the
   exclusive `mmaplock` and cannot get it past the leaked read transactions. The raft FSM
   apply goroutine stalls there.
4. `beginTx` can no longer acquire `db.metalock`, so all remaining database access piles
   up behind it. The `bolt.read` health check starts reporting
   `previous bolt ping is still running` and the controller stops serving entirely.

In a goroutine dump of a controller in this state, essentially every goroutine is parked
on one of those two locks. In an HA cluster the affected member's clients fail over to
the remaining members, which can then hit command rate limits during the resulting
terminator re-registration burst.

## Fix

Establish a single lock order everywhere: **shard lock -> `identityConnections` lock**.
In `ScanForDisconnectedRouters`, do the emptiness pre-check under the value lock, release
it, and take the value lock inside the `RemoveCb` callback instead. The callback re-checks
the value currently in the map, which is the authoritative check.

Separately, reduce the entry churn that makes the race easy to hit, by having
`SyncAllFromRouter` only mark identities actually recorded against the syncing router
rather than iterating every key in the map.

## Affected

`main`, `release-v2.0.x`, `release-v1.6.x`. Backports to 2.0 and 1.6 to follow.

plorenz:
Fixes #4206

`ConnectionTracker` acquires the cmap shard lock and the per-identity
`identityConnections` lock in opposite orders. `ScanForDisconnectedRouters` holds the
value lock while calling `RemoveCb`, which needs the shard lock, whereas
`MarkConnected`/`MarkDisconnected` hold the shard lock inside the cmap `Upsert` callback
while taking the value lock. Racing on the same identity deadlocks permanently.

Because the shard lock is then never released, the failure escalates well past the
tracker: every identity read blocks in `GetIdentityOnlineState`, those reads run inside
bbolt read transactions, the leaked read transactions eventually block a database grow,
and the controller stops serving entirely. #4206 has the full chain.

This establishes a single lock order, shard lock then value lock, by releasing the value
lock before `RemoveCb` and taking it inside the callback instead. That also fixes an
unsynchronized read of the router map in the old callback.

The new `TestConnectionTrackerLockOrdering` deadlocks reliably without the fix and passes
with it.

Backports are tracked in #4207 (2.0) and #4208 (1.6). A follow-up will reduce the entry
churn in `SyncAllFromRouter` that makes this race easy to hit.

### DIFF
--- a/controller/model/identity_manager.go
+++ b/controller/model/identity_manager.go
@@ -1099,6 +1099,14 @@ const (
 	IdentityStateUnknown IdentityOnlineState = 2
 )
 
+// identityConnections tracks which routers an identity is currently connected to, along
+// with the last connectivity state reported for it.
+//
+// Two locks protect the connection tracker: the shard lock inside the ConcurrentMap of
+// these, and the lock on each of these. They must always be acquired in that order,
+// shard lock first. The cmap Upsert and RemoveCb callbacks run with the shard lock held,
+// so taking this lock inside one of them is fine; taking the shard lock while already
+// holding this one is not, and will deadlock.
 type identityConnections struct {
 	sync.RWMutex
 	routers           map[string]ctrlchan.CtrlChannel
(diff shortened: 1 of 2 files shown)


controller/model/identity_manager.go
```diff
@@ -1099,6 +1099,14 @@ const (
 	IdentityStateUnknown IdentityOnlineState = 2
 )
 
+// identityConnections tracks which routers an identity is currently connected to, along
+// with the last connectivity state reported for it.
+//
+// Two locks protect the connection tracker: the shard lock inside the ConcurrentMap of
+// these, and the lock on each of these. They must always be acquired in that order,
```
dovholuknf, line 1106: > They must always be acquired in that order

if this is the case these should be a function and somehow the actual mutex'es discouraged from being used (encapsulated if possible)?

plorenz: This is intended to be a minimal change. On main, I'm going to investigate a larger simplification along with expanding the associated chaos test and verifying performance.

dovholuknf (approved):

PR #4214, opened 2026-08-03
Auto-GC stale router links

### LINKED ISSUE #4005: Auto-GC stale router links
Part of #3743. Builds on the controller-managed link configuration work. Detect and optionally close router links that have gone stale relative to the current configuration.

Scope:
- A structured `xlink.LinkKey` (`{DialerBinding, Protocol, DestId, ListenerBinding}`) used by all staleness/GC checks.
- A `CheckStaleLinks` ctrl message plus a `ziti ops verify stale-links` CLI for operator-driven, two-sided staleness verification, with an optional `--gc` to close confirmed-stale links.
- Auto-GC via the `router.link.v1` `gcMode` (`preserve` / `orphaned` / `changed`): the router walks its xlinks after each Apply that mutates listeners, dialers, or gcMode and closes one-sided-stale entries under the configured mode.

plorenz:
Adds opt-in stale-link garbage collection for the router link subsystem, driven by a `gcMode` field on the `router.link` config (`preserve` (default) / `orphaned` / `changed`). After a link-config change, the router closes established links its current listeners/dialers can no longer support, according to the configured mode. Adds a `ziti ops verify stale-links` command for a controller-aggregated view.

Fixes #4005.

Stacked on `router-link-management` (#4084).


router/router.go
```diff
@@ -869,14 +869,36 @@ func (self *Router) currentLinkListeners() *ctrl_pb.Listeners {
...
+}
+
+// runLinkGcIfConfigured reads the current gcMode from the active link
+// config and, if non-preserve, walks the xlink registry closing any
+// stale entries. Safe to call on every subsystem-change notification:
+// preserve / no-config short-circuits before doing any work.
+func (self *Router) runLinkGcIfConfigured() {
+	cfg := self.linkSubsystem.GetConfig()
+	if cfg == nil {
+		return
```
andrewpmartinez, line 891: On `Remove` the config is already nil here and this will abort. Correct?

- The rescan that `Remove` triggers now only detaches the orphaned link states.
- An operator running `orphaned` or `changed` and then disassociating the config gets the `preserve` outcome, with dialed links left up and undialable.

controller/network/validate_stale_links.go
```diff
@@ -0,0 +1,355 @@
...
+		connected := n.GetConnectedRouter(r.Id)
+		if connected == nil {
+			unqueryable[r.Id] = fmt.Sprintf("router %s is not connected", r.Id)
+			continue
+		}
+		if blocker := staleLinkQueryBlocker(connected); blocker != "" {
+			unqueryable[r.Id] = blocker
+			continue
+		}
+		wg.Add(1)
```
andrewpmartinez, line 180: Should this cap the fan-out the way the other validate sweeps do? 

`ValidateLinks` and others hold a `concurrenz.NewSemaphore(10)`, and here every filtered router gets a goroutine and a minute-long timeout request at the same instant.

andrewpmartinez (approved):
Just the one question about the remove sweep being nuetered.

router/router.go
plorenz: Change it so that if you remove the config, all links are torn down. The felt like the most consistent thing to do

controller/network/validate_stale_links.go
plorenz: Added the fan-out cap

PR #4216, opened 2026-08-03
Remove unused link costTags support

plorenz:
Removes the unused `costTags` option from router link listeners. It was parsed from router config, advertised to the controller, and stored on the router model, but was never used for path selection or any other behavior. Drops `GetLinkCostTags` from the `xlink.Listener` interface and its transport implementation, reserves the corresponding `ctrl_pb.Listener.costTags` and `RouterLinks.RouterLink.linkCostTags` protobuf fields, and documents the removal in the changelog.

This is independent cleanup of a pre-existing feature and sits at the base of the router link-management stack, so the feature commits build on a costTags-free foundation.


dovholuknf (approved):
seems perfectly cromulent

PR #4218, opened 2026-08-04
Don't require a client cert on TLS layers that don't request one. Fixes #4202

### LINKED ISSUE #4202: wss edge listener rejects all clients on 2.0: certValidatingIdentity forces client-cert verification on a listener that intentionally sets NoClientCert
On ziti 2.0.x, an edge listener bound with address: wss: completes the outer TLS handshake and then immediately drops every connection with no client certificate presented. Because browsers cannot present client certificates, this breaks all browser-based wss access (BrowZer, browser/WASM SDK clients). The same router config works on 1.6.12.

The cause is router-side: router/state/cert_validating_identity.go (new in 2.0) attaches a VerifyConnection callback to every edge listener's TLS config, including the wss listener, which deliberately does not request a client certificate on the outer TLS.

### To Reproduce

Router config:

listeners:
  - binding: edge
    address: wss:0.0.0.0:443
    options:
      advertise: <router-public-name>:443

Then:

curl -sv --max-time 10 --resolve <router-public-name>:443:127.0.0.1 \
  -H "Connection: Upgrade" -H "Upgrade: websocket" \
  -H "Sec-WebSocket-Version: 13" -H "Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==" \
  "https://<router-public-name>/ws"

###Expected

HTTP/1.1 101 Switching Protocols. The outer TLS is server-auth only; the client certificate is presented in the inner TLS established over the websocket.

###Actual

TLS completes from the client's perspective, then the connection is closed before the request can be sent:

* ALPN: server accepted http/1.1
* SSL certificate verify ok.
* Send failure: Broken pipe
* OpenSSL SSL_write: Broken pipe, errno 32
* Failed sending HTTP request

A node ws client surfaces the alert directly: tlsv13 alert certificate required.

###Workaround

Give the wss listener its own identity, which bypasses the wrapper. wss.Config.Load replaces cfg.Identity with a freshly loaded plain identity.ID when a transport.wss.identity section is present:

transport:
  wss:
    identity:
      server_cert: certs/public-server.cer.pem
      server_key:  certs/public-server.key.pem
      ca:          certs/intermediate-chain.pem

All three keys are required (ValidateForServerWithPathContext demands server_cert, ca, and one of key/server_key). Confirmed working. Note it also opts wss connections out of the new router-data-model client-cert validation — the inner TLS still requires and verifies a client cert against the configured CA bundle, which is the 1.6.12 behaviour.

### AI thinks the root cause is
Root cause

Factory.CreateListener wraps the router identity for all edge listeners — router/xgress_edge/factory.go:176:

wrappedId := state.WrapIdentityWithCertValidation(factory.env.GetRouterId(), factory.stateManager)
return newListener(wrappedId, factory, options, headers), nil

The wrapper attaches VerifyConnection unconditionally — router/state/cert_validating_identity.go:53-57:

func (self *certValidatingIdentity) ServerTLSConfig() *tls.Config {
      cfg := self.Identity.ServerTLSConfig()
      cfg.VerifyConnection = self.verifyConnection
      return cfg
}

and the callback rejects a connection with no peer certificates — cert_validating_identity.go:62-68:

func (self *certValidatingIdentity) verifyConnection(state tls.ConnectionState) error {
      if len(state.PeerCertificates) == 0 {
              // No client cert presented. The edge listener requires a cert (RequireAnyClientCert),
              // so this should not happen. Reject to be safe.
              return errors.New("no client certificate presented")
      }

That comment states the faulty assumption. It holds for the tls: edge binding but not for wss:, where the transport listener deliberately disables client-cert requests on the outer TLS — transport/v2 wss/listener.go:146-148:

tlsConfig := cfg.Identity.ServerTLSConfig()
tlsConfig.ClientAuth = tls.NoClientCert
tlsConfig.NextProtos = append(tlsConfig.NextProtos, "h2", "http/1.1")

The client certificate for wss belongs to the inner TLS layer, established after the websocket upgrade — wss/listener.go:73-74:

cfg := listener.cfg.Identity.ServerTLSConfig()
cfg.ClientAuth = tls.RequireAndVerifyClientCert

So: outer handshake requests no certificate → completes with zero peer certificates → VerifyConnection fires → rejects → processConn logs handshake failed and closes. Setting ClientAuth = NoClientCert disables the request for a certificate but does nothing about a VerifyConnection callback that requires one.

Why this is a 2.0 regression

router/state/cert_validating_identity.go does not exist in 1.6.12. The transport-side wss code is unchanged across the two releases — wss/listener.go's outer-TLS block is byte-identical between transport/v2@v2.0.193 (1.6.12) and v2.0.215 (2.0.0), as is identity.ServerTLSConfig() between identity@v1.0.116 and v1.0.129. The behavioural change is entirely in the router.

Impact

Any router with a wss: edge listener on 2.0.x rejects every browser client. This silently breaks BrowZer and browser/WASM SDK deployments on upgrade from 1.6.x, with no configuration error reported — the listener logs a clean startup and only fails per-connection.

Suggested fix

Don't attach VerifyConnection to listeners that don't require a client certificate. Options:

1. Don't wrap the identity handed to a wss: binding.
2. Have verifyConnection return nil when there are no peer certificates and the listener's ClientAuth doesn't require one, deferring enforcement to the inner TLS layer.
3. Gate on the negotiated protocol — the wss listener registers ALPN h2/http/1.1, distinct from ziti-edge/ziti-link.

plorenz:
## What

`wss:` edge listeners on 2.0.x reject every connection right after the outer TLS handshake, breaking all browser-based access (BrowZer, browser/WASM SDK clients). The same config works on 1.6.x.

The router's cert-validating identity wrapper (new in 2.0) attaches a `VerifyConnection` callback to every edge listener's TLS config, and that callback fails any connection with no peer certificates. That assumption holds for the `tls:` binding, but the wss transport deliberately sets `NoClientCert` on its outer TLS layer and requires the client certificate on the inner TLS layer established over the websocket. Setting `NoClientCert` stops the server asking for a cert; it does nothing about a callback that demands one.

## Fix

Skip validation when the TLS layer's `ClientAuth` doesn't require a certificate, leaving enforcement to the layer that does. `ClientAuth` is read at verification time rather than when the config is built, because transport listeners set it after obtaining the config.

Net effect: the wss outer layer lets certless clients through to the websocket upgrade, the wss inner layer still requires and validates the client cert against the router data model CA pool, and `tls:` edge listeners are unchanged.

Verified end-to-end against the real `transport/v2` wss listener: a certless browser-style client now gets `101 Switching Protocols`, while the pre-fix code reproduces the reported broken pipe. Inner-layer enforcement confirmed still active.

### DIFF
--- a/router/state/cert_validating_identity.go
+++ b/router/state/cert_validating_identity.go
@@ -31,7 +31,8 @@ import (
 // certValidatingIdentity wraps an identity.Identity to add client certificate chain verification
 // at the TLS level via a VerifyConnection callback. It dynamically builds a CA pool from the
 // router data model's PublicKeys with ClientX509CertValidation usage, ensuring the latest
-// trust anchors are always used.
+// trust anchors are always used. Verification is skipped for TLS layers that don't require a
+// client certificate.
 type certValidatingIdentity struct {
 	identity.Identity
 	stateManager Manager
@@ -51,18 +52,38 @@ func WrapIdentityWithCertValidation(id *identity.TokenId, stateManager Manager) 
 	}
 }
 
+// ServerTLSConfig returns a tls.Config which validates client certificate chains against the
+// router data model CA pool, unless the config is later set to not require a client certificate.
 func (self *certValidatingIdentity) ServerTLSConfig() *tls.Config {
 	cfg := self.Identity.ServerTLSConfig()
-	cfg.VerifyConnection = self.verifyConnection
+	if cfg == nil {
+		return nil
+	}
+
+	// ClientAuth is read at verification time rather than here because transport listeners may
+	// change it after this call. The wss listener sets NoClientCert on the outer TLS layer and
+	// requires the client certificate on the inner TLS layer established over the websocket.
+	cfg.VerifyConnection = func(state tls.ConnectionState) error {
+		if len(state.PeerCertificates) == 0 && !requiresClientCert(cfg.ClientAuth) {
+			return nil
+		}
+		return self.verifyConnection(state)
+	}
 	return cfg
+}
+
+// requiresClientCert reports whether the given client auth type makes a handshake fail when the
+// client presents no certificate.
+func requiresClientCert(clientAuth tls.ClientAuthType) bool {
+	return clientAuth == tls.RequireAnyClientCert || clientAuth == tls.RequireAndVerifyClientCert
 }
 
 // verifyConnection is a TLS VerifyConnection callback that verifies client certificates against
 // the CA pool built from RDM PublicKeys.
 func (self *certValidatingIdentity) verifyConnection(state tls.ConnectionState) error {
 	if len(state.PeerCertificates) == 0 {
-		// No client cert presented. The edge listener requires a cert (RequireAnyClientCert),
-		// so this should not happen. Reject to be safe.
+		// The listener requires a client cert, so the handshake should not have gotten this far
+		// without one. Reject to be safe.
 		return errors.New("no client certificate presented")
 	}
 
--- a/router/state/cert_validating_identity_test.go
+++ b/router/state/cert_validating_identity_test.go
@@ -1,0 +1,92 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


dovholuknf:
this feels like it weakens the security posture of WSS and should have to be an intentional configuration change. @andrewpmartinez should definitely review this

andrewpmartinez:
Says broken in 2.0.x (LTS), but targets main, and I can't find a backport PR.

andrewpmartinez (approved):
Backport to 2.0.x needed.

plorenz:
> Backport to 2.0.x needed.

Backport is here: https://github.com/openziti/ziti/pull/4290

PR #4220, opened 2026-08-04
Update deps and changelog for 1.5.17 release

plorenz:


zititest/go.mod
andrewpmartinez, line 3: I'm seeing the patches for dep updates should floor this at  1.26.4. CI is 1.26.x.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# Release 1.5.17
+
+## What's New
+
+Update libraries and build with the latest Go version.
```
andrewpmartinez, line 5: Updates libs but not go version.

andrewpmartinez (approved):
Comments but not stoppers.

zititest/go.mod
plorenz: This just puts a floor on the go version. The actual version used is the latest 1.26. I double-check, the last released was built with 1.26.4, this one will build with 1.26.5

CHANGELOG.md
plorenz: see above

zititest/go.mod
andrewpmartinez: Seems odd for a security-oriented PR that we don't make sure we get the latest security patches from go.

It's nice to know that CI does the right thing. It would be even nicer to know it requires the right thing.

Issue #4234, opened 2026-08-07
Ensure terminator operations are scoped by source router


plorenz:
The fabric control-channel terminator handlers act on a message-supplied terminator id without checking that
the requesting router owns that terminator. A router can therefore operate on terminators belonging to other
routers.

Three handlers are affected on `main`:

- `remove_terminator.go` (singular) reads the terminator, then calls `Terminator.Delete` with no owner check.
- `remove_terminators.go` (batch) calls `Terminator.DeleteBatch` on the supplied ids with no owner check.
- `update_terminator.go` sets cost/precedence and calls `Terminator.Update` with no owner check.

Effects: a router can delete other routers' terminators, which takes down services hosted elsewhere. Via
update it can also set another router's terminator to `Failed` (effectively disabling it) or to
`Required`/`Cost=0`, which changes where the network sends traffic for that service rather than just denying
it.

The edge control channel already enforces this. `handler_edge_ctrl`'s `verifyTerminator` rejects a request
whose terminator's router does not match the source router, so this is a gap specific to the fabric handlers.

### Relationship to #4165

#4165 covered the remove path and is closed, but `main` is not actually protected:

- The 2.0.x backport (#4166 / #4168) landed on `release-v2.0.x` and covers the **batch** handler only. The
  singular remove handler and the update handler are untouched there.
- The main-line PR (#4167) merged into the feature branch `fix-sdk-terminator-congestion-collapse`, which is
  still unmerged behind an open PR stack (#4159 -> #4158). On that branch only the `lookupTerminatorOwner`
  helper is present, with no call site, so the fix is inert there too.

Net: no branch destined for `main` currently enforces ownership on any of the three handlers.

### Proposed fix

Apply the ownership check consistently across all three handlers, reusing the helper shape already on
`release-v2.0.x` (`selectOwnedTerminators` / `filterOwnedTerminators` in `base.go`) so the two lines converge:

- Remove paths: drop ids owned by a different router, keeping ids that are absent so a delete racing a
  not-yet-applied create still orders correctly through raft.
- Update path: reject outright when the terminator resolves to a different router.

Unit tests for the filtering helper and per-handler tests asserting a foreign-owned terminator is not
deleted, updated, or reported as success.

Issue #4166, opened 2026-07-24
[Backport-2.0] Fabric terminator remove handlers don't verify the terminator belongs to the requesting router


plorenz:
Backport of #4165 to release-v2.0.x.

release-v2.0.x has only the v1 fabric remove handler (RemoveTerminators); RemoveTerminatorsV2 does not exist on 2.0, so the backport adds the router-ownership check to the v1 handler only. The edge control channel already enforces ownership via verifyTerminator on 2.0.

PR #4238, opened 2026-08-07
[Backport-2.0] Scope terminator operations to the requesting router

### LINKED ISSUE #4236: [Backport-2.0] Ensure terminator operations are scoped by source router
Backport of #4234 to `release-v2.0.x`.

The batch remove handler is already scoped on this line via #4166, which added `selectOwnedTerminators` to
`base.go`. Two gaps remain, both reachable (all three handlers are registered in `bind.go:66-68`):

- `remove_terminator.go` (singular) deletes a terminator by id with no ownership check.
- `update_terminator.go` sets cost/precedence with no ownership check, so a router can set another router's
  terminator to `Failed`, or to `Required`/`Cost=0` to steer traffic.

Scope here is therefore smaller than on `main`: reuse the existing helper and add the two missing checks.

plorenz:
Backport of #4234 / #4235 to `release-v2.0.x`. Fixes #4236.

The batch remove handler was already scoped on this line by #4166, so this covers only the two handlers that
fix missed: the singular `remove_terminator.go` and `update_terminator.go`. Both were reachable, since all
three handlers are registered in `bind.go`.

Reuses the existing `ownsTerminator`-style check against the terminator each handler has already read, so
neither adds a second lookup. The helper is byte-identical to the one on `main`, so the two lines stay
convergent.

Also brings over the end-to-end test from #4235, which drives the fabric control channel from a real enrolled
router against a second router's terminator. Its batch case passes here on the strength of the existing
#4166 fix, so the test now covers all three paths on this line.

### DIFF
--- a/controller/handler_ctrl/base.go
+++ b/controller/handler_ctrl/base.go
@@ -34,6 +34,13 @@ func (self *baseHandler) newChangeContext(ch channel.Channel, method string) *ch
 	return change.NewControlChannelChange(self.router.Id, self.router.Name, method, ch)
 }
 
+// ownsTerminator reports whether terminator belongs to the router on the other end of this handler's
+// control channel. Terminator operations arriving on the fabric control channel are scoped to the
+// requesting router, mirroring the edge control channel's verifyTerminator.
+func (self *baseHandler) ownsTerminator(terminator *model.Terminator) bool {
+	return terminator.Router == self.router.Id
+}
+
 // lookupTerminatorOwner returns the id of the router that owns terminator id and whether the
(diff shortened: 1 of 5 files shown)


dovholuknf:
this PR is missing the change to remove_terminators.go that both the 1.6 and main PRs have. I didn't approve it based on that missing file?

plorenz:
> this PR is missing the change to remove_terminators.go that both the 1.6 and main PRs have. I didn't approve it based on that missing file?

That change already exists on this branch (fixed incidentally as part of another change), so it didn't need to be made in this PR.

dovholuknf (approved):

PR #4249, opened 2026-08-08
Index links by router id instead of scanning at connect time

### LINKED ISSUE #4246: Router connect scans the entire link table, making a reconnect wave scale with the cube of the router count
`BuildRouterLinks` runs once per router connect (`network.go:352`, inside `ConnectRouter`) and scans the
entire link table to find the links belonging to the connecting router:

```go
func (self *LinkManager) BuildRouterLinks(router *Router) {
	self.linkTable.links.IterCb(func(_ string, link *Link) {
		if link.DstId == router.Id {
			router.routerLinks.Add(link, link.Src.Id)
			link.Dst.Store(router)
		}
	})
}
```

The scan is O(links) and the link count in a full mesh is itself O(routers²), so a single connect is
O(routers²) and a wave in which every router reconnects is O(routers³) in link visits. In a 400-router mesh
that is ~79,800 links, so a full reconnect wave walks on the order of 32M links, each visit taking the
concurrent-map shard lock and being discarded for all but a handful of matches.

This is only reached on connect, so it does not show up in steady state. It shows up exactly when the
controller is least able to absorb it: a mass reconnect after a controller restart or a network event, when
every router is arriving at once and the connect path is already the bottleneck.

The scan exists for a real reason. A link references `Router` objects, and a reconnect replaces the object
while keeping the id, so anything caching object references goes stale. Ids do outlive the objects, which is
what makes an id-keyed index viable where an object-keyed one would not be.

### Fix

Index every link under both endpoint router ids in the link table, maintained in the add and remove paths
that all link mutations already funnel through, and have `BuildRouterLinks` walk only the connecting router's
links. That makes a connect proportional to the router's own link count rather than to the whole table.

Two things fall out of adding the index:

- Removal has to unindex only when the index still refers to the link being removed. Only removals matching a
  link's own iteration hold its id lock, so an unlocked removal can race a replacement that has already
  indexed itself, and removing by id alone erases the replacement. That is invisible in the link table, which
  still holds the replacement, and surfaces later as a link that `BuildRouterLinks` never finds.
- An index must not be dropped once created, even when empty, since discarding it could lose a link another
  goroutine had already fetched the index to record.

### Also here

A router announces its links once per reconnect and is not asked again, so a report the controller discards
leaves it with no links for a router that believes it announced them, with nothing on either side to re-offer
or re-ask. Discards are currently silent and neither side records how many links it sent or received, which
leaves nothing but inference when the two disagree. Logging the count on both sides of the exchange, and the
reason a report was discarded, makes that visible.

plorenz:
Fixes #4246

`BuildRouterLinks` runs on every router connect and scanned the whole link table to find the connecting
router's links. The table is O(routers²) in a full mesh, so a reconnect wave costs O(routers³) in link
visits. It never shows in steady state; it lands during a mass reconnect, when the connect path is already
the bottleneck.

This indexes every link under both endpoint router ids, so a connect walks only its own router's links. The
full-refresh prune was the same whole-table walk one message later, and now uses the index too.

Three commits: the index and its consequences, a small `LockedSet` it needs, and an unrelated test race
fixed along the way.

### What's worth a close look

**The index holds link ids, not link pointers.** A reconnect replaces a link's `Router` object and a higher
iteration replaces the `Link` object, but neither replaces the id. Per-router queries resolve ids against
the table, which stays authoritative.

**The table and the index are two structures written in sequence,** so `Add`, `Remove` and the connect-time
pairing now serialize on the link id. Previously only `RouterReportedLink` took that lock, and the
connect-time pairing was outside it entirely: a removal reading no destination could complete mid-pairing
and leave the link in the destination's link set after the table had dropped it.

**Pairing a link with its destination was broken two ways,** both leaving a link that reads as healthy
everywhere while carrying no adjacency. A router connecting between a report resolving its destination and
that link reaching the table was seen by neither path; and because the resolve happens under the source
router's connect stripe rather than the destination's, the destination could be replaced before the report
landed. The destination is now re-resolved unconditionally once the link is in the table, which works
because the connect path registers a router before building its links.

**Dropping a router's index on delete is guarded.** Fabric router ids are the enrollment certificate's
common name, so re-adding from the same cert reuses the id, and store event handlers run after bolt
releases the writer lock, so a delete's cleanup can land after the id was recreated and reported links. The
drop happens only while the store still has no router under that id, checked inside the index map's removal
callback — indexing a link takes that same shard, so a live router's entry can't appear without its create
having committed first.

### Also here

- Logs the link count on both sides of the reconnect exchange, and why a report was discarded. A router
  announces its links once per reconnect and is never asked again, so a discarded report left it with no
  links and nothing recorded to compare.
- Test routers get version info, which a connected router always has and the router sync path dereferences
  without a nil check.

### Tests

The index checked against the table for every router after a mesh is built, partly torn down, and has links
replaced by higher iterations; the prune's selection, including that a report from one end never removes
links the other end dialed; both destination-repair interleaves; the pairing serializing on the link id;
index removal in both directions and the concurrent removal/replacement case; and the delete cleanup
running after the same id has been recreated.

### DIFF
--- a/controller/model/link_manager.go
+++ b/controller/model/link_manager.go
@@ -28,7 +28,8 @@ import (
 	"github.com/openziti/ziti/v2/controller/models"
 	"github.com/openziti/ziti/v2/controller/storage/boltz"
 	"github.com/openziti/ziti/v2/controller/storage/objectz"
-	"github.com/orcaman/concurrent-map/v2"
+	cmap "github.com/orcaman/concurrent-map/v2"
+	"go.etcd.io/bbolt"
 )
 
 // linkLog is the logger for the controller's link manager. Its channel name is
@@ -40,6 +41,9 @@ type LinkManager struct {
 	linkLocks      *concurrency.StripedIdLocker
 	initialLatency time.Duration
 	models.BaseObjectStoreManager[*Link]
+	// env is held rather than the router manager, which does not exist yet when this one is built. Nil in
+	// tests that do not need router lookups.
+	env Env
 }
 
 func NewLinkManager(env Env) *LinkManager {
@@ -52,6 +56,7 @@ func NewLinkManager(env Env) *LinkManager {
 		linkTable:      newLinkTable(),
 		linkLocks:      concurrency.NewStripedIdLocker(256),
 		initialLatency: initialLatency,
+		env:            env,
 	}
 
 	result.InitStore(objectz.NewObjectStore[*Link](func() objectz.ObjectIterator[*Link] {
@@ -109,16 +114,41 @@ func (self *LinkManager) BaseLoad(id string) (*Link, error) {
 	return entity, nil
 }
 
+// BuildRouterLinks points every link destined for router at it, and adds each to router's link set. Safe
+// to call repeatedly; links already pointing at router are left alone.
 func (self *LinkManager) BuildRouterLinks(router *Router) {
-	self.linkTable.links.IterCb(func(_ string, link *Link) {
-		if link.DstId == router.Id {
-			router.routerLinks.Add(link, link.Src.Id)
-			link.Dst.Store(router)
-		}
-	})
+	for _, linkId := range self.linkTable.linkIdsForRouter(router.Id) {
+		self.buildRouterLink(router, linkId)
+	}
 }
 
+// buildRouterLink pairs one link with router, taking the link's stripe so the pairing cannot interleave
+// with a removal. Without it a removal that reads no destination completes while this is still deciding,
+// and the link lands in router's link set after the table has dropped it, where it is still routed over.
+func (self *LinkManager) buildRouterLink(router *Router, linkId string) {
+	defer self.linkLocks.LockFor(linkId)()
+
+	// Resolved under the stripe: a replacement may have taken this id since the index was read, and it is
+	// the link the table holds now that has to be paired.
+	link, found := self.Get(linkId)
+	if !found {
+		return
+	}
+	if link.DstId == router.Id && link.PointDestAt(router) {
+		router.routerLinks.Add(link, link.Src.Id)
+	}
+}
+
+// Add publishes link to the link table, to its endpoint routers' link sets, and to the per-router index.
+// Serialized against any other change to the same link id: the table and the index are written in separate
+// steps, and an interleaved removal would leave the link indexed but no longer in the table.
 func (self *LinkManager) Add(link *Link) {
+	defer self.linkLocks.LockFor(link.Id)()
+	self.addLocked(link)
+}
+
+// addLocked publishes a link. The caller must hold the link's stripe.
+func (self *LinkManager) addLocked(link *Link) {
 	self.linkTable.add(link)
 	link.Src.routerLinks.Add(link, link.DstId)
 	if dest := link.GetDest(); dest != nil {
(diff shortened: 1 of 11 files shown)


controller/model/link_manager.go
```diff
@@ -110,15 +114,30 @@ func (self *LinkManager) BaseLoad(id string) (*Link, error) {
 }
 
 func (self *LinkManager) BuildRouterLinks(router *Router) {
-	self.linkTable.links.IterCb(func(_ string, link *Link) {
-		if link.DstId == router.Id {
+	// Only this router's links, not the whole table: this runs for every connect, and in a full mesh a
+	// scan would cost the square of the router count.
+	for _, link := range self.linkTable.forRouter(router.Id) {
```
andrewpmartinez: Your workflows keep littering every file with comments like this, which are questionably useful.

Inline comments are only useful when they explain something non-intuitive or security-sensitive.

controller/model/link_manager.go
```diff
@@ -143,6 +162,38 @@ func (self *LinkManager) ScanForDeadLinks() {
...
+// router connecting between those two moments is seen by neither: the resolve found it absent, and the repair
+// did not yet find the link.
+//
+// Re-reading here closes that window rather than narrowing it, but only because Network.ConnectRouter
+// registers the router before building its links. A router whose build ran too early is therefore already in
+// the connected map by the time the link lands in the table. Reverse that ordering and this repair can also
+// miss, leaving the link with no destination and nothing left to notice.
+//
+// Left unrepaired the link is skipped by ConnectedNeighborsOfRouter, so it carries no adjacency and cannot be
+// routed over, while every operator-facing view still reports it healthy.
```
andrewpmartinez: This is specific and also not useful at the same time. The godoc being generated is ...so broad.

Consider:

```
// repairDest points link at whichever Router is currently connected for its destination id, and adds it to
// that router's link set. For a link recorded while that router was not connected. No-op if the router is
// still absent or the link already points at it, so it is safe to call repeatedly. The link must already be
// in the link table.
```

controller/model/link_manager.go
```diff
@@ -151,6 +202,11 @@ func (self *LinkManager) RouterReportedLink(reportedLink *ctrl_pb.RouterLinks_Ro
 
 	link, _ := self.Get(reportedLink.Id)
 	if link != nil && link.Iteration >= reportedLink.Iteration {
+		// The destination is repaired here though, since a link already holding none is one that both paths
+		// that set it have missed, and a later report is the only thing left that will notice.
```
andrewpmartinez: Adds no value.

controller/model/link_manager.go
```diff
@@ -272,14 +343,69 @@ func (self *LinkManager) LinksInMode(mode LinkMode) []*Link {
 
 type linkTable struct {
 	links cmap.ConcurrentMap[string, *Link]
+	// byRouter indexes every link under both of its endpoint router ids, so per-router work does not have
+	// to scan the whole table. A router's Link objects outlive the Router objects that reference it, and
+	// a reconnect replaces the object but never the id, which is what makes an id-keyed index stable.
```
andrewpmartinez: Added detail that, while true for this work, it might not be true for other words and can be determined by looking at the code.

```
//byRouter indexes router id -> link for fast lookups
```

andrewpmartinez (changes requested):
- other places can use the new map (reconnect, outisde of line changeso on this pr)
- godoc needs another pass, consider stripping most of the obvious inline comments
- a number of related near code issues send in dms

controller/model/link_manager.go
plorenz: fixed

controller/model/link_manager.go
plorenz: fixed

controller/model/link_manager.go
plorenz: fixed

controller/model/link_manager.go
plorenz: fixed

plorenz:
> * other places can use the new map (reconnect, outisde of line changeso on this pr)
> 
>     * godoc needs another pass, consider stripping most of the obvious inline comments
> 
>     * a number of related near code issues send in dms

Fixed all the review issues and then ran it through a couple more review cycles locally as well.

andrewpmartinez (approved):

Issue #4248, opened 2026-08-08
Router startup check samples the controller count at a single instant and kills a router that is momentarily between control channels


plorenz:
The router's startup check is a one-shot timer that samples the current controller count when it lands:

```go
pfxlog.Logger().Fatal("unable to connect to any controllers before timeout")
```

The count is an instantaneous reading. A router that connected successfully but happens to be between
control channels at that exact moment is killed as having failed to start. Under control channel churn,
briefly holding no registration is a routine state rather than a failure.

Seen in testing: a router held registrations to two controllers until 25ms before the check fired, made no
connection attempts at all during the timeout window because it had no reason to, and was terminated anyway.

Raising `startupTimeout` does not help. It moves the instant being sampled, it does not stop the check from
depending on a single instant.

### Fix

Track whether a control channel was ever established, set when one is and never cleared, and fail startup on
that rather than on the current controller count. That preserves the check's intent, since a router that can
never reach a controller has never connected and still exits, while a router that has connected at least once
is not killed for a momentary gap.

Log what the check saw either way, including whether a controller was ever reached and the count at the time,
so the outcome is recorded rather than inferred.

Issue #4247, opened 2026-08-08
Router never closes a control channel whose controller stops answering heartbeats, leaving recovery to a ~15 minute TCP timeout


plorenz:
The controller closes a control channel when the router stops answering heartbeats. The router does not do
the same for the controller. On the router side `unresponsive` is only ever a selection signal: it
deprioritizes a controller in `isMoreResponsive` and nothing tears the channel down.

That leaves no recovery for a channel whose underlays are dead but not yet detected. The channel object is
still open, so the router keeps its channel group and its connection id, and the redial path dials only
*additional* underlays for that existing group rather than establishing a new one. Once the controller has
torn down its side of the group, it refuses an underlay that does not claim `IsFirstGroupConnection`,
because there is no group left to join. Neither side will build a new group while the router believes it
still has one.

Nothing breaks the deadlock except the operating system eventually abandoning the dead TCP connection, which
on default Linux settings takes on the order of fifteen minutes. For that whole window the router holds a
registration for a controller it cannot exchange anything with, and because a present registration reads as
connected, it does not redial.

### Fix

Close the channel from the router's heartbeat check once no heartbeat response has arrived within
`closeUnresponsiveTimeout`, matching what the controller already does. Closing runs the normal close path, so
the registration is given up and the reconnect logic re-establishes the group.

Being unresponsive should stay purely a selection signal for the merely-slow case, so a controller that is
answering late is deprioritized rather than disconnected, and a zero timeout disables the teardown. The
last-response time needs seeding when the channel is created, so a channel is not judged before its first
heartbeat could have been answered.

Issue #4196, opened 2026-07-29
Router control-channel connect/disconnect race can leave a reconnected router de-registered


plorenz:
### Summary

Under control-channel churn, a stale or racing disconnect for a router can clobber a newer connection for
the same router: the old (dead) channel's late-detected disconnect de-registers the router that the newer
connection just registered and tears down its links. The router's data plane stays up, but the controller
no longer tracks it as connected, and its links go missing until anti-entropy/re-report recovers them.

The re-report half of that recovery is itself unreliable, so the two compound: the race loses the
controller's view of a router's links, and the mechanism meant to restore it can silently fail to deliver.

### Details

Connect and disconnect for a given router are not serialized, and only the `connected` map clear
(`MarkDisconnected`) is currency-guarded; the link teardown and presence-handler fan-out in
`DisconnectRouter` run unconditionally against whichever `*Router` instance the close handler captured.
Channel v5 gives each reconnect a distinct connection id, so a reconnecting router is a genuinely new
connection that races the old one's teardown — turning what used to be a coupled reconnect into a
connect-vs-disconnect race.

#### The reconnect announcement

`linkRegistryImpl.NotifyOfReconnect` announces a router's full link set once per controller reconnect, and
nothing re-offers or re-asks. That announcement is also the only message the controller treats as
authoritative, so it is the only thing that prunes: a controller that misses it can neither route over the
router's links nor drop the ones it should no longer hold, and neither side can tell, because the links stay
healthy on both ends.

The announcement was sent with no deadline while holding the registry lock, so it waited for a send-queue
slot indefinitely, stalling link accept and dial-succeeded for as long as the wait. Bounding it with a
deadline alone is not safe: `Send` returns once the message is queued and the deadline stays live, so the tx
loop discards it if the queue drains too slowly, and a plain send's listener ignores the error. The
announcement would be reported as sent, and the links marked synchronized, for a message that never left the
router.

Clearing the marks that record links as announced, and letting the periodic notification re-offer them, does
not substitute for delivering the announcement. That path sends without `FullRefresh`, so the controller
cannot prune, and it sends nothing at all when there are no established links, which is exactly when stale
controller state most needs clearing.

### Fix

- Serialize a router's connect and disconnect with a per-router striped lock.
- Keep at most one connection per router in the connected map: a connect into an occupied slot is rejected
  (its bind fails, so no rx starts and it is not registered) and the occupant is kicked closed; the router
  redials into the freed slot.
[...]

PR #4265, opened 2026-08-14
[Backport-2.0] Router control-channel connect/disconnect race can leave a reconnected router de-registered

### LINKED ISSUE #4264: [Backport-2.0] Router control-channel connect/disconnect race can leave a reconnected router de-registered
Backport of #4196 (PR #4197, merged as 40580f6de) to `release-v2.0.x`.

### Why

The controller-side connect path on `release-v2.0.x` is the same shape the fix was written against: `handler_ctrl/accept.go` loads the router with `GetReloadedRouter(ch.Id())` and calls `ConnectRouter(r)`, with no serialization between a router's connect and its disconnect, and only the connected-map clear guarded by connection currency. The link teardown and presence-handler fan-out in `DisconnectRouter` run unconditionally against whichever instance the close handler captured, so a stale or superseded disconnect can de-register a router that a newer connection just registered.

### Scope

PR #4197 carries seven commits and closes three issues. All of them are control-channel connect correctness and belong together:

- Serialize a router's connect and disconnect with a per-router striped lock (#4196)
- Keep at most one connection per router in the connected map (#4196)
- Close a control channel whose controller stops answering heartbeats (#4247)
- Fail router startup only if no controller was ever reached (#4248)
- Retry the reconnect link announcement and wait for the wire rather than the queue (#4196)
- Emit `ControllerReconnected` on an underlay-flap reconnect, which had no caller at all (#4196)

### This is a port, not a cherry-pick

`release-v2.0.x` is on `channel/v4` (v4.3.11); `main` is on `channel/v5` (v5.0.22), and the two APIs differ structurally rather than cosmetically. Found while trial-applying the series:

| v5 (main) | v4 (2.0.x) |
|---|---|
| `ch.GetSenders().(ctrlchan.CtrlChannel)` | `ch.(channel.MultiChannel).GetUnderlayHandler().(ctrlchan.CtrlChannel)` |
| `CtrlChannel.InitChannel(ch)` | framework calls `ChannelCreated(ch)`; no explicit call needed |
| `newBindHandler(...).BindChannel(binding)` | `binding.Bind(newBindHandler(...))` |
| `MultiChannelFactory` returns `(Channel, error)` | returns `(MultiChannel, error)` |
| `channel.NewChannel(*channel.Config)` | `channel.NewChannel(logicalName, factory, bindHandler, options)` — no `Config` struct at all |

Consequences for the port:

- `common/concurrency/striped_id_locker.go` does not exist on `release-v2.0.x` and has to come across (it is standalone, `sync` only).
- The `TestReject_BindErrorNotEstablishedOrRxStarted` integration test in `common/ctrlchan/channel_test.go` is written against v5's `channel.Config` API, which v4 does not have. It cannot be ported mechanically; it needs rewriting against v4's positional `NewChannel`, or dropping from the backport.
- Test doubles that stand in for a control channel have to implement `channel.MultiChannel` rather than supply `GetSenders`, and have to be wired via `ChannelCreated` to reproduce what the framework does on v4.

### Worth deciding before the PR lands

#4196's write-up attributes the race's frequency to v5: each reconnect gets a distinct connection id, so a reconnecting router is a genuinely new connection racing the old one's teardown. On v4 the reconnect is coupled, so the same race is reachable but not triggered by every reconnect. The fix is still correct on 2.0.x and the reported field symptoms (routers holding links the controllers did not know about) came from a 2.0.0 deployment, but the risk/benefit of each of the seven commits is worth weighing individually rather than assuming the whole series is as load-bearing there as on main.

plorenz:
Backport of #4197 to `release-v2.0.x`. Fixes #4264.

Carries all seven commits, each with a `cherry picked from` trailer. Together they serialize a router's control-channel connect and disconnect, keep one connection per router, and close the gaps that left a controller's view of a router's links wrong after churn.

- Serialize a router's connect and disconnect with a per-router striped lock (#4196)
- Keep at most one connection per router in the connected map (#4196)
- Close a control channel whose controller stops answering heartbeats (#4247)
- Fail router startup only if no controller was ever reached (#4248)
- Retry the reconnect link announcement, waiting for the wire rather than the queue (#4196)
- Emit `ControllerReconnected` on an underlay-flap reconnect, which had no caller at all (#4196)

### Ported rather than cherry-picked

`release-v2.0.x` is on `channel/v4`; `main` is on `channel/v5`, and the APIs differ structurally. Every adaptation below was needed to make the series build and pass here, and each is a behaviour-preserving translation rather than a change of intent:

| v5 (main) | v4 (this branch) |
|---|---|
| `ch.GetSenders().(ctrlchan.CtrlChannel)` | `ch.(channel.MultiChannel).GetUnderlayHandler().(ctrlchan.CtrlChannel)` |
| `CtrlChannel.InitChannel(ch)` | dropped; v4 wires the channel through `ChannelCreated` |
| `channel.Config` + `channel.NewChannel(cfg)` | `channel.MultiChannelConfig` + `channel.NewMultiChannel(cfg)` |
| `channel.NewClassicListenerWithAcceptor(...)` | `channel.NewClassicListenerF(..., listener.AcceptUnderlay)` |
| `MultiChannelFactory` returns `(Channel, error)` | returns `(MultiChannel, error)` |

Also carried over: `common/concurrency/striped_id_locker.go`, which does not exist on this branch and the serialization depends on. It is standalone (`sync` only) and comes with its tests.

No test coverage was dropped. The two tests that could not apply mechanically were ported and then checked against their own regressions:

- `TestReject_BindErrorNotEstablishedOrRxStarted` (`common/ctrlchan`) rebuilt on v4's multi-channel API; verified it fails when the bind handler stops rejecting.
- The connect-path test doubles now implement `channel.MultiChannel` and are wired via `ChannelCreated`, so they reproduce what the framework does on v4 rather than what v5 does.

### Verification

Build, vet, and the full `go test ./...` suite pass. `-race` is clean on `router/env`, `router/link`, `controller/model`, `controller/network` and `common/ctrlchan`, and the quickstart suite passes. Each fix was also re-checked against its own regression by reverting the fix and confirming the test fails.

### Worth a reviewer's eye

#4196 attributes the race's frequency to v5 giving each reconnect a distinct connection id, which makes a reconnecting router a genuinely new connection racing the old one's teardown. On v4 the reconnect is coupled, so the same race is reachable but not provoked by every reconnect. The controller-side connect path here is the shape the fix was written against (`GetReloadedRouter` then `ConnectRouter`, unserialized), and the field reports that prompted this work came from a 2.0.0 deployment, so the series applies. It is still worth deciding whether all seven commits carry their weight on this branch or whether a subset is preferred.


router/env/ctrl.go
```diff
@@ -149,6 +153,25 @@ func (self *networkCtrl) CheckHeartBeat() {
 	} else {
 		self.unresponsive.Store(false)
 	}
+
+	// Unresponsive alone only deprioritizes. A channel whose underlays are dead but undetected keeps its
+	// group and dials only additional underlays, which the controller refuses once its side is gone, so
+	// nothing re-establishes the group until the OS abandons the connection minutes later.
+	if timeout := self.heartbeatOptions.CloseUnresponsiveTimeout; timeout > 0 && self.timeSinceLastResponse() > timeout {
```
andrewpmartinez, line 161: Is this configuration actually usable?

On this branch a configured `closeUnresponsiveTimeout` never reaches this field. channel/v4's `LoadHeartbeatOptions` assigns it to `CheckInterval` instead, so the zero-disables case cannot actually be set.


Setting `0s` leaves the heartbeat pulse calling `time.NewTicker(0)`, which panics?

andrewpmartinez (approved):
I'd verify the configuration comment.

router/env/ctrl.go
plorenz: I fixed the channel library and pulled in the update here. Also added some config validation

Issue #4184, opened 2026-07-28
Router leaks LinkSendBuffer goroutines in `drainDeadlines()` — circuits accumulate until the router OOMs


msbusk:
## Summary

On a private edge router carrying Zabbix monitoring traffic, goroutines and memory grow monotonically from boot until the process exhausts host memory (~1 day). A goroutine profile shows ~25,000 goroutines parked in `LinkSendBuffer.drainDeadlines()`, waiting on `buffer.x.closeNotify` — i.e. the send buffer has closed, but its parent `Xgress` never does.

The growth is **independent of connection volume**: after we reduced new connections from 45/sec to 1.1/sec, the accumulation rate was unchanged.

## Environment

| | |
|---|---|
| Router | `openziti/ziti-router:2.0.0` (Docker, `network_mode: host`) |
| SDK | `sdk-golang v1.7.0` (per stack traces) |
| Controller | HA cluster, 2 endpoints |
| Router role | private — `link.dialers` only, no link listeners, no router roles |
| Listeners | `edge` (tls:0.0.0.0:3022) + `tunnel` (`mode: tproxy`) |
| Host | Debian 13, 4 vCPU, 7.9 GB RAM |
| Workload | Zabbix 7.4.11, ~53 monitored hosts |

## Symptom

Goroutines and RSS grow monotonically from container start and never plateau:

```
uptime      goroutines     memory
13 min          10,625      312 MB
~2.5 h          49,735     1.07 GB
~5 h            52,707     2.71 GB
```

Memory growth is ~5–7 MB/min ≈ 8–10 GB/day. The router is eventually OOM-killed or becomes unresponsive, monitoring fails, a restart resets the cycle, and it repeats.

## Where the goroutines are parked

`ziti agent stack`, grouped by creating function (25,922 xgress goroutines at the time of capture):

```
  24585  github.com/openziti/sdk-golang/xgress.(*Xgress).Start
   1254  github.com/openziti/sdk-golang/xgress.(*LinkSendBuffer).run
     67  github.com/openziti/sdk-golang/xgress.(*Xgress).acceptPayload

  24975  [select]
    944  [runnable]
```

A representative parked goroutine:

```
goroutine 271098 [select, 17 minutes]:
github.com/openziti/sdk-golang/xgress.(*LinkSendBuffer).drainDeadlines(0x3f1b2d4362a0)
	github.com/openziti/sdk-golang@v1.7.0/xgress/link_send_buffer.go:351 +0xda
[...]

PR #4274, opened 2026-08-17
Check that circuits drain in the fablab circuit test

plorenz:
`ValidateCircuits` checks that the controllers, forwarders and SDKs agree about the circuits that exist. A circuit that never completes its close handshake is consistent everywhere, so it passes that check while still holding its xgress instances, buffers and goroutines for as long as it survives. That is the gap that let #4184 go unnoticed.

Adds `ValidateCircuitsDrained`, which polls every controller until no circuit remains on the services under test, and reports the survivors grouped by service and sorted by age when they don't. It runs at the end of each `circuit-test` iteration, after the sim has finished and settled.

A few notes on the shape:

* It polls rather than asserting immediately. Teardown is asynchronous, and `checkForClose` has its own five-second-empty rule with `MaxCloseWait` at 30s, so a hard zero the moment traffic stops would flake for reasons unrelated to any bug. A healthy run costs one API call.
* Every controller is queried, since in an HA cluster each one owns the circuits it created.
* Results are paged. A leak can produce thousands of circuits, and the default page size would report only the first handful and understate it badly.
* Age is the reported signal, since it is what distinguishes a stuck circuit from one still in flight.

`circuit-test` excludes `sim-control` and `metrics`, which carry the model's own control and telemetry traffic and stay up for the life of the run.

Also picks up the latest fablab, for component version pinning.

Validated against a live run: it correctly failed on the model's own infrastructure circuits before those were excluded, and passes now.

For #4184


zititest/zitilab/validations/circuits.go
```diff
@@ -148,3 +153,151 @@ func validateCircuitsOnce(id string, clients *zitirest.Clients, routerFilter str
...
+	}
+
+	return result, nil
+}
+
+// listCircuits pages through all circuits known to a controller. The page size is well above
+// what a drained network should return, but a leak can produce thousands, and reporting only
+// the first page would understate it.
+func listCircuits(clients *zitirest.Clients) (rest_model.CircuitList, error) {
+	ctx, cancelF := context.WithTimeout(context.Background(), 30*time.Second)
```
andrewpmartinez, line 279: What happens if we go past 30s? Under heavy load, this looks like it might be possible.

I assume it just ...stops? Never reporting errors or returning the surviving circuits.

zititest/zitilab/validations/circuits.go
```diff
@@ -148,3 +153,151 @@ func validateCircuitsOnce(id string, clients *zitirest.Clients, routerFilter str
...
+			break
+		}
+		logger.Infof("\tcircuit %s not drained: service=%s age=%v", c.id, c.service, c.age.Round(time.Second))
+	}
+}
+
+func listUndrainedCircuits(run model.Run, ctrls []*model.Component, includeService func(string) bool) ([]*circuitSummary, error) {
+	var result []*circuitSummary
+
+	for _, ctrl := range ctrls {
```
andrewpmartinez, line 240: Is there a reason not to do this in parallel rather than sequentially?

andrewpmartinez (approved):
In the PR body it says "one API call" but I see multiple controllers being contacted. Not sure what "one API call" means in that case.

PR #4279, opened 2026-08-18
Fix two flaky OIDC integration tests

plorenz:
Two OIDC integration tests fail intermittently on CI for unrelated reasons. Both are fixed here.

## `Test_Authenticate_OIDC_Auth`

The test asserts that the sdk/env info posted during OIDC login is persisted on the identity, reading the identity back immediately after the token exchange. The controller applies that update through the background command queue, which runs the work inline only while recent background writes have been fast and hands it to a worker pool otherwise. On a loaded CI runner the read wins the race and the assertion sees empty sdk info, which is what turns unrelated pull requests red.

Both assertion sites now poll until the update lands, replacing an immediate read on the cert path and a fixed one second sleep on the updb path. The test is sound under a forced-async queue and finishes faster than before.

## `Test_Revocation_IdentityCutoff_PostCutoffSessionSurvives_OIDC`

The test creates an identity revocation and then authenticates a fresh session, asserting that a session issued after the cutoff survives. A token's `iat` claim carries whole seconds, while the revocation cutoff keeps full precision, so enforcement cannot distinguish a session issued just before the cutoff from one issued just after it within the same second and rejects both. The test passed only when the reaper's latency happened to push the re-authentication into the next second.

It now waits until a whole second after the revocation before authenticating, so the fresh token's issued-at is unambiguously later. The wait returns immediately when the clock has already moved on, which is the usual case, so it costs nothing outside the race.

Confirmed by authenticating immediately after the revocation, which reproduces the CI failure exactly, and again after the wait, which succeeds.

### DIFF
--- a/tests/api_client_management.go
+++ b/tests/api_client_management.go
@@ -198,6 +198,21 @@ func (helper *ManagementHelperClient) ListIdentitiesByFilter(filter string) ([]*
 	return resp.Payload.Data, nil
 }
 
+// RequireIdentitySdkInfoUpdated polls until the identity reports appId as its SDK app id, returning the
+// identity as read and failing the test if it does not appear within 10 seconds. The controller applies
+// sdk/env info updates on a background queue, so they are not guaranteed to be visible by the time
+// authentication returns.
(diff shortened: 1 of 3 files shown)


andrewpmartinez (approved):
Provided a comment or two on the backport version. https://github.com/openziti/ziti/pull/4280

PR #4284, opened 2026-08-20
Replicate admin MFA removal across HA cluster members

### LINKED ISSUE #4291: Admin MFA removal doesn't replicate across HA cluster members
Admin MFA removal (DELETE /identities/{id}/mfa) deletes the row with a direct bolt write, so on an HA cluster only the serving controller drops the enrollment and the other members keep challenging the identity for a TOTP code indefinitely. 
DeleteAllForIdentity bypasses raft and goes right to bbolt.
The removal needs a raft command carrying the identity id, so the lookup and delete run inside the log apply on every member and a lagging follower can't act on a stale view.

CLBRITTON2:
- Admin MFA removal is now a raft command carrying the identity id, so the delete applies on every cluster member
- Closes #4291


plorenz (approved):
Code changes look good. Can you do two things?
1. Add an issue and reference it from one of the commits
2. Either squash when you merge or rebase and squash before you merge

PR #4288, opened 2026-08-21
Bound ziti ops verify traffic server wait so a failed dial can't hang

plorenz:
## What

Fixes three defects in `ziti ops verify traffic` that together turned a momentary fabric race into an open-ended CI hang.

- **Hang:** in `both` mode the context had no timeout and `doClient` only cancelled the server on the success path, so a failed dial left the server blocked in `listener.Accept()` forever (`mode=server` used a bare context with no cancel at all). The server's Accept now has a `--max-wait` bound (default 30s) and `doClient` tears the server down on every exit path.
- **False pass:** `doBoth`/`doBothExtJwt` logged goroutine errors but returned nil, so a failed traffic test exited 0 and callers (e.g. a `retry` wrapper) never retried. The client error now propagates so the command exits non-zero.
- The command exits within a bounded time in all modes instead of hanging.

## Why

In a fresh-install CI run, verify-traffic hit a transient terminator-propagation race: `waitForTerminator` confirmed the terminator on the controller (control-plane), but the immediate dial routed through a router whose data-plane terminator table hadn't caught up yet ("service has no terminators"). That is expected eventual-consistency behavior and clears on retry, but the tool had no timeout and no non-zero exit, so instead of failing fast and being retried it hung for over an hour.

### DIFF
--- a/ziti/cmd/ops/verify/ops_verify_traffic.go
+++ b/ziti/cmd/ops/verify/ops_verify_traffic.go
@@ -530,19 +530,28 @@ func (t *traffic) doBoth() error {
 	wg := &sync.WaitGroup{}
 	wg.Add(2)
 	ctx, cancel := context.WithCancel(context.Background())
+	var serverErr, clientErr error
 	go func() {
 		defer wg.Done()
-		if err := t.doServer(ctx, false); err != nil {
-			log.Error(err)
+		if serverErr = t.doServer(ctx, false); serverErr != nil {
+			log.Error(serverErr)
 		}
 	}()
 	go func() {
 		defer wg.Done()
-		if err := t.doClient(cancel); err != nil {
-			log.Error(err)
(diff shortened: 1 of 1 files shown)


ziti/cmd/ops/verify/ops_verify_traffic.go
```diff
@@ -56,6 +56,7 @@ type traffic struct {
 	allowMultipleServers bool
 	extJwtSigner         string
 	redirectURL          string
+	maxWait              time.Duration
```
dovholuknf: i think we can just drop maxWait entirely. it should 'work' or 'fail' quickly. i think the OTHER changes in here will let that happen?

plorenz: change made :+1:

dovholuknf (approved):

Issue #4304, opened 2026-08-25
Support build time provided capabilities in the /version capabilities list


andrewpmartinez:
Add a list of capabilities to the client and management API `/version` response, because
downstream builds of `ziti` need to advertise capabilities that are build time provided.

- Accept a build time provided set of capability names, empty in a stock build.
- Merge them into the capabilities reported by `/version` on both the client and management APIs.
- Report the same names from `/enumerated-capabilities`.
- Accept any well-formed name without validating it against a known set.
- Show the capabilities under `ziti version -v`.

PR #4305, opened 2026-08-25
fixes openziti/ziti#4304 build time provided capabilities

andrewpmartinez:
- adds common/build.buildCapabilities, a linker -X target holding a comma-separated list of capability names, empty in a stock build
- parses it into well-formed [A-Z0-9_] names, dropping blanks, duplicates, and anything past the 32-name and 64-character caps, without validating against a known set
- merges the injected names into /version and /enumerated-capabilities on both the client and management APIs, deduped against the names the controller reports for itself
- prints a Capabilities line under ziti version -v when the list is non-empty

### DIFF
--- a/controller/internal/routes/version_router.go
+++ b/controller/internal/routes/version_router.go
@@ -105,6 +105,7 @@ func (ir *VersionRouter) buildVersions(ae *env.AppEnv) *rest_model.Version {
 		Version:        buildInfo.Version(),
 		APIVersions:    map[string]map[string]rest_model.APIVersion{},
 		Capabilities:   []string{},
+		BuildFlags:     build.GetBuildFlags(),
 	}
 
 	for apiBinding, apiVersionToPathMap := range webapis.AllApiBindingVersions {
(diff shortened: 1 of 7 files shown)


ekoby (approved):

controller/internal/routes/version_router.go
```diff
@@ -165,18 +165,46 @@ func (ir *VersionRouter) buildVersions(ae *env.AppEnv) *rest_model.Version {
 		v.Capabilities = append(v.Capabilities, string(rest_model.CapabilitiesHACONTROLLER))
 	}
 
+	v.Capabilities = mergeCapabilities(v.Capabilities, build.GetBuildCapabilities())
```
ekoby: this breaks API contract: Capabilities are enums in spec. This change will certainly break some generated clients

ekoby (changes requested):

controller/internal/routes/version_router.go
andrewpmartinez: Moved to `buildFlags`

ekoby (approved):

PR #4307, opened 2026-08-26
align router checks for posture with controller logic

dovholuknf:


router/posture/process.go
```diff
@@ -70,7 +70,7 @@ func (p *ProcessCheck) requireAll(data *InstanceData) *CheckError {
 			}
 		}
 
-		failedValue := p.compareProcesses(osType, cacheProcess, process)
+		failedValue := p.compareProcesses(cacheProcess, process)
```
andrewpmartinez, line 73: It's been a minute, but I am pretty sure this is incorrect. PROCESS (deprecated) and MULTI_PROCESS checks have an OS filter. Ignoring the OS filter isn't the answer.

If the SDK doesn't submit OS info, that is an SDK bug.

If the legacy posture checks weren't checking the OS field, the legacy check has a bug.

andrewpmartinez (changes requested):

router/posture/os.go
```diff
@@ -43,6 +43,10 @@ func (m *OsCheck) Evaluate(data *InstanceData) *CheckError {
 		}
 	}
 
+	if len(foundOs.OsVersions) == 0 {
+		return nil
+	}
+
```
andrewpmartinez, line 49: I swear this fix it also in another PR.

dovholuknf: it was on a different and failed pr of mine https://github.com/openziti/ziti/pull/3981/changes#diff-2c52d42a9170f44225f7f018d733c729ea1f984e6cc7eb24fec17b48ee915082

PR #4308, opened 2026-08-27
serve the overlapping-kid ext-jwt test's jwks endpoints over loopback…

andrewpmartinez:
… http

- serves the test's two jwks providers over plain http instead of https, the JWKS resolver uses the external system pool for external chain validation.
- removes the newTlsJwksServer helper and the http.DefaultTransport CA injection, which the resolver's own transport never consulted


dovholuknf (approved):

andrewpmartinez:
Fixed by https://github.com/openziti/ziti/pull/4333

PR #4331, opened 2026-09-02
Close ER/T dial circuits when the router loses dial access

### LINKED ISSUE #4330: ER/T does not close existing circuits when the router loses dial access to a service
Removing an edge router tunneler's identity from a dial service policy stops new intercepts for the service, but circuits already established through the intercept keep flowing until the application or the router ends them. Restarting the router is currently the only way to drop them.

Two related gaps in the same path:

* The router's pre-dial policy check for tunneler circuits was miswritten (`err != nil && policy != nil`, a state the check never returns) and so never denied a dial.
* The controller did not verify the dial service policy when a router requested a tunneler circuit (`CreateTunnelCircuitV2`). It verified only edge router and service edge router policy.

The router-embedded SDK (`xgress_sdk`) has the same untracked circuits and the same miswritten check.

Fixing the router check also means posture checks on policies that grant router identities are now evaluated. Routers submit no posture data, so such a policy denies the router. The release notes need a warning telling admins to split routers out of posture-gated policies before upgrading.

Applies to 2.0.x. Present since the router data model based tunneler was introduced in 1.x.

plorenz:
Fixes #4330

When an edge router tunneler loses dial access to a service it is intercepting, the router only stopped
intercepting; circuits already established kept flowing until the application or the router ended them.
This tracks each tunneler (and embedded-SDK) dial circuit by service and closes them when access is lost,
whether through a service policy change or the identity being disabled or deleted.

Two related gaps are closed in the same path. The router's pre-dial policy check used a condition that
could never deny, so it is corrected. The controller now verifies the dial service policy for tunneler
circuit requests, which it previously did not check at all.

Correcting the router check means posture checks on policies granting router identities are now
evaluated, and since routers submit no posture data, such a policy denies the router. The release notes
carry a warning to split routers out of posture-gated policies before upgrading.

Tests cover circuit close with legacy and OIDC hosts, controller denial without a dial policy, and
denial by posture check. All three original tests fail on the previous code.

### DIFF
--- a/router/xgress_common/dial_circuits.go
+++ b/router/xgress_common/dial_circuits.go
@@ -1,0 +1,127 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package xgress_common
+
+import (
+	"github.com/michaelquigley/pfxlog"
+	"github.com/openziti/sdk-golang/v2/xgress"
+	cmap "github.com/orcaman/concurrent-map/v2"
+)
+
+// DialCircuitRegistry tracks the initiating xgress of every dial circuit a router-local
+// client (an edge router tunneler or an embedded SDK) has established, keyed by circuit id,
+// so they can be closed when the router's identity loses dial access to a service.
+// Safe for concurrent use.
+type DialCircuitRegistry struct {
+	circuits cmap.ConcurrentMap[string, *dialCircuit]
+}
+
+type dialCircuit struct {
+	serviceId string
+	x         *xgress.Xgress
+}
+
+// NewDialCircuitRegistry returns an empty registry. Register a dial circuit with PrepareTrack
+// followed by Publish, and close tracked circuits with CloseForService, CloseAll, or RevalidateAll.
+func NewDialCircuitRegistry() *DialCircuitRegistry {
+	return &DialCircuitRegistry{
+		circuits: cmap.New[*dialCircuit](),
+	}
+}
+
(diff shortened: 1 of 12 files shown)


router/xgress_common/dial_circuits.go
```diff
@@ -0,0 +1,127 @@
...
+			toClose = append(toClose, closing{circuit: c, reason: reason})
+		}
+	})
+
+	for _, entry := range toClose {
+		pfxlog.Logger().WithField("circuitId", entry.circuit.x.CircuitId()).
+			WithField("serviceId", entry.circuit.serviceId).
+			WithField("reason", entry.reason).
+			Info("closing dial circuit")
+		entry.circuit.x.Close()
```
andrewpmartinez, line 125: This close runs on the data model subscriber goroutine.

`Xgress.Close` retries the circuit end as long as the destination keeps timing out. Could a congested circuit block this?

router/xgress_edge_tunnel/tunneler.go
```diff
@@ -154,28 +161,40 @@ func (self *tunneler) NotifyIdentityEvent(state *common.IdentityState, eventType
...
-func (self *tunneler) NotifyServiceChange(state *common.IdentityState, _, service *common.IdentityService, eventType common.ServiceEventType) {
+func (self *tunneler) NotifyServiceChange(state *common.IdentityState, previousService, service *common.IdentityService, eventType common.ServiceEventType) {
 	pfxlog.Logger().Infof("service changed for %s. service %s was %s", state.Identity.Name, service.Service.Name, eventType)
 	tunSvc := self.mapRdmServiceToRest(service)
 	switch eventType {
 	case common.ServiceAccessGainedEvent:
 		self.serviceListener.HandleServicesChange(ziti.ServiceAdded, tunSvc)
 	case common.ServiceUpdatedEvent:
 		self.serviceListener.HandleServicesChange(ziti.ServiceChanged, tunSvc)
+		if previousService != nil && previousService.IsDialAllowed() && !service.IsDialAllowed() {
```
andrewpmartinez, line 192: A posture check that is already in the identity's check set through another policy, say a bind policy, produces no `IdentityPostureChecksUpdatedEvent` when it is added to the dial policy granting this router, so nothing closes the circuit.

- `DialAllowed` stays true, so `Equals` reports no change and no `ServiceUpdatedEvent` fires.
- `ServiceDialPoliciesChanged` is the event that does fire

andrewpmartinez (changes requested):
Worried about the `Close()` blocking.

Issue #4338, opened 2026-09-03
Router does not relay an SDK xgress terminator's half-close to a legacy edge client


plorenz:
## Summary

When a service is hosted by an SDK app that binds with SDK flow control, the terminator is an
xgress conn and the router bridges it to the dialing edge client. If the hosted server
half-closes, the terminator sends a native xgress EOF, but the router never relays that to the
client as an edge FIN. The client's read never ends: every byte arrives and the stream stays
open until the client's own timeout fires.

Any protocol that signals end-of-transfer by closing the connection is affected. FTP
passive-mode transfers are the obvious case, since the transfer completes and the control
channel returns `226 Transfer complete` while the data connection never ends, so the client
reports a timeout on a transfer that actually succeeded. Length-delimited protocols such as
HTTP never notice.

The circuit only stalls while the client holds its own write side open, which is exactly the
shape of a download. If the client half-closes first, the host's close is delivered normally.

## Affected versions

Reproduced on 1.6.12, 1.6.20, 2.0.3 and 2.0.4, with three client SDKs and two independent
target servers. `main` is not affected: `edgeXgressConn.FlowFromFabricToXgressClosed` in
`router/xgress_edge/fabric.go` arrived with connect-v2 (#3884) and does the relay. Both LTS
lines lack it, so there is no released version to upgrade to.

`xgress.SignalConnection` and its call site already exist in the sdk-golang versions those
lines pin (v1.6.0 and v1.7.0). The router simply does not implement the interface, so the type
assertion fails silently. No SDK change or dependency bump is needed.

This is not the ER/T half-close problem from #3261 / #3526 / #3781: that fix shipped in 1.6.16
and 1.6.20 still fails. A service hosted on an edge router (terminator binding `tunnel`) is
also unaffected; it is specifically the SDK-hosted xgress terminator that stalls.

## Reproducing

A protocol-free probe is enough: host a TCP service through `ziti tunnel host`, have the target
write N bytes and close, and have the client read to EOF without writing. All N bytes arrive
and EOF never does.

`ziti tunnel` opts into an xgress terminator by default, since `--default-connections` defaults
to 2 and more than one default connection requires SDK flow control. An SDK app calling a bare
`ctx.Listen()` gets a legacy terminator and is not affected, which is why this reproduces
through the tunneler but not through a minimal SDK listener.

## Workaround

Run the tunneler with `--default-connections 1`, which gives a legacy terminator and delivers
[...]

PR #4341, opened 2026-09-03
[Backport-2.0] Relay an xgress terminator's half-close to edge clients

### LINKED ISSUE #4339: [Backport-2.0] Router does not relay an SDK xgress terminator's half-close to a legacy edge client
Backport of #4338 to release-v2.0.x.

Verified failing on the released 2.0.x line and fixed by the port. See #4338 for the
mechanism and affected versions.

plorenz:
Backports the router-side half-close relay to this line. Fixes #4339, part of #4338.

A service hosted by an SDK app that binds with SDK flow control gets an xgress terminator,
which the router bridges to the dialing edge client. When the hosted server half-closed, the
terminator's native xgress EOF was never relayed to that client as an edge FIN, so the
client's read never ended: all bytes arrived and the stream stayed open until the client's
own timeout fired. Downloads over FTP and anything else that signals end-of-transfer by
closing were reported as failures despite completing.

`main` already does this, via `FlowFromFabricToXgressClosed` added with connect-v2. Both LTS
lines lack it, so there was no released version to upgrade to. The interface and its call
site already exist in the sdk-golang version this line pins, so this is router-only with no
dependency change.

Includes a regression test that binds the host with SDK flow control so the router bridges an
xgress terminator to an edge client. It fails on this branch without the fix and passes with
it. Verified end to end against a released tunneler on this line, both with the protocol-free
case and with a real FTP passive transfer.


router/xgress_edge/fabric.go
```diff
@@ -492,6 +493,31 @@ func (self *edgeXgressConn) close(notify bool, reason string) {
 	}
 }
 
+// FlowFromFabricToXgressClosed implements xgress.SignalConnection. It signals the
+// end of the fabric->app half of the circuit to the edge client as an edge FIN, so
+// a blocked Read there returns io.EOF while the app->fabric half stays open and the
+// client can still write. The FIN is sent at most once; a later full close still
+// sends StateClosed and tears down both halves. It is a no-op once the conn or its
+// channel is closed.
+func (self *edgeXgressConn) FlowFromFabricToXgressClosed() {
```
andrewpmartinez, line 502: Does a tunneled client hit the same stall? The conn built in `router/xgress_edge_tunnel/fabric.go` is the initiator peer for an intercepted flow, and it doesn't implement `SignalConnection` either.

Its only write-side close path is the FIN payload flag in `XgressConn.WritePayload`, which an xgress terminator never sends.

andrewpmartinez (approved):
One question.

router/xgress_edge/fabric.go
plorenz: I'm going to create a separate issue for that and try for an integration test that tests all the different xgress conn types against each other. The problem exists on main, so I'll create the fixes there and then backport them.

Issue #4352, opened 2026-09-03
[Backport-1.6] Policy enforcers leak a metrics reference on every run


plorenz:
Backport of #4350 to `release-v1.6.x`.

The policy enforcers resolve a meter inside `Run`, and `Registry.Meter` takes a reference on every call, so each run leaks one. See #4350 for the mechanism and the affected call sites.

PR #4353, opened 2026-09-03
Hold the policy enforcers' meters rather than resolving one per run

### LINKED ISSUE #4350: Policy enforcers leak a metrics reference on every run
The policy enforcers resolve a meter on every run rather than holding one, which leaks a metrics registry reference each time and leaves the meter running after the registry is disposed.

`Registry.Meter` is reference counted: every call takes a reference, including calls that find the meter already present. Only `Dispose` releases one, and it tears the meter down only once the last is gone. So a lookup on a repeated path accumulates references it never sheds, and `DisposeAll` then decrements instead of stopping the meter — it leaves the registry, but its sampling keeps going.

Four call sites, three files:

- `controller/internal/policy/revocation_enforcer.go:75`
- `controller/internal/policy/service_policy_enforcer.go:112`
- `controller/internal/policy/service_policy_enforcer.go:177`
- `controller/internal/policy/api_session_enforcer.go:111`

All four are `GetMetricsRegistry().Meter(...).Mark(...)`. The fix is to resolve each meter once where the enforcer is built and hold it. The `Timer` lookups alongside them (`service_policy_enforcer.go:73`, `:126`) are safe — `Timer` is not reference counted — but read better cached in the same place.

These run per enforcement cycle rather than per event, so the growth is one reference per run rather than one per message. Unbounded over a controller's lifetime, but slow, which is why this is not urgent.

The asymmetry is easy to miss because nothing at the call site shows it: `registry.Timer(x).UpdateSince(t)` and `registry.Meter(x).Mark(1)` read identically and only one is safe. openziti/metrics#79 documents that on the library side.

Affects `release-v2.0.x` and `release-v1.6.x`; both have all three files. Not checking older branches, which are not LTS.

plorenz:
Fixes #4350

`Registry.Meter` is reference counted: every call takes a reference, including calls that find the meter already present, and only `Dispose` releases one. The policy enforcers resolved a meter inside `Run`, so each run took a reference it never shed. Disposing such a meter decrements the count instead of stopping it, so it leaves the registry and keeps sampling.

Four call sites across the three enforcers. Each meter is now resolved once where the enforcer is built and held.

The timers alongside them are held too, even though `Timer` is not reference counted — otherwise a reader is left working out why one of two adjacent lookups was cached. That asymmetry is the whole reason this was easy to write: `registry.Timer(x).UpdateSince(t)` and `registry.Meter(x).Mark(1)` read identically and only one is safe. openziti/metrics#79 documents it on the library side.

These run per enforcement cycle rather than per event, so the growth was one reference per run — unbounded over a controller's lifetime, but slow.

`grep -rnE '\.(Meter|Histogram)\([^)]*\)\.(Mark|Update)'` finds this shape; there are no others left outside tests.

The session enforcer test builds its enforcer directly, to get a negative session timeout that `NewSessionEnforcer` refuses, so it now supplies the metrics itself.

Backports: #4351 (2.0), #4352 (1.6).

### DIFF
--- a/controller/internal/policy/api_session_enforcer.go
+++ b/controller/internal/policy/api_session_enforcer.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/michaelquigley/pfxlog"
+	"github.com/openziti/metrics"
 	"github.com/openziti/ziti/v2/common/runner"
 	"github.com/openziti/ziti/v2/controller/change"
 	"github.com/openziti/ziti/v2/controller/env"
@@ -39,6 +40,11 @@ const (
 type ApiSessionEnforcer struct {
 	appEnv         model.Env
 	sessionTimeout time.Duration
+	// Resolved once and held. A Meter lookup takes a reference on every call and only Dispose releases
+	// one, so looking one up per run accumulates references and leaves the meter sampling after the
+	// registry is disposed. Timer is not reference counted, and is held here so the two read alike.
+	runTimer    metrics.Timer
+	deleteMeter metrics.Meter
 	*runner.BaseOperation
 }
 
@@ -55,6 +61,8 @@ func NewSessionEnforcer(appEnv *env.AppEnv, frequency time.Duration, sessionTime
 	return &ApiSessionEnforcer{
 		appEnv:         appEnv,
 		sessionTimeout: sessionTimeout,
+		runTimer:       appEnv.GetMetricsRegistry().Timer(ApiSessionEnforcerRun),
+		deleteMeter:    appEnv.GetMetricsRegistry().Meter(ApiSessionEnforcerDelete),
 		BaseOperation:  runner.NewBaseOperation("ApiSessionEnforcer", frequency),
 	}
 }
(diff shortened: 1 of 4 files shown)


controller/internal/policy/api_session_enforcer.go
```diff
@@ -108,7 +116,7 @@ func (s *ApiSessionEnforcer) Run() error {
 					logrus.WithError(err).Errorf("failure while deleting expired api session: %v", id)
 				}
 			}
-			s.appEnv.GetMetricsRegistry().Meter(ApiSessionEnforcerDelete).Mark(int64(len(ids)))
+			s.deleteMeter.Mark(int64(len(ids)))
```
andrewpmartinez, line 119: Is this mark in an `err != nil` intentional?

The delete meter only marks inside the failure branch, so a successful batch records nothing, and `api.session.enforcer.delete` reads zero while api sessions are being expired.

andrewpmartinez (approved):
One question about a `Mark` in an error branch.

controller/internal/policy/api_session_enforcer.go
andrewpmartinez: The backport doesn't seem to have this, too. Odd.

plorenz: I'll create a separate issue/pr to fix this, since it's unrelated to the PR

PR #4355, opened 2026-09-03
[Backport-1.6] Hold the policy enforcers' meters rather than resolving one per run

### LINKED ISSUE #4352: [Backport-1.6] Policy enforcers leak a metrics reference on every run

plorenz:
Fixes #4352

Backport of #4353 to `release-v1.6.x`.

The cherry-pick conflicted on imports only: this branch uses unversioned `github.com/openziti/ziti/...` paths where main uses `/v2`. Resolved by keeping the branch's own paths and adding `github.com/openziti/metrics`, which `go.mod` here already requires at v1.4.5. No `/v2` import remains in the changed files.

The first CI run failed `Test_RevocationSkipThreshold/immediate refresh creates revocation (TTL > threshold)` at `tests/revocation_test.go:430`. That is a timing flake unrelated to this change, and a re-run passed. The test configures a 15s refresh token against a 14s skip threshold, so it has about a second of slack between authenticating and refreshing; on a slow runner the remaining TTL drops below the threshold and the controller correctly skips the revocation, which is the nil the assertion trips on. Nothing in this change touches revocation creation — it caches metric handles in the enforcer that deletes expired revocations.

Worth noting that test exists only on this branch: main and 2.0 have `RevocationMinTokenLifetime` but no coverage for it, so there is no fixed version to backport, and the flake wants fixing here.

Also carries the fix for #4380 as a separate commit (backport of #4381), since that fix lands on main and 2.0 independently and this PR is the open 1.6 vehicle: the api session enforcer delete meter now marks on every delete attempt rather than only when the batch delete fails.

### DIFF
--- a/controller/internal/policy/api_session_enforcer.go
+++ b/controller/internal/policy/api_session_enforcer.go
@@ -18,13 +18,15 @@ package policy
 
 import (
 	"fmt"
+	"time"
+
 	"github.com/michaelquigley/pfxlog"
+	"github.com/openziti/metrics"
 	"github.com/openziti/ziti/common/runner"
+	"github.com/openziti/ziti/controller/change"
 	"github.com/openziti/ziti/controller/env"
 	"github.com/openziti/ziti/controller/model"
-	"github.com/openziti/ziti/controller/change"
 	"github.com/sirupsen/logrus"
-	"time"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


controller/internal/policy/api_session_enforcer.go
```diff
@@ -107,7 +116,7 @@ func (s *ApiSessionEnforcer) Run() error {
 					logrus.WithError(err).Errorf("failure while deleting expired api session: %v", id)
 				}
 			}
-			s.appEnv.GetMetricsRegistry().Meter(ApiSessionEnforcerDelete).Mark(int64(len(ids)))
+			s.deleteMeter.Mark(int64(len(ids)))
```
andrewpmartinez: This mark only runs when `err != nil` (error on `DeleteBatch`), so `api.session.enforcer.delete` stays at zero on a controller where the batch deletes succeed.

andrewpmartinez (changes requested):

PR #4356, opened 2026-09-04
Evict posture data for disconnected api sessions

### LINKED ISSUE #4323: Router posture cache grows without bound as clients re-authenticate
Applies to router-side posture enforcement (HA / v2.x).

The router's posture cache (`router/posture/checks.go`) keys its data by api session id and never removes anything:

- `apiSessionInstances` has no removal path at all. Every api session that ever sends posture data (or is seeded from its token at connect) leaves an `Instance` behind for the life of the router process. Because a full re-authentication mints a new api session id, ordinary client restarts and token-lifecycle churn grow the map indefinitely.
- `apiSessionInstanceHistory` retains up to 51 `InstanceData` snapshots per api session, each holding the OS, domain, MAC, and process-list values. It has never had a reader: it is written by `saveHistory` and read nowhere in the tree, so it is pure retained memory.

Impact is proportional to client churn rather than to steady-state client count, so it is least visible in short-lived test environments and most visible on long-running routers with many endpoints. Routers that are never restarted accumulate indefinitely.

Related: #4314 (posture response handling). Separate concern, same subsystem.

plorenz:
Fixes #4323

The router's posture cache never removed anything, so it grew for the life of the process: every re-authentication produces a new api session id, leaving behind an instance that nothing could reclaim. It also retained up to 51 posture snapshots per api session in a history that has never had a reader.

The connection tracker now tells the posture cache when an api session gains its first connection or loses its last, and the cache records the disconnect time. A sweep on the existing connection tracker loop evicts sessions whose recorded disconnect is older than the retention window, and the history is removed.

Eviction is deliberately deferred rather than immediate at channel close: the SDK only reports posture for services it is actively dialing or binding, so a client that reconnects before dialing anything may send nothing, and dropping its data at close would leave it with none. A reconnect, or posture data arriving, clears the recorded disconnect and cancels a pending eviction.

The same sweep reconciles recorded state against the tracker's connected set, but only corrects it and never evicts, so a missed connect or disconnect notification delays an eviction instead of causing or preventing one.

A reconnect and a removal are decided under the same lock, and the removal confirms the exact disconnect time its verdict was formed against, so a verdict overtaken by a reconnect cannot delete a live session's posture data.

### DIFF
--- a/router/posture/checks.go
+++ b/router/posture/checks.go
@@ -16,15 +16,14 @@ import (
 )
 
 type Cache struct {
-	apiSessionInstances       cmap.ConcurrentMap[string, *Instance]
-	apiSessionInstanceHistory cmap.ConcurrentMap[string, []*InstanceData]
-	updateListeners           []func(data *InstanceData)
-	totpParser                TotpTokenParser
+	apiSessionInstances cmap.ConcurrentMap[string, *Instance]
+	updateListeners     []func(data *InstanceData)
+	totpParser          TotpTokenParser
 }
 
-// NewCache creates a new posture data cache for managing device state information
-// across API sessions. The cache maintains both current posture data and historical
-// snapshots to support policy evaluation and audit requirements.
+// NewCache creates a new posture data cache holding the current posture data reported by each
+// API session, for evaluating posture checks. Data is retained until the session's posture
+// instance is evicted; see EvictInactive.
 //
 // Parameters:
 //   - parser: A TOTP token parser implementation, used to verify ToTP tokens on posture response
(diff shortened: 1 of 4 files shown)


router/posture/checks.go
```diff
@@ -7,24 +7,23 @@ import (
...
 	"github.com/openziti/ziti/v2/common"
 	"github.com/openziti/ziti/v2/common/pb/edge_ctrl_pb"
-	"github.com/openziti/ziti/v2/common/posture"
 	cmap "github.com/orcaman/concurrent-map/v2"
 	"google.golang.org/protobuf/proto"
 )
 
 type Cache struct {
-	apiSessionInstances       cmap.ConcurrentMap[string, *Instance]
-	apiSessionInstanceHistory cmap.ConcurrentMap[string, []*InstanceData]
```
andrewpmartinez, line 20: `apiSessionInstanceHistory` was slated for an operator debugging feature. It lets an operator request posture data history for a live session. 

Granted, it never got added, but that was the use case. Operator debugging of posture-checked access denials is temporal and difficult. This was meant to help live debugging features.

We can add it back later.

andrewpmartinez (approved):

PR #4358, opened 2026-09-04
[Backport-2.0] Evict posture data for disconnected api sessions

### LINKED ISSUE #4357: [Backport-2.0] Router posture cache grows without bound as clients re-authenticate
Backport of #4323 (fixed on main by #4356) to release-v2.0.x.

See #4323 for details: the posture cache never removes anything, so every api session that sends posture data leaves an instance behind for the life of the router process, and a full re-authentication mints a new api session id. The per-session instance history retains up to 51 posture snapshots and has never had a reader.

Confirmed present on release-v2.0.x: `apiSessionInstances` has no removal path and `apiSessionInstanceHistory` is written by `saveHistory` and read nowhere.

plorenz:
Fixes #4357. Backport of #4356 (#4323) to release-v2.0.x.

Cherry-pick of 33eea80f3 with three adaptations for this branch:

- sdk-golang import path in the test (no `/v2` here)
- the `Manager` interface and implementation additions were placed without the `SeedMfaFromApiSession` and `GetPostureData` context they sit beside on main, neither of which exists on this branch
- dropped the mid-connect test case: on this branch `AddResponses` is the only path that creates a posture instance (there is no token-seeded instance), so that case reduces to "never reported disconnected", which `Test_EvictInactive_KeepsConnectedSessions` already covers

Worth a look during review: the hunk that clears the recorded disconnect on posture activity applies to the inline `Upsert` in `AddResponses` here, where main has since factored that out into `getOrCreateInstance`. Same behavior, different shape.

### DIFF
--- a/router/xgress_edge/connections.go
+++ b/router/xgress_edge/connections.go
@@ -41,6 +41,18 @@ import (
 	cmap "github.com/orcaman/concurrent-map/v2"
 )
 
+const (
+	// postureEvictionInterval is how often disconnected api sessions' posture data is swept.
+	postureEvictionInterval = time.Minute
+
+	// postureDataRetention is how long posture data is kept after an api session's last
+	// connection to this router goes away. It is generous relative to a reconnect so that a
+	// client that drops and comes back keeps its posture state, most importantly the MFA-passed
+	// time, which a reconnect alone does not always re-establish: the SDK only reports posture
+	// for services it is actively dialing or binding.
+	postureDataRetention = 5 * time.Minute
+)
+
 type identityConnect struct {
 	srcAddr     string
 	dstAddr     string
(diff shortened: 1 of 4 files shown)


router/xgress_edge/connections.go
```diff
@@ -41,6 +41,18 @@ import (
...
+const (
+	// postureEvictionInterval is how often disconnected api sessions' posture data is swept.
+	postureEvictionInterval = time.Minute
+
+	// postureDataRetention is how long posture data is kept after an api session's last
+	// connection to this router goes away. It is generous relative to a reconnect so that a
+	// client that drops and comes back keeps its posture state, most importantly the MFA-passed
+	// time, which a reconnect alone does not always re-establish: the SDK only reports posture
+	// for services it is actively dialing or binding.
+	postureDataRetention = 5 * time.Minute
```
andrewpmartinez, line 53: I spotted a gap.  LTS 2.0.x is missing some TOTP capabilities that were missed in a backport. Probably because that work was done around the time the backport process was introduced.

Nothing to fix here; the issue is logged here: https://github.com/openziti/ziti/issues/4375 I'll start it here shortly.

andrewpmartinez (approved):

Issue #4367, opened 2026-09-08
[Backport-2.0] REST requests no longer reset the legacy API session timeout


plorenz:
Backport of #4365 to release-v2.0.x. Fixed on main by #4366.

PR #4368, opened 2026-09-08
[Backport-2.0] Restore legacy API session activity marking for REST requests

### LINKED ISSUE #4365: REST requests no longer reset the legacy API session timeout
## Problem

Since 2.0.0, REST requests carrying a legacy `zt-session` token no longer count as activity on that API session. The session's `lastActivityAt` is set once at login and never updated by API use, so the API session enforcer deletes it exactly `sessionTimeout` (default 30m) after login regardless of how often the token is used.

Clients that only use the management API and have no edge router connection are hit hardest: there is no other source of activity for their sessions, so any keep-alive strategy based on periodic API calls stops working. On 1.6.x the same client code kept sessions alive indefinitely, and the controller config still documents that "timeouts are reset by API requests".

The session lifetime response headers (`expiration-seconds`, `expires-at`) also stopped being emitted for requests other than the authenticate response.

## Cause

Request handling moved from eager per-request session processing to a lazily resolved `SecurityCtx` in #3561. The eager path called `MarkLastActivityById` for every request with a resolvable `zt-session`; the lazy path never marks activity, and only resolves the session when a permission check or handler asks for it. The wrapper writes the lifetime headers before any handler runs, so under lazy resolution it never sees a session.

Only router heartbeats and tunnel bind/dial handlers mark activity now.

## Fix

Restore the 1.6.x behavior for legacy sessions only:

- resolve a `zt-session` eagerly when the request context is created, on every endpoint, so the request marks activity and the lifetime headers are populated. Bearer tokens stay lazy.
- mark activity in `resolveZtSession` once the session is loaded.

Reported on Discourse against the 2.0.0 quickstart.

plorenz:
For #4367. Backport of #4366 (fixes #4365) to release-v2.0.x.

Since 2.0.0, REST requests carrying a legacy `zt-session` token no longer mark activity on the API session, so sessions for management-API-only clients expire `sessionTimeout` after login regardless of use, and the session lifetime response headers stopped being emitted outside the authenticate response. This restores the 1.6.x behavior for legacy sessions only: a `zt-session` is resolved eagerly when the request context is created and resolution marks the session's last activity. Bearer tokens remain lazy.

Cherry-picked cleanly; build and the API session, authentication, and auth header tests pass on this branch.

### DIFF
--- a/common/security_tokens.go
+++ b/common/security_tokens.go
@@ -17,6 +17,9 @@ import (
 // MaxBearerTokensProcessed represents the maximum number of bearer tokens that will be processed.
 const MaxBearerTokensProcessed = 3
 
+// ZtSessionHeader is the request and response header carrying a legacy API session token.
+const ZtSessionHeader = "zt-session"
+
 // TokenIssuer represents a JWT token issuer capable of verifying tokens.
 // Implementations provide token verification, configuration queries, and claim extraction.
 type TokenIssuer interface {
@@ -397,6 +400,12 @@ func (s *SecurityTokenCtx) AddToRequest(r *http.Request) {
 	*r = *r.WithContext(context.WithValue(r.Context(), SecurityTokenCtxKey, s))
 }
 
+// HasZtSessionHeader reports whether the request carries a non-empty zt-Session header. It
+// inspects the header only; no tokens are parsed or verified.
+func (s *SecurityTokenCtx) HasZtSessionHeader() bool {
+	return s.httpRequest != nil && strings.TrimSpace(s.httpRequest.Header.Get(ZtSessionHeader)) != ""
+}
+
 // IsZtSession returns true if the request carried a legacy zt-Session header. It triggers
 // header processing if not already done.
 func (s *SecurityTokenCtx) IsZtSession() bool {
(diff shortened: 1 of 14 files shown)


common/security_tokens.go
```diff
@@ -397,6 +397,12 @@ func (s *SecurityTokenCtx) AddToRequest(r *http.Request) {
 	*r = *r.WithContext(context.WithValue(r.Context(), SecurityTokenCtxKey, s))
 }
 
+// HasZtSessionHeader reports whether the request carries a non-empty zt-Session header. It
+// inspects the header only; no tokens are parsed or verified.
+func (s *SecurityTokenCtx) HasZtSessionHeader() bool {
+	return s.httpRequest != nil && strings.TrimSpace(s.httpRequest.Header.Get("zt-Session")) != ""
```
andrewpmartinez: Same as main: `zt-session`, all lowercase. You might be able to use a const here, but I'm not sure.

`Header.Get` handles case mismatches for you, so this is a style comment and a future search issue.

andrewpmartinez (approved):

common/security_tokens.go
plorenz: Same as main: There are two existing consts, which were not reachable, so I added a new const in common and pointed all the references there.

PR #4370, opened 2026-09-08
Deliver router ctrl events to each listener in order

### LINKED ISSUE #4369: Router control channel connectivity events can be delivered out of order
## Problem

`networkControllers.notifyOfChange` delivers each ctrl event to each listener on a new goroutine. When a control channel loses its last underlay and regains one moments later, the `ControllerDisconnected` and `ControllerReconnected` events race to the listener and can arrive in either order.

Listeners that key on the transition cannot recover from seeing them backwards. The ER/T hosted service registry tracks `connectedToLeader` from exactly this pair: with the reconnect arriving first it observes a connected leader and does nothing, and the disconnect that follows also sees the leader connected and does nothing, so `TryReestablish` never runs for a real down-and-up and hosted terminators are not re-established.

The producer has the same problem one level up. `notifyOfConnectivityChange` decides the edge with a compare-and-swap and emits afterwards outside any lock, so two callbacks reporting counts zero and one can have their emits interleave even if delivery were ordered.

`TestNotifyOfConnectivityChange_ReconnectReportsBothHalves` asserts the order and fails intermittently in CI because of this. The reordering is scheduler dependent, so it reads as a flake rather than the bug it is.

## Fix

- deliver ctrl events to each listener through a per-listener queue drained one event at a time, so every listener sees events in the order they were emitted and a slow listener delays only itself
- hold the per-channel connectivity state lock from deciding an edge through queuing its event, so a disconnect is always queued before the reconnect that follows it
- collect events in the tests under a lock rather than from a channel drained inside `Eventually` conditions, which raced under `-race`

plorenz:
Fixes #4369

Ctrl events were fanned out to listeners with a goroutine per listener per event, so a disconnect and the reconnect that followed it could reach a listener in either order. The ER/T hosted service registry keys its leader-connected state on exactly that pair, and a reordered pair leaves it believing nothing changed, so hosted terminators are not re-established after a real down-and-up. The same race made `TestNotifyOfConnectivityChange_ReconnectReportsBothHalves` fail intermittently in CI.

Each listener now gets its own queue, drained one event at a time, so it sees events in emission order and a slow listener delays only itself. On the producer side, the per-channel connectivity state lock is held from deciding an edge through queuing its event, so the disconnect is always queued first. The test collectors also move off a channel drained inside `Eventually` conditions, which raced under `-race`.

Extracted from a larger in-progress branch so the ordering fix and the CI flake can land on their own. `go test -race -count=200` on the affected tests is clean.

### DIFF
--- a/router/env/ctrl_event_queue.go
+++ b/router/env/ctrl_event_queue.go
@@ -1,0 +1,62 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package env
+
+import "sync"
+
+// ctrlEventQueue delivers ctrl events to one listener in the order they were queued, one at a time,
+// on a goroutine of its own. A listener that keys on transitions must see a disconnect before the
+// reconnect that follows it, and a slow listener must delay neither the caller nor other listeners.
+type ctrlEventQueue struct {
+	listener CtrlEventListener
+	lock     sync.Mutex
+	pending  []CtrlEvent
+	draining bool
+}
+
+func newCtrlEventQueue(listener CtrlEventListener) *ctrlEventQueue {
+	return &ctrlEventQueue{listener: listener}
+}
+
+// enqueue adds event and starts a drain if none is running. It never blocks on the listener.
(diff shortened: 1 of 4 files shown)


router/env/ctrl_event_queue.go
```diff
@@ -0,0 +1,62 @@
...
+			self.draining = false
+			self.lock.Unlock()
+			return
+		}
+		event := self.pending[0]
+		self.pending = self.pending[1:]
+		self.lock.Unlock()
+
+		self.listener.NotifyOfCtrlEvent(event)
+	}
```
andrewpmartinez, line 61: Won't this hot loop? I'd expect to see a `for` on a channel or something.

andrewpmartinez (approved):
Concerned about the `for` loop that isn't operating on a channel.

router/env/ctrl_event_queue.go
plorenz: The goroutine gets spun up when there are events and only runs until it doesn't have any events to process. It will exit the for loop once the loop is empty, so it should never be looping, not doing work. When events are queued they'll spin up a new goroutine for delivery if one isn't running yet.

Issue #4371, opened 2026-09-08
[Backport-2.0] Router control channel connectivity events can be delivered out of order


plorenz:
Backport of #4369 to release-v2.0.x. Fixed on main by #4370.

PR #4372, opened 2026-09-08
[Backport-2.0] Deliver router ctrl events to each listener in order

### LINKED ISSUE #4369: Router control channel connectivity events can be delivered out of order

plorenz:
For #4371. Backport of #4370 (fixes #4369) to release-v2.0.x.

Ctrl events were fanned out to listeners with a goroutine per listener per event, so a disconnect and the reconnect that followed it could reach a listener in either order. The ER/T hosted service registry keys its leader-connected state on exactly that pair, so a reordered pair leaves hosted terminators un-re-established after a real down-and-up. Each listener now gets its own queue drained in order, and the producer holds the connectivity state lock from deciding an edge through queuing its event.

This branch had already taken the test-only relaxation from #4266 via #4265, which asserted the pair as an unordered set. The cherry-pick conflicted there; the resolution restores the ordered assertions from main, since the order is now guaranteed, and drops the `awaitCtrlEvents` helper that only that relaxation used. `go test -race -count=200` on the affected tests is clean.

### DIFF
--- a/router/env/ctrl_event_queue.go
+++ b/router/env/ctrl_event_queue.go
@@ -1,0 +1,62 @@
+/*
+	Copyright NetFoundry Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package env
+
+import "sync"
+
+// ctrlEventQueue delivers ctrl events to one listener in the order they were queued, one at a time,
+// on a goroutine of its own. A listener that keys on transitions must see a disconnect before the
+// reconnect that follows it, and a slow listener must delay neither the caller nor other listeners.
+type ctrlEventQueue struct {
+	listener CtrlEventListener
+	lock     sync.Mutex
+	pending  []CtrlEvent
+	draining bool
+}
+
+func newCtrlEventQueue(listener CtrlEventListener) *ctrlEventQueue {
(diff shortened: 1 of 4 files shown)


router/env/ctrl_event_queue.go
```diff
@@ -0,0 +1,62 @@
...
+			self.draining = false
+			self.lock.Unlock()
+			return
+		}
+		event := self.pending[0]
+		self.pending = self.pending[1:]
+		self.lock.Unlock()
+
+		self.listener.NotifyOfCtrlEvent(event)
+	}
```
andrewpmartinez, line 61: Same concern as on the `main` PR.

andrewpmartinez (approved):

router/env/ctrl_event_queue.go
plorenz: see response on main PR

PR #4374, opened 2026-09-09
[Backport-2.0] propagate crypto material supplied in DialSuccess

### LINKED ISSUE #4373: [Backport-2.0] improve crypto exchange
backport of #3864 (fixed in #3865)

ekoby:
[fixes #4373] 

cherry-picked fix from #3874

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,7 @@
 
 ### End-to-End Encryption (e2ee) Improvements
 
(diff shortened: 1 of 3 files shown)


plorenz (changes requested):
Passing the crypto header through breaks go -> c-sdk compatibility because one uses a number and the other uses a string. We need to fix that somehow before we enable the pass-through

plorenz (approved):
changelog needs a small tweak to be correct, but otherwise looks good

Issue #4377, opened 2026-09-09
Circuit endpoint faults are sent inline from arbitrary goroutines and lost on failure


plorenz:
Applies to the router (v2.x).

When a circuit's ingress or egress endpoint goes away, the router tells the controller with an `IngressFault`/`EgressFault` so it can remove the circuit. Two places send that fault, both inline and both with a bare `Send`:

- `router/xgress_edge/listener.go` (`cleanupXgressCircuit`)
- `router/handler_xgress/close.go`

Two problems.

**The send blocks whichever goroutine tore the circuit down.** `Send` with no deadline exits only when the queue drains or the channel closes, and the router-to-controller channel's priority queues are 16 deep. The callers are not dedicated to that send: circuit teardown reaches `cleanupXgressCircuit` from the connection read loop, and from the posture evaluation pool when access is revoked. A controller that stops draining can therefore park a read loop, or exhaust the shared posture pool and stall enforcement for unrelated identities. This is CH-S1.

**A failed send loses the fault, with no backstop.** The `Faulter` does not carry these faults: it sends `ForwardFault` (from `Report`) and `LinkFault` (from `NotifyInvalidLink`) only, and it has no retry — it removes a circuit id before sending on the assumption that the forwarder will report again because traffic keeps arriving. A revoked endpoint generates no traffic, so nothing re-triggers it. The controller handles these subjects with `RemoveCircuit`, and there is no stale-circuit scan, so a dropped fault leaves the circuit in the controller until the other endpoint faults or a router restarts.

Endpoint faults should be handed to an owner that sends them off the caller's goroutine and has explicit behavior for retry and expiry, rather than each call site deciding.

PR #4391, opened 2026-09-10
[Backport-2.0] Deliver circuit endpoint faults through the faulter

### LINKED ISSUE #4390: [Backport-2.0] Circuit endpoint faults are sent inline from arbitrary goroutines and lost on failure
Backport of #4377 (fixed on main by #4388) to release-v2.0.x.

Confirmed present on release-v2.0.x: `cleanupXgressCircuit` sends the circuit fault inline with a bare `Send`, from paths including the connection read loop, and a failed send loses the fault with nothing to re-report it.

plorenz:
Fixes #4390. Backport of #4388 (#4377) to release-v2.0.x.

Cherry-pick of 6c7d07876 with three adaptations for this branch:

- `channel/v4` and non-`/v2` sdk-golang import paths. All the APIs the change needs exist on channel/v4 v4.3.12: `AddReceiveHandlerF`, `SendAndWaitForWire` with the same `NotifyErr` buffering, and the same `DefaultOutQueueSize` of 4. `common/ctrlchan` with its priority senders, `GetCtrlChannel`, and the `ControllerRemoved`/`ControllerReconnected` events are all present here too.
- `pfxlog` in place of the structured `faultLog` logging channel, which is main-only.
- `handler_xgress/close.go` drops its `protobufs` import, now unused once the fault send moves to the faulter.

Two pre-existing failures on this branch, neither caused by this change, both verified against pristine `release-v2.0.x` in a clean worktree:

- `go vet -tags=perf ./router/test/` already fails to build, on `dataPlaneAdapter` not satisfying `env.RouterEnv`. This change shifts the reported line from 162 to 165 because the test double gains a method, but the error is the same one. The double is updated for the new interface method so no second breakage is added.
- `TestDataFlowAndCloseLogic` in `router/xgress_validation` fails under `-race` (consistently, ~15.7s, so a timeout), and passes without it. Same on pristine.

### DIFF
--- a/router/env/config_forwarder.go
+++ b/router/env/config_forwarder.go
@@ -23,9 +23,16 @@ import (
 )
 
 const (
-	DefaultXgressCloseCheckInterval    = 5 * time.Second
-	DefaultXgressDialDwellTime         = 0
-	DefaultFaultTxInterval             = 15 * time.Second
+	DefaultXgressCloseCheckInterval = 5 * time.Second
+	DefaultXgressDialDwellTime      = 0
+	DefaultFaultTxInterval          = 15 * time.Second
+
+	// DefaultEndpointFaultRetention is how long the router keeps retrying an unsent circuit
+	// endpoint fault before giving up on it. Configured as a duration, such as '5m'. Retention bounds the memory held while a controller
+	// is unreachable, since a fault outlives the circuit it refers to: the set grows with the
+	// teardown rate for as long as the outage lasts. Giving up leaves the controller holding a
+	// circuit the router no longer has, until the controller reconciles circuits on reconnect.
+	DefaultEndpointFaultRetention      = 5 * time.Minute
 	DefaultIdleTxInterval              = 60 * time.Second
 	DefaultIdleCircuitTimeout          = 60 * time.Second
 	DefaultXgressDialWorkerQueueLength = 1000
@@ -54,6 +61,7 @@ const (
 )
 
 type ForwarderOptions struct {
+	EndpointFaultRetention   time.Duration
 	FaultTxInterval          time.Duration
 	IdleCircuitTimeout       time.Duration
 	IdleTxInterval           time.Duration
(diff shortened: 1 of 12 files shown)


router/env/config_forwarder.go
```diff
@@ -96,6 +105,18 @@ func DefaultForwarderOptions() *ForwarderOptions {
...
+		if !ok {
+			return nil, errors.New("invalid value for 'endpointFaultRetention', expected a duration, such as '5m'")
+		}
+
+		var err error
+		if options.EndpointFaultRetention, err = time.ParseDuration(val); err != nil {
+			return nil, errors.Wrap(err, "invalid value for 'endpointFaultRetention'")
+		}
+	}
+
```
andrewpmartinez, line 119: Should a zero or negative `endpointFaultRetention` be rejected here? `expirePending` returns early when retention is not positive, so `0s` turns expiry off and the pending set grows for as long as a controller is unreachable, which is the opposite of what an operator writing zero would expect.

router/forwarder/endpoint_faulter.go
```diff
@@ -0,0 +1,278 @@
...
+func (self *endpointFaultSender) run() {
+	log := pfxlog.Logger().WithField("ctrlId", self.ctrlId)
+	log.Info("endpoint fault sender started")
+	defer log.Info("endpoint fault sender exited")
+
+	ticker := time.NewTicker(endpointFaultSweepInterval)
+	defer ticker.Stop()
+
+	for {
+		self.expirePending()
```
andrewpmartinez, line 141: During an outage, this runs a full scan of the pending set on every wake. Every report wakes it.

- The scan holds the same lock `report` takes, so the reporters on the connection read loop wait behind it.
- The set grows with the teardown rate for the length of the outage, the scan gets more expensive

router/forwarder/endpoint_faulter_test.go
```diff
@@ -0,0 +1,327 @@
...
+	"github.com/openziti/ziti/v2/common/pb/ctrl_pb"
+	"github.com/openziti/ziti/v2/router/env"
+	cmap "github.com/orcaman/concurrent-map/v2"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/proto"
+)
+
+// testCtrlChannel completes ctrlchan.CtrlChannel over BaseCtrlChannel and runs a goroutine in
+// place of an underlay, so that sends waiting for a write actually complete. Setting writeErr
+// makes those writes fail the way channel v5.0.27 reports it: NotifyErr, then NotifyAfterWrite.
```
andrewpmartinez, line 22: This names channel v5.0.27. The branch builds against channel/v4 v4.3.12?

andrewpmartinez (approved):

Issue #4394, opened 2026-09-11
Router does not relay an SDK xgress terminator's half-close to tunneler, proxy, or embedded SDK initiators


plorenz:
## Summary

When a service is hosted by an SDK app that binds with SDK flow control, the terminator is an
xgress conn and the only end-of-write signal that crosses the fabric when the host half-closes
is the native xgress EOF payload flag. On the dialing router that flag ends the xgress tx loop,
which can only relay the half-close to its peer conn through `xgress.SignalConnection`. The
shared `xgress_common.XgressConn` does not implement it, so any initiator built on it never
half-closes the local connection: the intercepted or proxied client receives every byte and its
read then hangs until its own timeout fires.

Affected initiators, all built on `XgressConn` with half-close enabled:

- `xgress_edge_tunnel` intercepts, both tproxy and proxy mode (a tunneling edge router as the client)
- `xgress_proxy` listeners
- `xgress_sdk` (router-embedded SDK)

This is the same shape as #4338, which covered a legacy edge client peer (`edgeXgressConn`),
and it surfaced from the review of that fix's backport (#4341). The circuit only stalls while
the client holds its own write side open, so downloads and any protocol that signals end of
transfer by closing are the cases that break.

## Mechanism

The SDK host's `CloseWrite` sends a plain edge Data message carrying the FIN flag and then calls
`CloseRxTimeout`, which emits the xgress EOF flag. For a flow-control terminator the router only
registers an xgress edge forwarder for the circuit and never adds a message sink for the conn
id, so the Data message is dropped in `handleDataMessage` as an unknown conn id. FIN as a
payload header only ever originates from an `XgressConn`-backed terminator or a legacy SDK
terminator, which is why `XgressConn.WritePayload`'s FIN handling never fires here.

## Affected versions

`main`, `release-v2.0.x` and `release-v1.6.x`. Unlike #4338 there is no version where this
already works, since `XgressConn` has never implemented `SignalConnection`.

## Fix

Implement `FlowFromFabricToXgressClosed` on `XgressConn`, reusing the write-side close that
`WritePayload` already performs on a FIN header (`CloseWrite` when the conn supports it, else
`Close`) and guarding it so the FIN-header and EOF-flag paths together close the write side once.

`Test_TunnelerDataflowHalfClose` hosts the service on the tunneler router itself, so both peers
are `XgressConn` and the seam is not covered. Coverage should pair each initiator conn type with
each terminator conn type rather than adding another single-pair test.

PR #4403, opened 2026-09-15
Route library slog output through pfxlog on release-v2.0.x

### LINKED ISSUE #4401: Route library slog output through pfxlog on release-v2.0.x and release-v1.6.x
## Summary

The OpenZiti libraries are moving from logrus/pfxlog to `log/slog`, taking their loggers from
foundation's `logging.For(name)` registry. Ziti 1.6.x and 2.0.x still log through pfxlog, and
neither line configures that registry. Its bootstrap root is a `slog.TextHandler` on stderr at
trace level, so the first converted library either line pulls in (transport at v2.0.216 and later,
which #4394 needs for its tcp half-close fix) will emit every trace and debug line it has,
unfiltered and in a different format from the rest of the process.

## Fix

Add a `slog.Handler` that turns records into logrus entries and install it as foundation's root
handler and as the stdlib `slog` default:

- `Enabled` defers to `logrus.IsLevelEnabled` on every call, so the process's configured level and
  `ziti agent set-log-level` apply to library output with no level mirroring. The foundation
  registry is set to trace so it never filters ahead of the bridge.
- Records carry their caller: the record's PC is resolved to a `runtime.Frame` and set on the
  entry before logrus can fill it, so pfxlog attributes the line to the library function that
  logged it.
- Attributes become fields, with groups flattened to dotted keys. Fatal and panic records are
  emitted at their own logrus severity through `Entry.Log`, which unlike `Entry.Fatal` and
  `Entry.Panic` never exits or panics; that stays with the caller.
- pfxlog channels are only used inside the ziti tree, so no per-channel level plumbing is needed
  for library loggers.

Applies to `release-v2.0.x` and `release-v1.6.x`. Not needed on `main`, which has moved to slog.

plorenz:
Fixes #4401.

The OpenZiti libraries are moving to `log/slog` through foundation's `logging` registry. Its bootstrap root is a stderr text handler at trace level, so the first converted library this line pulls in (transport past v2.0.215, which #4394 needs) would emit every trace and debug line unfiltered and in a foreign format.

This adds `common/slogbridge`, a `slog.Handler` that turns records into logrus entries: matching level, attrs as fields with groups flattened, and the record's caller set on the entry so pfxlog attributes lines to the library function. It gates on the logrus level at call time, so the process level and `ziti agent set-log-level` govern library output with no mirroring. `Install` makes it the registry root and the `slog` default, with the registry opened to trace, and runs from the CLI root init and the test package init.

The registry arrived in foundation 2.0.96, so this also moves foundation forward, to a pre-release commit until v2.0.102 is tagged.

Verified by temporarily pulling transport v2.0.221 and running a tunneler dataflow test: no raw slog lines, and transport's lines come out as pfxlog with the real caller.


common/slogbridge/bridge.go
```diff
@@ -0,0 +1,170 @@
...
+	logging.Configure(h)
+	logging.SetGlobalLevel(logging.LevelTrace)
+	slog.SetDefault(slog.New(h))
+}
+
+func (h *Handler) Enabled(_ context.Context, level slog.Level) bool {
+	return h.logger.IsLevelEnabled(logrusLevel(level))
+}
+
+func (h *Handler) Handle(_ context.Context, r slog.Record) error {
```
andrewpmartinez, line 75: Should `Handle` check the logrus level before it builds anything?

- The registry's named handler forwards to the root without asking the root whether it is enabled
- Opening the registry to trace means a suppressed library debug or trace line still allocates the fields map and symbolizes its caller frame, then gets dropped inside `Entry.Log`.

common/slogbridge/bridge.go
```diff
@@ -0,0 +1,170 @@
...
+func Install() {
+	InstallTo(logrus.StandardLogger())
+}
+
+// InstallTo is Install for a specific logrus logger.
+func InstallTo(logger *logrus.Logger) {
+	h := New(logger)
+	logging.Configure(h)
+	logging.SetGlobalLevel(logging.LevelTrace)
+	slog.SetDefault(slog.New(h))
```
andrewpmartinez, line 68: There is a side effect tied to net/https logging. I think we are okay with it, but just to check:

- `slog.SetDefault` also points `log.Default()` at this handler, at info, and calls `log.SetFlags(0)`. Anything using `log.Print` now runs through logrus.
- net/http's default server error log goes that way, and at warn or above it now disappears

andrewpmartinez (approved):
Some small questions for you to consider, but no show stoppers.

common/slogbridge/bridge.go
plorenz: short circuit added

common/slogbridge/bridge.go
plorenz: I think we want logging to all go through our handler. I did add a call to `log.SetFlags(log.Lshortfile)`.

PR #4404, opened 2026-09-15
Route library slog output through pfxlog on release-v1.6.x

### LINKED ISSUE #4401: Route library slog output through pfxlog on release-v2.0.x and release-v1.6.x

plorenz:
Fixes #4401.

Same change as #4403: `common/slogbridge` turns slog records into logrus entries with matching level, attrs as fields, and the record's caller on the entry, gated on the logrus level at call time. `Install` makes it the root of foundation's logging registry and the `slog` default, with the registry opened to trace, and runs from the CLI root init and both test package inits.

The registry arrived in foundation 2.0.96 and this line pinned foundation to 2.0.86, so this also drops that pin and moves foundation forward, to a pre-release commit until v2.0.102 is tagged. Crossing that range renames `errorz.ApiError.Code` to `AppCode`, which accounts for the controller changes here; they are mechanical and the build enumerates them.

Verified by temporarily pulling transport v2.0.221 and running a tunneler dataflow test: no raw slog lines, transport lines attributed correctly through pfxlog. The `dataflow`-tagged test package does not compile on this branch today for an unrelated, pre-existing reason and is not run in CI here.


andrewpmartinez (approved):
Sent a comment about `controller/handler_peer_ctrl/response.go:52` potentially trying to marshal a `func()` into JSON.

plorenz:
> Sent a comment about `controller/handler_peer_ctrl/response.go:52` potentially trying to marshal a `func()` into JSON.

fixed

PR #4405, opened 2026-09-15
Publish first-party CA roots without intermediates. Fixes #4406

### LINKED ISSUE #4406: Router data model root CA public keys carry different intermediates on each controller
### Symptom

In an HA cluster where each controller has its own intermediate CA, the router data model's root CA public key entry has different `Intermediates` depending on which controller you ask. The router keeps whichever controller's version arrived last.

Visible as:

- The fablab `router-data-model-test` validation passes against one controller and reports `public-keys ... Intermediates ... mismatch` against the others, on every run since v2.0.4.
- `ziti fabric inspect router-data-model` diffs report the same drift.
- A client presenting only its leaf certificate (intermediate not in its identity) passes the router's first-party cert check only if the router last synced from the controller that enrolled it. That can flip on a controller restart with no configuration change.

### Cause

Since v2.0.4 (#4094 and the 2.0.x backport #4153), each controller publishes the root CA public key with the intermediates from its own CA bundle. The godoc on `BuildPublicKeys` already states the rule this breaks: a key id emitted by several controllers must carry identical content so the sender and router models converge under last-writer-wins by kid. Per-controller intermediates make the same kid carry three different payloads.

### Fix

Publish first-party root CA keys without intermediates. Routers take intermediates from the chain the client presents, which is how first-party verification worked before v2.0.4 and what every SDK and tunneler does. The bare-leaf case that the intermediates enabled as a side effect never worked before v2.0.4 and worked only nondeterministically in HA since.

Affects main and release-v2.0.x (v2.0.4 onward).

plorenz:
## What

Each controller publishes the root CA public key into the router data model with intermediates drawn from its own CA bundle. In an HA cluster where every controller has its own intermediate, the controllers emit different payloads under the same key id. Routers keep whichever arrived last, so a router's copy depends on which controller synced it most recently, and every controller-to-router comparison (the RDM validation test, `ziti fabric inspect router-data-model`) reports a mismatch on every run.

This change publishes first-party root CA keys without intermediates. A root's kid is shared by every controller and a controller's bundle intermediates are its own, so they cannot be part of a shared entry. Routers already take intermediates from the chain a client presents, which is how first-party verification worked before v2.0.4.

## Why

Present since v2.0.4, where publishing intermediates arrived as a side effect of publishing bundle roots as first-party anchors. Before that, a client had to present its chain, and every SDK and tunneler does, since the controller returns the chain at enrollment. The only behavior that goes away is a client presenting a bare leaf issued by an intermediate, which never verified before v2.0.4 and only verified nondeterministically in HA since.

Third-party CA entries are unchanged; their chains are admin-provided and stored, so they are already consistent across controllers.

## Notes

- The ha-3 test configs now give each controller only its own signing intermediate, matching real deployments. The new convergence assertion fails without the product change.
- In a mixed-version cluster, controllers still on v2.0.4 or v2.0.5 keep publishing their intermediates until upgraded.


andrewpmartinez:
I'm pretty sure this breaks multi-CA setups for some quickstart-provisioned HA environments. It works fine for modern quick-start-deployed setups, though.

I need to look at my notes for when I dug into this issue.

PR #4408, opened 2026-09-16
[Backport-2.0] Publish first-party CA roots without intermediates. Fixes #4407

### LINKED ISSUE #4407: [Backport-2.0] Router data model root CA public keys carry different intermediates on each controller
Backport of #4406 (PR #4405) to `release-v2.0.x`.

The v2.0.4 change for first-party certs issued by a separate signing CA (#4153) has each controller publish the root CA public key with the intermediates from its own CA bundle. In an HA cluster with per-controller intermediates the same kid carries a different payload on each controller, the router keeps whichever arrived last, and the `router-data-model-test` validation fails against non-leader controllers on every run.

Same fix as main: publish first-party root CA keys without intermediates. Routers take intermediates from the chain the client presents, as they did before v2.0.4.

plorenz:
Backport of #4405 to `release-v2.0.x`. Fixes #4407, for #4406.

## What

Each controller publishes the root CA public key into the router data model with intermediates drawn from its own CA bundle. In an HA cluster where every controller has its own intermediate, the controllers emit different payloads under the same key id. Routers keep whichever arrived last, so a router's copy depends on which controller synced it most recently, and every controller-to-router comparison (the RDM validation test, `ziti fabric inspect router-data-model`) reports a mismatch on every run.

This change publishes first-party root CA keys without intermediates. A root's kid is shared by every controller and a controller's bundle intermediates are its own, so they cannot be part of a shared entry. Routers already take intermediates from the chain a client presents, which is how first-party verification worked before v2.0.4.

## Why

Present on this line since v2.0.4 (#4153), where publishing intermediates arrived as a side effect of publishing bundle roots as first-party anchors. The `router-data-model-test` validation has failed against non-leader controllers on every release-v2.0.x run since. The only behavior that goes away is a client presenting a bare leaf issued by an intermediate, which never verified before v2.0.4 and only verified nondeterministically in HA since.

Third-party CA entries are unchanged.

## Notes

- Clean cherry-pick; the zititest module sync from the main PR was dropped since it belongs to main's dependency set.
- Second commit adds `unparseable` to the codespell ignore list to match main, so backports of code using that spelling pass the check.
- Controllers still on v2.0.4 or v2.0.5 keep publishing their intermediates until upgraded.


andrewpmartinez:
As stated on the other PR for main, I'm pretty sure this breaks a slice of quickstart networks that created HA environments with multiple CAs.

Issue #4411, opened 2026-09-16
[Backport-2.0] REST error responder logs ApiError.Code as a method value


plorenz:
Backport of #4410 to `release-v2.0.x`. See #4410 for the mechanism. Fixed alongside #4401 in #4403, since that PR is already on the branch.

Issue #3747, opened 2026-04-03
Update controller peer error marshalling for app code changes


plorenz:
Encoding is still using Code instead of replacement AppCode

PR #4412, opened 2026-09-16
Log the REST error responder's application code, not a method value

### LINKED ISSUE #4410: REST error responder logs ApiError.Code as a method value
When `errorz.ApiError.Code` became `AppCode` (openziti/foundation#470), `Code()` was added as a method returning the HTTP status. `EdgeResponseMapper.toRestModel` in `controller/response/responder.go` still logs `WithField("code", e.Code)`, which now binds the method rather than reading the field. It compiles, and the debug line renders a function value instead of the application code; the JSON formatter cannot marshal it at all.

Same class as the peer-controller error encoder fixed in #3747. The fix is `e.AppCode`.

Present on `main` and `release-v2.0.x`; `release-v1.6.x` picks it up with the rename in #4404.

plorenz:
Fixes #4410.

`EdgeResponseMapper.toRestModel` logs `WithField("code", e.Code)`. Since the `AppCode` rename, `Code` is a method returning the HTTP status, so the expression binds a method value: it compiles, renders as a function value in the debug line, and fails to marshal under the JSON formatter. Same class as the peer-controller encoder fix in #3747; a sweep of the tree for non-call `.Code` selectors on `ApiError` finds no others. Backported to 2.0.x in #4403 as #4411.


dovholuknf (approved):

controller/response/responder.go
```diff
@@ -55,7 +55,7 @@ func (self EdgeResponseMapper) MapApiError(requestId string, apiError *errorz.Ap
 }
 
 func (self EdgeResponseMapper) toRestModel(e *errorz.ApiError, requestId string) *rest_model.APIError {
-	pfxlog.Logger().WithError(e).WithField("code", e.Code).Debug("returning error to REST API")
+	pfxlog.Logger().WithError(e).WithField("code", e.AppCode).Debug("returning error to REST API")
```
dovholuknf, line 58: 'code' -> 'AppCode' or no?

dovholuknf, line 58: 'code' -> 'appcode' or no?

dovholuknf (approved):

controller/response/responder.go
plorenz: I think the log message should stay the same

plorenz: I think the log message should stay the same. Code was change to AppCode to handle an API naming conflict, but it shouldn't be visible to users

Issue #4413, opened 2026-09-16
debian controller & router packages hang on post-install script when installed via ansible


gladiatr72:
```sh
$ ansible-playbook -v ./ziti-router-playbook.yml --limit r1
[...]
TASK [ziti_router : Ensure the OpenZiti apt GPG keyring is present] *******************************************************************************************
ok: [r1] => {"changed": false, "cmd": "set -euo pipefail\ncurl -sSLf https://get.openziti.io/tun/package-repos.gpg | gpg --dearmor -o /usr/share/keyrings/openziti.gpg\n", "delta": null, "end": null, "msg": "Did not run command since '/usr/share/keyrings/openziti.gpg' exists", "rc": 0, "start": null, "stderr": "", "stderr_lines": [], "stdout": "skipped, since /usr/share/keyrings/openziti.gpg exists", "stdout_lines": ["skipped, since /usr/share/keyrings/openziti.gpg exists"]}

TASK [ziti_router : Add the OpenZiti apt repository] **********************************************************************************************************
ok: [r1.revsys.rocks] => {"changed": false, "repo": "deb [signed-by=/usr/share/keyrings/openziti.gpg] https://packages.openziti.org/zitipax-openziti-deb-stable debian main", "sources_added": [], "sources_removed": [], "state": "present"}

TASK [ziti_router : Install the openziti-router package] ******************************************************************************************************
[hangs here]
```

on r1:

```
    747 ?        Ss     0:00  \_ sshd-session: root [priv]
    755 ?        S      0:00      \_ sshd-session: root@pts/1
    892 pts/1    Ss+    0:00          \_ /bin/sh -c DEBIAN_FRONTEND=noninteractive /usr/bin/python3.13 /root/.ansible/tmp/ansible-tmp-1789577843.473947-1411466-90268571827054/AnsiballZ_apt.py && sl
    893 pts/1    S+     0:00              \_ /usr/bin/python3.13 /root/.ansible/tmp/ansible-tmp-1789577843.473947-1411466-90268571827054/AnsiballZ_apt.py
    900 pts/1    S+     0:00                  \_ /usr/bin/apt-get -y -o Dpkg::Options::=--force-confdef -o Dpkg::Options::=--force-confold -o DPkg::Lock::Timeout=60 install openziti-router=2.0.4
    925 pts/2    Ss+    0:00                      \_ /usr/bin/dpkg --force-confdef --force-confold --status-fd 23 --configure --pending
[...]

PR #4414, opened 2026-09-16
fix openziti/ziti#4413 debian package post-install hang

gladiatr72:
replaces simple stdin check in the current router and controller postinstall scripts with typical debconf mechanisms


gladiatr72:
recheck cla

dovholuknf:
Thanks for the pr @gladiatr72 - i received your cla in email. We are going to update the action to make it more clear to people . Something like:

To accept the cla through this pull request, add a comment with the words: "I have read the CLA Document and I hereby sign the CLA"

The recheck cla won't work until I add you to the cla manually (the email is a manual process) i'll do that now tho...

dovholuknf:
recheck cla

dovholuknf:
cla acceptance registered manually with https://github.com/netfoundry/cla/commit/d5cbb660e782268579e623b641868dcba2e0bfad

dist/dist-packages/linux/openziti-controller/templates
```diff
@@ -0,0 +1,9 @@
+Template: openziti-controller/configure-now
+Type: boolean
+Default: false
```
scareything: defaulting to false might be confusing, considering: any package users that type 'n' from muscle-memory will be disappointed, and the non-debconf path still defaults to yes.

I think it's ok to change this to true, since the debconf variable isn't checked at all if config.yml exists and I'm assuming the false default intended to avoid overwriting an existing config file?

dist/dist-packages/linux/openziti-router/templates
```diff
@@ -0,0 +1,9 @@
+Template: openziti-router/configure-now
+Type: boolean
+Default: false
```
scareything: same as above

scareything (approved):

